PR #11, opened 2018-12-14
Setup docker

dankmitchell:
### Context
Docker

Docker

### Guidance to review
`docker-compose up --build` 🤞


README.md
```diff
@@ -18,6 +18,14 @@
 
 It's best to lint just your app directories and not those belonging to the framework, e.g.
 
+## Docker
```
tvararu: This readme section should go before linting I think

tvararu: Also worth mentioning that you still need to `yarn` and set up the https certs

Dockerfile
```diff
@@ -0,0 +1,21 @@
+FROM ruby:2.5.3
```
defong: can we not use 2.5.3-alpine3.8 ?

PR #16, opened 2018-12-20
Add custom Kudu deployment script

tvararu:
### Context

Hopefully will lead to Azure recognising this app as a Rails app instead of a Node.js app.

`deploy.sh` is copied from the support app, autogenerated by Kudu.


tvararu:
This ended up being a dead end.

PR #23, opened 2019-01-02
Run tests in docker

timabell:
This removes a duplicate bundle etc that was happening outside the container


timabell:
[Use Docker to deploy manage-courses-frontend](https://trello.com/c/3bJZRX4K/820-use-docker-to-deploy-manage-courses-frontend)

PR #29, opened 2019-01-03
Login to AzureCR and pull master docker layers before building

tvararu:
This should allow us to build with a cache of the previous layers, which should be much quicker because it should skip bundling when the Gemfile doesn't change.


timabell:
This drops the build time from 5 mins to 2mins 40 secs, unless you touch the gemfile in which case you get a 6 min build.

PR #30, opened 2019-01-03
DRY the docker commands

timabell:


timabell:
Seems to break the caching at the Gemfile layer for reasons I can't fathom

timabell:
Probably was because we'd pushed straight to the master container from a local machine and that had different checksums for reasons.

timabell:
[Use Docker to deploy manage-courses-frontend on dev](https://trello.com/c/3bJZRX4K/820-use-docker-to-deploy-manage-courses-frontend)

PR #46, opened 2019-02-01
Going alpine

defong:
### Context
Going alpine

Gone alpine

### Guidance to review
I think there maybe some weird race condition in the docker-compose.yml

@tvararu can actually give it proper thrashing

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,15 +1,11 @@
-FROM ruby:2.5.3
+FROM ruby:2.5.3-alpine
+RUN apk add --update --no-cache tzdata && \
+    cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
+    echo "Europe/London" > /etc/timezone
 
-RUN apt-get update && apt-get install apt-transport-https
-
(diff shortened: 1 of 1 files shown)


defong:
[Switch manage-courses-frontend and manage-courses-backend to a lighter base Docker image](https://trello.com/c/4qhTJ1DZ/842-switch-manage-courses-frontend-and-manage-courses-backend-to-a-lighter-base-docker-image)

Dockerfile
```diff
@@ -1,23 +1,23 @@
...
 WORKDIR $APP_HOME
 
 ADD Gemfile $APP_HOME/Gemfile
 ADD Gemfile.lock $APP_HOME/Gemfile.lock
-RUN bundle install
+
+RUN apk add --update --virtual build-dependances \
+ build-base  && \
+ bundle install --jobs=4 && \
+ apk del build-dependances
```
misaka: Does using `--virtual` mean that you don't need to clear the cache/use `--no-cache`?

defong: https://stackoverflow.com/questions/49118579/alpine-dockerfile-advantages-of-no-cache-vs-rm-var-cache-apk

PR #49, opened 2019-02-11
[960] Add organisations list

dankmitchell:
### Context
Connect `manage-courses-backend` with `manage-courses-frontend`

Setup `get` request to `api/v2/providers` in `manage-courses-backend`

### Guidance to review
`/organisations`

# After
screenshot 2019-02-11 at 12 46 46

### DIFF
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,8 +1,6 @@
 class PagesController < ApplicationController
   before_action :authenticate
 
-  def home; end
-
   def show
     render template: "pages/#{params[:page]}"
   end
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -1,0 +1,7 @@
+class ProvidersController < ApplicationController
+  before_action :authenticate
+
+  def index
+    @providers = Provider.all
+  end
+end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -1,0 +1,2 @@
+class Provider < Base
+end
--- a/app/views/pages/home.html.erb
+++ b/app/views/pages/home.html.erb
@@ -1,3 +1,0 @@
-<%= content_for :page_title, "Home" %>
-
-<p>Lorem</p>
--- a/app/views/providers/_provider.html.erb
+++ b/app/views/providers/_provider.html.erb
@@ -1,0 +1,3 @@
+<li>
+  <h2 class="govuk-heading-m"><%= link_to provider.institution_name, "#{ENV['MANAGE_UI_BASE_URL']}/organisations/#{provider.institution_code}" %></h2>
+</li>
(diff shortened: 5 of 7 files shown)


app/models/organisation.rb
```diff
@@ -0,0 +1,17 @@
+class Organisation
+  include ActiveModel::Model
+
+  def self.all
```
benilovj: I'd pull out a `ManageCoursesBackend` class with a `get_all_organisations` method on it, that this method delegates to. This class then would encapsulate HTTParty, all the error handling, etc. Otherwise you're gonna get lots of repetition in your models. It's premature optimisation for this PR but future PRs will be less noisy as a result.

dankmitchell: After looking through the schema today I think you are right. Add `get_all_organisations` on the user model in the backend is def the best course of action 👌

benilovj:
Just realised we have to be quite careful with naming. We have two separate concepts: organisation (which is a DfE view) and provider (which is the marketing & UCAS view). We don’t actually ever surface DfE orgs to end-users, it’s purely an internal concept. I think the code in the frontend should probably align with the backend, so the model should probably be called `Provider` instead of `Organisation` even if the copy we use says Organisation.

app/models/organisation.rb
misaka: Although I'd go with the `ManageCoursesBackend` approach if we were to roll our own API client, I think I like the approach of using an existing ORM-like wrapper. I've used ActiveResource with success in the past, and it looks like it's been revived recently, but there's a couple other options to choose from these days.

* [ActiveResource](https://github.com/rails/activeresource)
* [Flexirest](https://github.com/flexirest/flexirest/)
* [Her](https://github.com/remiprev/her)

Like I said, I've used ActiveResource and would be comfortable using it, but after reading the docs for Flexirest I get the impression it might be worth a shot. It uses Faraday so may have more flexibility for authentication, and also mentions support for `json:api` which might be worth considering for our v2 API.

misaka:
@benilovj Is this something we need to sort out in the backend vs here?

app/models/organisation.rb
dankmitchell: ActiveResource looks great to me Misha. Thanks for the heads up 👌

misaka: @dankmitchell Cool. Let's have a chat about this and how we're going to publish the API in the backend.

misaka: Did a little more reading and I like the idea of using [JSON:API](https://jsonapi.org/) to define our API. We've already had some questions about how to include objects in our JSON output and that's one thing that I think JSON:API can really help us with. Have a look at the `relationships` and `included` sections in the example of their main site. Flexirest appears to have some optimisations around associations it can take advantage of too.

Unfortunately, though, ActiveModel::Serializer doesn't support JSON:API. Judging by the README it looks like they're working towards supporting it for their 1.x release, but we're on 0.10.x. They do have recommendations for alternatives, [jsonape-rb](http://jsonapi-rb.org/) looks reasonable.

Also worth mentioning: [The Benefits of Using JSON API](https://nordicapis.com/the-benefits-of-using-json-api/).

misaka: What card is this for, btw? I'd like to put some of that stuff there ... and can we put that into the commits & branch names?

PR #52, opened 2019-02-15
[970] Fix webpack-dev-server and local CSS pipeline

tvararu:
### Context

Developing locally on this app is painful due to the fact that CSS gets recompiled on every request, and will only get more painful over time. Fixing this actually required a very involved investigation.

- Expose govuk-frontend `assets` path directly to the public
- Import main `application.scss` file using `application.js`; this seems awkward but is "normal" in frontendlandia where webpack heralds from
- Remove `WEBPACKER_DEV_SERVER_HOST` entries in `.env*` files, as these were causing Rails to not progressively enhance to `webpack-dev-server` when it was successfully booted up in parallel
- Remove `webpack-dev-server` entry from Procfile.dev, because foreman does not respect `.nvmrc` when choosing to run a specific version of node, and can blow up on a user's system if node is upgraded to current latest version
- Update the `webpack-dev-server` version to allow it to run using https for local development
- Fix some broken meta tags in the `<head>`
- Advise against using docker on macOS in the README due to speed
- Create a `webpacker` folder and delete the Rails `assets` folder

### Guidance to review

Best reviewed commit by commit.

### TODO

- [x] Fix Dan's comments the meta tags in `application.html.erb`
- [x] Fix the tests
- [x] Update the docker compose setup so it works correctly for local development
- [x] Move `application.scss` and delete `app/assets` folder so the setup feels less weird
- [x] Remove asset pipeline related code
- [x] Refactor the commits if they feel unclear or if there's too many


app/views/layouts/application.html.erb
```diff
@@ -6,16 +6,16 @@
     <%= csp_meta_tag %>
     <%= canonical_tag %>
     <%= tag :meta, property: "viewport", content: "width=device-width, initial-scale=1" %>
```
dankmitchell: Whilst you are here can you make this `name:` instead of `property:`. I was an error in the boilerplate.

app/views/layouts/application.html.erb
```diff
@@ -6,16 +6,16 @@
...
-    <%= tag :meta, property: "og:image", content: image_url("govuk-opengraph-image.png") %>
+    <%= tag :meta, property: "og:image", content: "/assets/images/govuk-opengraph-image.png" %>
     <%= tag :meta, property: "theme-color", content: "#0b0c0c" %>
```
dankmitchell: theme-color should also be `name` instead of `property`

benilovj:
How did these things slow down the asset pipeline? How many should be back-propagated into the boilerplate and support?

tvararu:
@benilovj see the commit messages, but the main saving was just compiling SCSS through the Rails asset pipeline which does it on every request (the entirety of govuk-frontend) instead of only when we actually make a change to the CSS. It's just switching over to webpack requires a lot of prodding due to assets.

This is still missing webpack-dev-server, which I want to push alongside bumping the version to address the spooky vulnerability message we have on the repo.

Once we figure this stuff out yeah we should port it to support and the boilerplate.

defong:
does that mean we can revisit this? https://github.com/DFE-Digital/manage-courses-support/pull/82

tvararu:
@defong hopefully yeah!

config/webpack/development.js
```diff
@@ -4,9 +4,9 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 
 const environment = require('./environment')
 
-// environment.config.devServer.https = {
-//   key: fs.readFileSync('./config/localhost/https/localhost.key'),
-//   cert: fs.readFileSync('./config/localhost/https/localhost.crt')
-// }
+environment.config.devServer.https = {
+  key: fs.readFileSync('./config/localhost/https/localhost.key'),
```
defong: Just wondering can those string value be passed from yml, or have the yml comment to state its overridden here

tvararu: It doesn't work:

```
/Users/tvararu/dfe/manage-courses-frontend/node_modules/webpack-dev-server/bin/webpack-dev-server.js:403
    throw e;
    ^

Error: error:0906D06C:PEM routines:PEM_read_bio:no start line
    at Object.createSecureContext (_tls_common.js:113:17)
    at Server (_tls_wrap.js:870:27)
    at Server (https.js:62:14)
    at Server._init (/Users/tvararu/dfe/manage-courses-frontend/node_modules/spdy/lib/spdy/server.js:63:10)
    at new Server (/Users/tvararu/dfe/manage-courses-frontend/node_modules/spdy/lib/spdy/server.js:26:10)
    at Function.create (/Users/tvararu/dfe/manage-courses-frontend/node_modules/spdy/lib/spdy/server.js:31:12)
    at Object.create [as createServer] (/Users/tvararu/dfe/manage-courses-frontend/node_modules/spdy/lib/spdy/server.js:287:27)
    at new Server (/Users/tvararu/dfe/manage-courses-frontend/node_modules/webpack-dev-server/lib/Server.js:484:30)
    at startDevServer (/Users/tvararu/dfe/manage-courses-frontend/node_modules/webpack-dev-server/bin/webpack-dev-server.js:396:14)
    at processOptions (/Users/tvararu/dfe/manage-courses-frontend/node_modules/webpack-dev-server/bin/webpack-dev-server.js:360:5)
```

defong:
```
environment.config.devServer.https = {

 key: fs.readFileSync(environment.config.devServer.https.custom.key),

```

sorry i meant something like that...

tvararu: Ah, hmm... I'll give it a go, but first I would like it to just work. That gave me a last resort idea to maybe fix it. 💡

tvararu: Right, looked at this for a bit but it's buggy and weird, just leaving it as is for now.

tvararu:
Docker takes 40 seconds to compile webpack. It's use for local development is discouraged.

defong:
> Docker takes 40 seconds to compile webpack. It's use for local development is discouraged.

Only on mac

tvararu:
[Implement manage courses table in Rails frontend](https://trello.com/c/RIhSiYT3/970-implement-manage-courses-table-in-rails-frontend)

PR #59, opened 2019-02-25
Add new ENV for c# url

dankmitchell:
### Context
A number of features will continue to live in the current `manage-courses-ui` project hosted on a different domain so we need a way to keep the UX seamless as a user clicks around the app.

Add environment variables so we can specify the different URL's i.e. dev. and staging.


dankmitchell:
[Configuration for linking from Rails Publish to C# Publish](https://trello.com/c/9Q3z9OH3/1016-configuration-for-linking-from-rails-publish-to-c-publish)

defong:
Its looks fine to me....

Thro `settings gem`?

PR #66, opened 2019-02-28
[1020] Added mechianism for interacting with backend session creation

defong:
### Context
The frontend and backend must be in enforce with authentication and session.

External authentication needs to be funnelled to the backend to ensure authenticity of the user, the backend needs to have authenticated trusted client calls as well.

This needs to work in accordance with https://trello.com/c/gHNepiOF/993-create-new-user-sessions-in-api-backend.

Make it work

### Guidance to review

The configuration for both frontend and backend needs to share the same credentials, for example see the below.

settings.yml
```
authentication:
  algorithm: HS256
  secret: lets-keep-my-fav-dan-super-happy
```


app/controllers/sessions_controller.rb
```diff
@@ -5,6 +5,13 @@ def new
 
   def create
     session[:auth_user] = auth_hash
+
+    user_info = request.env["omniauth.auth"][:info]
```
dankmitchell: This could be shortened to `auth_hash[:info]`

defong:
[Create new user sessions in frontend](https://trello.com/c/TcgULXhH/1020-create-new-user-sessions-in-frontend)

spec/support/feature_helpers.rb
```diff
@@ -6,13 +6,18 @@ def stub_omniauth
...
+        email: "email@example.com"
       },
       credentials: {
         token_id: "123"
       }
     }
   end
+
+  def stub_session_create
+    allow(Session).to receive(:create).and_return(something: 'testing')
```
misaka: Shouldn't this be something that has an `id` method to simulate how it's actually used?

```
OpenStruct.new(id: 1).id #=> 1
```

or, _I think_, which might be better if it works:

```
double(id: 1).id #=> 1
```

spec/features/sign_in_spec.rb
```diff
@@ -3,7 +3,10 @@
 RSpec.feature 'Sign in', type: :feature do
   scenario 'using DfE Sign-in' do
     stub_omniauth
+    stub_session_create
+
     visit root_path
+
     # Redirect to DfE Signin and come back
```
misaka: So now you can do:

```
expect(Session).to have_received(:create).with(first_name: "...", last_name: "...")
```

app/controllers/sessions_controller.rb
```diff
@@ -5,6 +5,11 @@ def new
 
   def create
     session[:auth_user] = auth_hash
+
+    set_connection
+
+    session[:auth_user].user_id = Session.create(first_name: current_user_info[:first_name], last_name: current_user_info[:last_name]).id
```
misaka: Can we clean this up to be:

```
     session[:auth_user].user_id = Session.create(
       first_name: current_user_info[:first_name],
       last_name: current_user_info[:last_name]
     ).id 
```

Or maybe even use an intermediary variable (because pulling `id` out isn't too obvious).

spec/features/sign_in_spec.rb
defong: you a :star:

app/controllers/sessions_controller.rb
defong: went for variable

spec/support/feature_helpers.rb
defong: nice one

app/controllers/application_controller.rb
```diff
@@ -16,6 +16,35 @@ def current_user
...
+    end
+  end
+
+  def current_user_info
+    current_user['info']
+  end
+
+private
+
+  def set_connection
```
tvararu: `def add_token_to_connections`?

tvararu:
[Organisation details links to C# Publish](https://trello.com/c/aQH97fdv/1019-organisation-details-links-to-c-publish)

app/controllers/application_controller.rb
benilovj: 👍 to @tvararu's suggestion

defong: going with `add_token_to_connection` as pluralised has no meaning

PR #69, opened 2019-03-03
Add failOnStderr property to linting task

chelnak:
### Context
@defong pointed out that even though linters were failing the build was still passing.

This PR adds failOnStderr to the linting task so that the build will fail in the correct place

Add failOnStderr propert to linting task

### Guidance to review
Does this fail builds as expected?

### DIFF
--- a/docker-build-az.sh
+++ b/docker-build-az.sh
@@ -1,29 +1,30 @@
 #!/bin/bash
+# Build, test and push a custom docker image
 set -e
 
-# DEBUG: print env
+if [ $SYSTEM_DEBUG="true" ];
+then
+  env
+fi
 
-env
-
-# Set up script vars
 DOCKER_IMAGE_TAG=$TAG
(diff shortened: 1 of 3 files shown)


misaka:
I see that the mechanism to run the tests has been moved into a script, but I don't see where `failOnStderr` is enabled or anything. Is that a change done in the GUI?

Pretty annoying that `govuk-lint-ruby` doesn't return an error, sounds like that should also be fixed upstream.

docker-build-az.sh
```diff
@@ -0,0 +1,30 @@
+#!/bin/bash
+# Build, test and push a custom docker image
+set -e
+
+if [ "$SYSTEM_DEBUG" = 'true' ];
```
misaka: Don't need the `;` here unless `true` is on the same line.

misaka: > Don't need the `;` here unless `true` is on the same line.

`then` not `true`

chelnak:
PR is now redundant

PR #73, opened 2019-03-06
[1067] Amended the session to expire in 5 minutes

defong:
### Context
Previous default was when user close the browser

Amended the session to expire in 5 minutes

### Guidance to review
Sign in stay inactivity for 5+ minutes then traverse the site
When traversing the site browser first go to sign in and sign in via `openconnect` seamlessly.

After a period of time has elapse the user will not sign in via `openconnect` seamlessly, instead they will need to reenter their credentials.

### DIFF
--- a/config/application.rb
+++ b/config/application.rb
@@ -24,6 +24,8 @@ module ManageCoursesFrontend
     # -- all .rb files in that directory are automatically loaded after loading
     # the framework and any gems in your application.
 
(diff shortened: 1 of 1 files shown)


defong:
[Investigate session timeout in manage-courses-frontend](https://trello.com/c/j82EowOD/1067-investigate-session-timeout-in-manage-courses-frontend)

PR #77, opened 2019-03-06
Add course count to org list

dankmitchell:
### Context
Replicate https://bat-dev-manage-courses-ui-app.azurewebsites.net/organisations

Add course count below each org name

### Guidance to review
`/organisations`


app/models/provider.rb
```diff
@@ -1,2 +1,5 @@
 class Provider < Base
+  def course_count
+    relationships.courses[:meta][:count]
```
misaka: Got a spec?

PR #78, opened 2019-03-06
[1096] Add org show

dankmitchell:
### Context
Replica of the current c# ui 🤓 

Add org show view

### Guidance to review
screenshot 2019-03-06 at 15 12 22

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -4,4 +4,8 @@ class ProvidersController < ApplicationController
   def index
     @providers = Provider.all
   end
+
+  def show
+    @provider = Provider.where(institution_code: params[:id]).first
+  end
 end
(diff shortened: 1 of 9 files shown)


misaka:
Just noticed that the `course_count` method on `Provider` isn't tested, and also the view that uses it. Seems like we're missing request/feature specs here.

app/controllers/providers_controller.rb
```diff
@@ -4,4 +4,8 @@ class ProvidersController < ApplicationController
   def index
     @providers = Provider.all
   end
+
+  def show
+    @provider = Provider.where(institution_code: params[:id]).first
```
defong: Provider.find(institution_code: param[:id])

?

defong:
[Add org show view](https://trello.com/c/eWCanfEJ/1096-add-org-show-view)

app/controllers/providers_controller.rb
dankmitchell: You can't use `find` when looking up anything other than the `id` and the `json_api_client` doesnt support `find_by`

PR #79, opened 2019-03-06
[1068] Copy assets to public folder using webpack

tvararu:
This should re-enable the fonts and other contents of govuk-frontend.

This was initally a WIP with a worse solution, I'm much happier with the webpack copy method than re-enabling the asset pipeline compile flag in production mode.

### DIFF
--- a/config/webpack/production.js
+++ b/config/webpack/production.js
@@ -1,5 +1,13 @@
 process.env.NODE_ENV = process.env.NODE_ENV || 'production'
+const CopyWebpackPlugin = require('copy-webpack-plugin');
 
 const environment = require('./environment')
+
+environment.config.plugins = (environment.config.plugins || []).concat([
+  new CopyWebpackPlugin([
+    { from: 'node_modules/govuk-frontend/assets/fonts/**/*', to: '../assets/fonts/', flatten: true },
+    { from: 'node_modules/govuk-frontend/assets/images/**/*', to: '../assets/images/', flatten: true }
+  ])
+]);
 
 module.exports = environment.toWebpackConfig()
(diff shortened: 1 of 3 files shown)


tvararu:
[Missing assets on Manage Courses Frontend](https://trello.com/c/vdwDHuai/1068-missing-assets-on-manage-courses-frontend)

tvararu:
Using `git bisect` (which I'm starting to really like), I've determined this regression was introduced in 6060496e0b0310ae60a95323e7b046ba609dd884. I didn't notice it.

tvararu:
@dankmitchell import into the JS pack file? So like:

```js
// webpacker/packs/application.js
import "../stylesheets/application.scss";
import { initAll } from "govuk-frontend";
import CookieMessage from "scripts/cookie-banner";

import 'govuk-frontend/images/crest.png';
// ...
import 'govuk-frontend/images/font.woff';
// ...
```

There's a few tricky bits I don't like about this:

- It's an explicit listing of everything in govuk-frontend assets in our codebase, and we don't explicitly reference everything ourselves; the coupling for quite a bit is handled by `govuk-frontend-all.css`, they specify the fonts and such and we can pretend they don't even exist as long as we make them accessible in `public` somehow
- The fonts and some assets have fingerprinting like `bold-a2452cb66f-v1.woff2`. When we upgrade govuk-frontend, they're likely to bump that fingerprint, and we'll have to update all our references in our JS

Simply copying all the assets wholesale and not using `asset_pack_url` gets around this. The disadvantage is then:

- Rails won't do its own fingerprinting, but govuk-frontend already does it for fonts, and most assets like the crest are super unlikely to ever change

Happy to change my mind on this if you have a different view. :)

tvararu:
Have confirmed by building a local docker container and running `ls public/assets/fonts` that this definitely should copy over the assets when the existing `rake assets:precompile` command in the Dockerfile runs. :shipit:

PR #81, opened 2019-03-07
[960] Add missing providers spec

dankmitchell:
### Context
https://github.com/DFE-Digital/manage-courses-frontend/pull/49

Add missing spec

### Guidance to review
`rake spec` or `rspec`

### DIFF
--- a/spec/features/providers_spec.rb
+++ b/spec/features/providers_spec.rb
@@ -1,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.feature 'View providers', type: :feature do
+  scenario 'Navigate to /organisations' do
+    stub_omniauth
+    stub_session_create
+    stub_backend_api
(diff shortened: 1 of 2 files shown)


spec/features/providers_spec.rb
```diff
@@ -0,0 +1,13 @@
+require 'rails_helper'
+
+RSpec.feature 'View providers', type: :feature do
+  scenario 'Navigate to /organisations' do
+    stub_omniauth
+    stub_session_create
+    stub_backend_api
```
defong: A little bit tied down, but I guess when we work more on it it will evolve better

dankmitchell: Exactly @defong. Going to catch up with @misaka today on a testing strategy for this app but this is fine for now.

PR #83, opened 2019-03-07
[1108] Handle unknown users

benilovj:
### Context
It's possible for an existing DfE Sign-in user to try accessing Publish, but they're not a known user in the Manage Courses database.

This is a temporary stop-gap which redirects unknown users to Manage Courses UI.
Manage Courses UI then shows a "we don't know who you are" page to those users.

### Guidance to review
I paired with @defong on this.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -8,10 +8,14 @@ class SessionsController < ApplicationController
 
     add_token_to_connection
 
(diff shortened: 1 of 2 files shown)


defong:
[Handle unknown users in Manage Courses Frontend](https://trello.com/c/ztqQdiBD/1108-handle-unknown-users-in-manage-courses-frontend)

PR #85, opened 2019-03-07
[1112] extend session timeout to 6 hours

misaka:
We tried 5 minutes to see how sign-in would behave but ran afoul of it's ~20-30
minute session expiry. We need to implement access token renewal with the
refresh token, but until we do that (hopefully before go-live) we'll extend the
session to 6 hours here.


defong:
[Extend mc-fe session timeout to 6 hours](https://trello.com/c/7PAXIA9x/1112-extend-mc-fe-session-timeout-to-6-hours)

PR #87, opened 2019-03-08
Refactor pages spec

dankmitchell:
### Context
`pages#show` was from the boilerplate, we don't need that

- Be explicit with our controller actions to ensure they match our routes
- Add specs for all `pages`
- Add temp override for authentication because we always redirect to root after success which needs updating at a later date

### DIFF
--- a/spec/features/view_pages_spec.rb
+++ b/spec/features/view_pages_spec.rb
@@ -1,13 +1,27 @@
 require 'rails_helper'
 
 RSpec.feature 'View pages', type: :feature do
-  scenario "Navigate to home" do
+  scenario "Navigate to /cookies" do
     stub_omniauth
     stub_session_create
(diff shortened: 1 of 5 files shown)


spec/features/view_pages_spec.rb
```diff
@@ -1,13 +1,27 @@
 require 'rails_helper'
 
 RSpec.feature 'View pages', type: :feature do
-  scenario "Navigate to home" do
+  scenario "Navigate to /cookles" do
```
tvararu: Cookies

davidgisbey:
[Refactor controller spec matcher logic into spec helper](https://trello.com/c/M5O2d71f/1101-refactor-controller-spec-matcher-logic-into-spec-helper)

PR #88, opened 2019-03-08
[1115] Add turbolinks

dankmitchell:
### Context
🚀 

- Add turbolinks


dankmitchell:
Yes please @tvararu

defong:
[Add turbolinks to speed up page loads time](https://trello.com/c/bqHNATra/1115-add-turbolinks-to-speed-up-page-loads-time)

tvararu:
I've done some testing in Chrome and Safari using VoiceOver on MacOS.

There are differences in the behaviour of the site with and without Turbolinks.

When users navigate without Turbolinks, the focus between pages is reset to the top, usually the page title is announced, and the first element a user will encounter is the "skip to main content link". Clicking will take them to the first header.

When users navigate with Turbolinks, the focus usually moves to the **element before `<main>`** in body. So in the case of going from the orgs list to an individual org, the focus will be on the end of `<nav>`. In the case of going from an org to the orgs list, the focus will be at the start of body, before `<main>`. When users move to the next element, they move into `<main>` and then on to the `<h1>`. Very importantly, the new page title does get announced, like without Turbolinks.

In practice, this means that instead of landing on the skip link, users land somewhere in the middle of the page. I don't *think* this is a barrier, as they are usually 1 "next element" VoiceOver command from reaching the h1 and carrying on as before. But it is different. @fofr requested that we validate this with user research.

I think we should park this for now, and test with some users first. :)

PR #89, opened 2019-03-08
[1108] Rebuild the connection when adding session token

tvararu:
Without the `(true)` parameter, Base.connection will re-use the same instance of the connection class between connections. If multiple users are hitting the app simultaneously, this can cause re-use of tokens among them and undesirable behaviours.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -43,7 +43,7 @@ private
                           Settings.authentication.algorithm)
     end
 
-    Base.connection do |connection|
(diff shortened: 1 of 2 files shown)


defong:
[Handle unknown users in Manage Courses Frontend](https://trello.com/c/ztqQdiBD/1108-handle-unknown-users-in-manage-courses-frontend)

defong:
> Looks legit

Our code always looks legit and bananas

dankmitchell:
@defong that can’t be said for your c# code. That’s just 🍌

PR #91, opened 2019-03-11
[1124] Fixed issues whereby the user has circumvented session create hence the user is half authenicated

defong:
### Context
issues whereby the user has circumvented session create hence the user is half authenticated

Fixed issues whereby the user has circumvented session create hence the user is half authenticated


defong:
[Manually accessing manage-courses-frontend as a user without access to any orgs leads to a 500](https://trello.com/c/TV5N68MK/1124-manually-accessing-manage-courses-frontend-as-a-user-without-access-to-any-orgs-leads-to-a-500)

PR #92, opened 2019-03-13
[1139] Redirect to manage-courses-ui when providers list is empty

tvararu:
Users should be redirected to the "We don’t know which organisation you’re part of" page in manage-courses-ui when the providers list is empty.

### DIFF
--- a/spec/controllers/providers_controller_spec.rb
+++ b/spec/controllers/providers_controller_spec.rb
@@ -1,0 +1,34 @@
+require 'rails_helper'
+
+RSpec.describe ProvidersController, type: :controller do
+  context "with authenticated user" do
+    before do
+      stub_omniauth
+      stub_session_create
+    end
+
+    describe 'GET #index' do
+      context 'with providers' do
+        before do
+          stub_api_v2_request('/providers', 'providers.json')
+        end
+
+        it 'returns the index page' do
+          get :index
+          expect(response).to have_http_status(:success)
+        end
+      end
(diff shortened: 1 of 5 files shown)


spec/controllers/providers_controller_spec.rb
```diff
@@ -0,0 +1,34 @@
...
+
+        it 'returns the index page' do
+          get :index
+          expect(response).to have_http_status(:success)
+        end
+      end
+
+      fcontext 'without any providers' do
+        before do
+          stub_api_v2_request('/providers', 'providers-empty.json')
```
benilovj: I'm not wild about using fixtures, could we explore using `factory_bot` to create stubs? e.g. [like this](https://github.com/alphagov/specialist-publisher/blob/master/spec/fixtures/factories.rb#L67)

misaka: Here's some documentation from FactoryBot on how to build a [custom JSON strategy](https://github.com/thoughtbot/factory_bot/blob/master/GETTING_STARTED.md#custom-strategies).

dankmitchell: A nice blog from thoughtbot https://thoughtbot.com/blog/tips-for-using-factory-girl-without-an-orm

dankmitchell: I purpose we ship this first then refactor the fixtures into factorybot

PR #93, opened 2019-03-13
Test and fixture improvements

tvararu:

- Enable focusing in rspec
- Add factory_bot and use factories for provider fixtures in tests

### Guidance to review

🚢


spec/fixtures/factories.rb
```diff
@@ -0,0 +1,33 @@
+FactoryBot.define do
+  factory :provider, class: Hash do
```
misaka: Can we move this factory to `spec/factories/providers.rb`.

spec/fixtures/factories.rb
```diff
@@ -0,0 +1,33 @@
...
+            "meta" => {
+              "count" => 1
+            }
+          }
+        }
+      }
+    end
+  end
+
+  factory :providers_response, class: Hash do
```
misaka: Can we move this factory to `spec/factories/provider_responses.rb`.

spec/controllers/providers_controller_spec.rb
```diff
@@ -10,7 +10,7 @@
     describe 'GET #index' do
       context 'with providers' do
         before do
-          stub_api_v2_request('/providers', 'providers.json')
+          stub_api_v2_request('/providers', FactoryBot.build(:providers_response))
```
misaka: Can we use `build` without the `FactoryBot`?

spec/fixtures/factories.rb
misaka: nm, this belongs in `providers` factories.

PR #94, opened 2019-03-13
[1146] Redirect to organisations#show page when there is 1 provider

tvararu:
### Context

Redirects the index action if there's only one provider.

The breadcrumbs are also hidden because they are not useful in this scenario. This is done by querying the full Providers endpoint for the full list, which is a bit annoying/slow, but can be changed in the future if necessary.

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -4,9 +4,11 @@ class ProvidersController < ApplicationController
   def index
     @providers = Provider.all
     render_manage_ui if @providers.empty?
+    redirect_to provider_path(@providers.first.institution_code) if @providers.size == 1
   end
 
   def show
+    @has_multiple_providers = Provider.all.size > 1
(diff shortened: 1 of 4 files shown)


app/controllers/providers_controller.rb
```diff
@@ -4,9 +4,11 @@ class ProvidersController < ApplicationController
   def index
     @providers = Provider.all
     render_manage_ui if @providers.empty?
+    redirect_to controller: 'providers', action: 'show', code: @providers.first.institution_code if @providers.size == 1
```
dankmitchell: `redirect_to provider_path(@providers.first) if @providers.size == 1` ?

misaka: Might also be able to:

```
redirect_to @providers.first if @providers.size == 1
```

tvararu: yes

PR #96, opened 2019-03-14
[1136] setup course vacancies view

dankmitchell:
### Context
`providers#course#vacancies`

- Add course vacancies edit view

# After
Screenshot 2019-03-14 at 15 32 42

### DIFF
--- a/spec/factories/courses.rb
+++ b/spec/factories/courses.rb
@@ -1,0 +1,33 @@
+FactoryBot.define do
+  factory :course, class: Hash do
+    sequence(:course_code) { |n| "X10#{n}" }
+    name { "English" }
+
+    initialize_with do
+      {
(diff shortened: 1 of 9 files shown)


spec/factories/courses.rb
```diff
@@ -0,0 +1,33 @@
+FactoryBot.define do
+  factory :course, class: Hash do
+    sequence(:course_code) { |n| "X10#{n}" }
```
tvararu: This could result in course codes which are longer than 4 characters but it shouldn't break anything so it's fine, minor

defong: https://github.com/DFE-Digital/manage-courses-backend/blob/7847be1c5e1564f68ed931653f0372c86f2f05a2/db/seeds.rb#L40

PR #100, opened 2019-03-19
1136 add vacancies form with sites

dankmitchell:
### Context
Course vacancies

Add course vacancies scaffolding and form

### Guidance to review
i.e. https://localhost:3000/organisations/2KA/courses/3DDV/vacancies for no vacancies or https://localhost:3000/organisations/2KA/courses/3C6B/vacancies for several sites and vacancies

Screenshot 2019-03-19 at 12 22 53

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -1,3 +1,13 @@
 class Course < Base
-  belongs_to :provider
+  belongs_to :provider, param: :provider_code
+  has_many :site_statuses
+  has_many :sites, through: :site_statuses, source: :site
+
+  def full_time?
+    study_mode == 'full time'
+  end
+
+  def part_time?
+    study_mode == 'part time'
+  end
 end
--- a/app/views/courses/vacancies.html.erb
+++ b/app/views/courses/vacancies.html.erb
@@ -1,6 +1,6 @@
 <%= content_for :page_title, "Vacancies" %>
 
-<%= manage_ui_link_to "Back", "/organisation/#{params[:provider_code]}/courses/#{@course.course_code}", class: "govuk-back-link" %>
+<%= manage_ui_link_to_back("/organisation/#{params[:provider_code]}/courses/#{@course.course_code}") %>
 
 <main role="main" class="govuk-main-wrapper" id="main-content">
   <div class="govuk-grid-row">
--- a/app/views/courses/vacancies.html.erb
+++ b/app/views/courses/vacancies.html.erb
@@ -1,10 +1,61 @@
 <%= content_for :page_title, "Vacancies" %>
 
-<%= manage_ui_link_to "Back", "/organisation/#{params[:provider_code]}/courses/#{@course[:course_code]}", class: "govuk-back-link" %>
+<%= manage_ui_link_to "Back", "/organisation/#{params[:provider_code]}/courses/#{@course.course_code}", class: "govuk-back-link" %>
 
 <main role="main" class="govuk-main-wrapper" id="main-content">
-  <h1 class="govuk-heading-xl">
-    <span class="govuk-caption-xl"><%= @course[:name] %> (<%= @course[:course_code] %>)</span>
-    Edit vacancies
-  </h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= form_with url: provider_courses_path(@course.course_code) do |f| %>
+        <div class="govuk-form-group">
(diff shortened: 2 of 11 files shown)


app/views/courses/vacancies.html.erb
```diff
@@ -1,10 +1,61 @@
 <%= content_for :page_title, "Vacancies" %>
 
-<%= manage_ui_link_to "Back", "/organisation/#{params[:provider_code]}/courses/#{@course[:course_code]}", class: "govuk-back-link" %>
+<%= manage_ui_link_to "Back", "/organisation/#{params[:provider_code]}/courses/#{@course.course_code}", class: "govuk-back-link" %>
```
benilovj: maybe we should create another link helper that calls `manage_ui_link_to`?

dankmitchell: Not a bad shout 👌

app/models/course.rb
```diff
@@ -1,3 +1,13 @@
 class Course < Base
   belongs_to :provider
```
misaka: If we make this declaration:

```
belongs_to :provider, param: :provider_code
```

We'll be able to use it like this:

```
Course.where(provider_code: params[:provider_code]).find(...)
```

Which is a bit nicer than using the misnomer `id`.

fofr:
:100:

app/views/courses/vacancies.html.erb
```diff
@@ -1,10 +1,61 @@
...
 <main role="main" class="govuk-main-wrapper" id="main-content">
-  <h1 class="govuk-heading-xl">
-    <span class="govuk-caption-xl"><%= @course[:name] %> (<%= @course[:course_code] %>)</span>
-    Edit vacancies
-  </h1>
+  <div class="govuk-grid-row">
+    <div class="govuk-grid-column-two-thirds">
+      <%= form_with url: provider_courses_path(@course.course_code) do |f| %>
+        <div class="govuk-form-group">
+          <fieldset class="govuk-fieldset" aria-describedby="how-contacted-conditional-hint">
```
misaka: Is there a reason we don't use the GovukElementsFormBuilder?

dankmitchell: It doesn't support govuk-frontend, only the legacy govuk-elements

misaka:
Implementation looks good (why oh why isn't there a builder for `govuk-frontend`??), but definitely needs some tests.

dankmitchell:
> Implementation looks good (why oh why isn't there a builder for `govuk-frontend`??), but definitely needs some tests.

@misaka The design system team don't support erb just nunjucks :(

app/models/course.rb
```diff
@@ -1,3 +1,13 @@
...
+  belongs_to :provider, param: :provider_code
+  has_many :site_statuses
+  has_many :sites, through: :site_statuses, source: :site
+
+  def full_time?
+    study_mode == 'full time'
+  end
+
+  def part_time?
+    study_mode == 'part time'
```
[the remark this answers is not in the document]
misaka: @defong I think we're using what's returned by `study_mode_description` as in the API course model. But fair point.

@dankmitchell This reminds me, doesn't this change rely on a change to `mcb`? To pull out `study_mode_description`? I wonder if we can't expose both `study_mode` (and not just `study_mode_before_type_cast` like it is as present) and `study_mode_description` in the API, which means we would be able to use `study_mode` here.

app/views/courses/vacancies.html.erb
```diff
@@ -1,10 +1,61 @@
...
+              </h1>
+            </legend>
+            <div class="govuk-radios govuk-!-margin-bottom-7" data-module="radios">
+              <div class="govuk-radios__item">
+                <%= f.radio_button :has_vacancies, 'false', class: 'govuk-radios__input', checked: !@course.has_vacancies? %>
+                <%= f.label :has_vacancies_false, 'There are no vacancies', class: "govuk-label govuk-radios__label govuk-!-font-weight-bold" %>
+                <span id="has_vacancies_false" class="govuk-hint govuk-radios__hint">
+                  Close this course to new applications.
+                  <br>
+                  You can reopen a course later
```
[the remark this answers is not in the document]
defong: image

heidar:
> > Implementation looks good (why oh why isn't there a builder for `govuk-frontend`??), but definitely needs some tests.
> 
> @misaka The design system team don't support erb just nunjucks :(

This repo has form helpers for all govuk forms https://github.com/ministryofjustice/laa-apply-for-legal-aid

PR #101, opened 2019-03-19
[1165] Redirect to manage-courses-ui when API returns 403

tvararu:
### Context

When a new user logs into the service, they should be asked to tick the terms and conditions if they haven't done so already.


tvararu:
[New publishers should tick Ts & Cs](https://trello.com/c/z1KwaS8B/1165-new-publishers-should-tick-ts-cs)

PR #104, opened 2019-03-20
[1194] Add provider count cookie

tvararu:
### Context

We currently do a request for the whole list of providers when we hit the `organisation#show` page. This is fairly inefficient, and moreover, we'll need this in the future on every page that has breadcrumbs (almost all of them).

When initialising a user session, query the list of providers in order to get the total count and save it to a cookie. This can then be referenced from the controller/view templates in order to decide whether we need to show the breadcrumbs.

Note that this call in the session is still quite slow (for a God user), so it slows down initial signin, but it's only done once. We can choose to further optimise it down the line.

### Guidance to review

This is a spike more or less, so haven't touched the tests, but if the idea looks good, just request changes and I'll add/update the requisite tests. :)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,7 +25,7 @@ class ApplicationController < ActionController::Base
     if current_user
       add_token_to_connection
 
-      set_user_session if current_user[:user_id].blank?
+      set_user_session if current_user['user_id'].blank?
     else
       redirect_to '/signin'
     end
@@ -39,7 +39,7 @@ private
 
   def set_user_session
     user = Session.create(first_name: current_user_info[:first_name], last_name: current_user_info[:last_name])
-    session[:auth_user][:user_id] = user.id
+    session[:auth_user]['user_id'] = user.id
 
     add_provider_count_cookie
   end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -40,6 +40,16 @@ private
   def set_user_session
     user = Session.create(first_name: current_user_info[:first_name], last_name: current_user_info[:last_name])
     session[:auth_user][:user_id] = user.id
+
+    add_provider_count_cookie
+  end
+
+  def add_provider_count_cookie
+    begin
+      session[:auth_user][:provider_count] = Provider.all.size
+    rescue StandardError => e
+      logger.error "Error setting the provider_count cookie: #{e.class.name}, #{e.message}"
+    end
   end
 
   def add_token_to_connection
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -8,7 +8,8 @@ class ProvidersController < ApplicationController
   end
 
   def show
-    @has_multiple_providers = Provider.all.size > 1
+    provider_count = session.try(:[], :auth_user).try(:[], :provider_count)
+    @has_multiple_providers = provider_count.nil? || provider_count > 1
     @provider = Provider.find(params[:code]).first.attributes
   end
 end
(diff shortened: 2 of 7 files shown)


tvararu:
@dankmitchell the motivation is that I didn't know I could just chuck it in there... I'll update.

tvararu:
[Optimise provider count lookup in manage-courses-frontend](https://trello.com/c/hl4GfE7S/1194-optimise-provider-count-lookup-in-manage-courses-frontend)

app/controllers/application_controller.rb
```diff
@@ -40,6 +40,16 @@ def current_user_info
...
+
+    add_provider_count_cookie
+  end
+
+  def add_provider_count_cookie
+    begin
+      session[:auth_user][:provider_count] = Provider.all.size
+    rescue => e
+      logger.error "Error setting the provider_count cookie: #{e.class.name}, #{e.message}"
+    end
```
misaka: Tests?

misaka:
Looking good!

app/controllers/application_controller.rb
tvararu: Have added some, give it a look 👀

app/controllers/providers_controller.rb
```diff
@@ -8,7 +8,7 @@ def index
   end
 
   def show
-    @has_multiple_providers = Provider.all.size > 1
+    @has_multiple_providers = session.to_hash.dig(:auth_user, :provider_count).to_i > 1
```
tvararu: 🤓 http://mitrev.net/ruby/2015/11/13/the-operator-in-ruby/#arraydig-and-hashdig

benilovj: `nil.to_i # => 0`. is that the right default if the thing isn't set for some reason?

tvararu: Good question, maybe not... rather always show the breadcrumbs I think as not having them is more harmful than having useless ones?

benilovj: agree that "show if in doubt" is safer

tvararu: Updated

tvararu:
I can't test this locally because I can't run `yarn` on this computer... but it should work, I have tested it yesterday (before some iteration) on my other Mac.

spec/controllers/application_controller_spec.rb
```diff
@@ -51,6 +53,7 @@
         it "has set connection" do expect(Base).to have_received(:connection) end
         it "has not called session create" do expect(Session).to_not have_received(:create) end
         it "has set user_id" do expect(controller.request.session[:auth_user][:user_id]).to eq user_id end
+        it "has set provider_count" do expect(controller.request.session[:auth_user][:provider_count]).to eq nil end
```
misaka: These blocks should use braces (`{}`). How did the lines above get in ...

tvararu: 🙈 🙉 🙊

misaka: You let him put those in!!

misaka: Actually, I take it back about using `{}`s ... these should be multi-line blocks.

app/controllers/providers_controller.rb
tvararu: This whole `dig` thing actually ended up working incorrectly so I swapped it to `.try`...

tvararu:
@misaka done

misaka:
🚀

app/controllers/application_controller.rb
```diff
@@ -25,7 +25,7 @@ def authenticate
     if current_user
       add_token_to_connection
 
-      set_user_session if current_user[:user_id].blank?
+      set_user_session if current_user['user_id'].blank?
```
dankmitchell: Using a symbol is faster than a string because it gets cached

tvararu: Yeah but it doesn't work... ❓

dankmitchell: Fair point 😂

PR #106, opened 2019-03-25
1136 add vacancies form

dankmitchell:
### Context

Editing course vacancies.

Edit course vacancy form. One checkbox per site status, two if it's both full and part time.

### Guidance to review

Some of this is WIP. A couple of the tests need refactoring. Some classes are missing tests. These will be fixed in a followup PR.

Validation is also missing and will also be done in a followup PR.

Currently this redirects back to the form itself with no feedback. This needs to redirect to the C# app, which still needs implementing.

Finally, we have not implemented the radio button's functionality. Only the checkboxes work.


dankmitchell:
[Editing vacancies: multiple sites/both full-time and part-time](https://trello.com/c/rru4w3BN/1136-editing-vacancies-multiple-sites-both-full-time-and-part-time)

app/controllers/courses/vacancies_controller.rb
```diff
@@ -0,0 +1,45 @@
...
+      :authenticate,
+      :build_course,
+      :build_site_statuses,
+      :build_first_course
+    )
+
+    def edit; end
+
+    def update
+      params.dig(:course, :site_status_attributes)
```
misaka: Minor thing. If `govuk-lint-ruby` will let you, would it be worth indenting this code here so that it's more obviously a block?

heidar:
@misaka Thanks for the review, have fixed the indentation on that block. 👍

app/helpers/vacancy_helper.rb
```diff
@@ -0,0 +1,22 @@
...
+        true
+      elsif site_status.full_time_and_part_time_vacancies?
+        true
+      else
+        false
+      end
+    else
+      false
+    end
+  end
```
misaka: Honest question, but would this be more readable as case statement:

```
    case course.study_mode
    when :full_time
      site_status.full_time_vacancies?
    when :part_time 
      site_status.part_time_vacancies?
    when course.full_time_or_part_time?
      case vacancy_study_mode
      when :part_time
        site_status.part_time_vacancies?
      when :full_time
        site_status.full_time_vacancies?
      when full_time_and_part_time_vacancies?
        true
      else
        false
      end
    else
      false
    end
```

app/services/vacancy_status_determination_service.rb
```diff
@@ -0,0 +1,47 @@
...
+  attr_reader :vacancy_status_full_time,
+              :vacancy_status_part_time,
+              :course
+
+  def self.call(vacancy_status_full_time:, vacancy_status_part_time:, course:)
+    new(
+      vacancy_status_full_time: vacancy_status_full_time,
+      vacancy_status_part_time: vacancy_status_part_time,
+      course:                   course
+    ).call
```
misaka: 👍

app/helpers/vacancy_helper.rb
heidar: it has a todo to refactor so we can ship this

spec/features/courses/vacancies/edit_spec.rb
```diff
@@ -0,0 +1,243 @@
...
+          type: 'site_statuses',
+          attributes: {
+            vac_status: 'no_vacancies'
+          },
+          relationships: {
+            site: {
+              data: {
+                type: 'sites',
+                id: '3'
+              }
```
misaka: All the `SiteStatus` objects here have the same `Site`. That doesn't quite sound right to me, is it? If not, should we bother to make this better represent what's actually real data?

In a future refactor maybe?

PR #113, opened 2019-03-28
Fix courses vacancies edit

dankmitchell:
### Context
The attributes on the checkboxes changed in this big PR https://github.com/DFE-Digital/manage-courses-frontend/pull/106 and this change missed.

The happy path is working but adding vacancies for a course where the vacancy status is nil isn't working. This should fix it

Update checkbox input attributes in the controller

### Guidance to review
Example course `https://localhost:3000/organisations/2DQ/courses/3272/vacancies`

The test here doesn't actually test the failure but it's good to add. Once we have proper factories from https://github.com/DFE-Digital/manage-courses-frontend/pull/109 then we should add some tests and probably a controller test.

### DIFF
--- a/app/controllers/courses/vacancies_controller.rb
+++ b/app/controllers/courses/vacancies_controller.rb
@@ -14,8 +14,8 @@ module Courses
         &.values&.each do |vacancy_status|
           site_status            = find_site_status vacancy_status[:id]
           site_status.vac_status = VacancyStatusDeterminationService.call(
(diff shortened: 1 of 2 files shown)


dankmitchell:
[Editing vacancies: multiple sites/both full-time and part-time](https://trello.com/c/rru4w3BN/1136-editing-vacancies-multiple-sites-both-full-time-and-part-time)

PR #115, opened 2019-03-28
[1210] Add courses table

tvararu:
### Context

This will allow users to view their courses and edit vacancies from the shiny new Rails frontend.

### Guidance to review

I've added strident red borders around the parts of the table that are still unimplemented / stubbed out.

### TODO

- [x] TESTS

### Screenshot

Screenshot 2019-03-28 at 15 15 47

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -3,6 +3,8 @@ class ApplicationController < ActionController::Base
   rescue_from JsonApiClient::Errors::NotAuthorized, with: :render_manage_ui
   rescue_from JsonApiClient::Errors::AccessDenied, with: :render_manage_ui
 
+  before_action :set_has_multiple_providers
+
   def not_found
     respond_to do |format|
       format.html { render 'errors/not_found', status: :not_found }
@@ -33,6 +35,15 @@ class ApplicationController < ActionController::Base
 
   def current_user_info
     current_user['info']
+  end
+
+  def set_has_multiple_providers
+    @has_multiple_providers = has_multiple_providers?
+  end
+
+  def has_multiple_providers?
+    provider_count = session.to_hash.dig("auth_user", "provider_count")
+    provider_count.nil? || provider_count > 1
   end
 
 private
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -1,0 +1,80 @@
+<%= content_for :page_title, "Courses" %>
+
+<nav class="govuk-breadcrumbs">
+  <ol class="govuk-breadcrumbs__list">
+    <% if @has_multiple_providers then %>
+      <li class="govuk-breadcrumbs__list-item">
+        <%= link_to "Organisations", providers_path, class: "govuk-breadcrumbs__link" %>
+      </li>
+    <% end %>
+    <li class="govuk-breadcrumbs__list-item">
+      <%= link_to @provider[:institution_name], provider_path(code: @provider[:institution_code]), class: "govuk-breadcrumbs__link" %>
+    </li>
+    <li class="govuk-breadcrumbs__list-item" aria-current="page">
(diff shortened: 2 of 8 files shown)


spec/features/courses/index_spec.rb
```diff
@@ -0,0 +1,15 @@
...
+  before do
+    stub_omniauth
+    stub_session_create
+    stub_api_v2_request('/api/v2/providers', create(:course))
+  end
+
+  scenario 'viewing the courses index page' do
+    visit '/organisations/AO/courses'
+
+    expect(find('h1')).to have_content('Courses')
```
heidar: Can we test that the course actually gets listed?

app/controllers/application_controller.rb
```diff
@@ -35,6 +37,15 @@ def current_user_info
     current_user['info']
   end
 
+  def set_has_multiple_providers
+    @has_multiple_providers = has_multiple_providers?
+  end
+
+  def has_multiple_providers?
+    provider_count = session.try(:[], :auth_user).try(:[], "provider_count")
```
heidar: You can use `#dig` instead of `try`, it's a bit nicer perhaps? Also `fetch` if you want a default if it's `nil`, to avoid to `||` below?

app/views/courses/index.html.erb
```diff
@@ -0,0 +1,85 @@
...
+
+  <p class="govuk-body">Use this section to:</p>
+
+  <ul class="govuk-list govuk-list--bullet">
+    <li>write about each course</li>
+    <li>preview and publish courses</li>
+    <li>copy content between courses</li>
+  </ul>
+
+  <h2 class="govuk-heading-m" style="border: 3px solid red">
```
heidar: Do we want to be using `style` attributes?

app/webpacker/stylesheets/application.scss
```diff
@@ -1,2 +1,11 @@
 @import "../../../node_modules/govuk-frontend/all";
+
+// Not in design system:
+$button-colour: #00823b;
+
 @import "cookie-banner";
+@import "phase-tag";
+
+.text-decoration-underline-dotted {
```
heidar: Do we want this in the main application.scss file or would it be better if it lived somewhere else?

app/views/courses/index.html.erb
```diff
@@ -0,0 +1,85 @@
...
+    <span class="govuk-caption-m">Accredited body</span>
+    Bradford College
```
defong: Hardcoded, dont forget this

app/views/courses/index.html.erb
tvararu: Yes, because they are temporary and will be removed. I just want to merge this because it's not accessible to the user and to prevent it from getting bigger.

app/views/courses/index.html.erb
tvararu: Yes, that's why there's the red border around it. 👍

app/webpacker/stylesheets/application.scss
tvararu: I like it here. :)

spec/features/courses/index_spec.rb
tvararu: The test is incomplete, will update it...

app/controllers/application_controller.rb
tvararu: I tried `dig` and it didn't work correctly. I'll try again.

tvararu:

spec/features/courses/index_spec.rb
tvararu: Done!

app/controllers/application_controller.rb
tvararu: Now `dig`ging, figured it out, cheers :)

PR #124, opened 2019-03-29
[1210] Add search-and-compare-ui base url to settings

tvararu:
### Context

The courses table links directly to find, so we need to add this. Requires a Terraform change as well: https://github.com/DFE-Digital/bat-infrastructure/pull/120

Also updates the VSTS settings with some missing keys.


tvararu:

PR #125, opened 2019-03-29
[1210] Fix stubbed out link to course page in courses table

tvararu:
Very small but self contained so may as well get it out while I do the rest.

### After

Screen Shot 2019-03-29 at 18 30 37


tvararu:

PR #127, opened 2019-03-31
Setup proper staging and qa environment

dankmitchell:
### Context
Proper environments i.e. `staging` and `qa` (for https://bat-dev-manage-courses-frontend-app.azurewebsites.net/)

- Add `{staging, qa}.rb` - duplicates of `production.rb`
- Add settings for `staging` and `qa`

### Guidance to review
`bin/rails server -e staging -b "ssl://localhost:3000?key=config/localhost/https/localhost.key&cert=config/localhost/https/localhost.crt" `

### DIFF
--- a/config/environments/qa.rb
+++ b/config/environments/qa.rb
@@ -1,0 +1,1 @@
+require Rails.root.join("config/environments/production")
--- a/config/environments/staging.rb
+++ b/config/environments/staging.rb
@@ -1,0 +1,1 @@
+require Rails.root.join("config/environments/production")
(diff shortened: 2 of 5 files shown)


heidar:
We could reduce a lot of the duplication in the environment files, here's how: https://stackoverflow.com/questions/18193946/cascading-inherited-shared-rails-configuration-environments

tvararu:
Agree with Heidar, would be good to set up some kind of base settings so that there's less duplicate files.

PR #129, opened 2019-04-01
[1210] Fix Find URL on courses table

tvararu:
Now that the environment variable is defined, we can update the URL.


tvararu:

tvararu:
@heidar done, also have added `.to_s` so that it doesn't blow up on the way into the link_to. :)

PR #133, opened 2019-04-03
[1210] Implement content_status in courses table

tvararu:
This is now being pulled in from the API and displayed in the table.

### After

Screenshot 2019-04-03 at 11 43 00

### DIFF
--- a/app/views/courses/_content_status_tag.html.erb
+++ b/app/views/courses/_content_status_tag.html.erb
@@ -1,1 +1,7 @@
-<span class="govuk-tag phase-tag--small phase-tag--published" style="border: 3px solid red">Published</span>
+<div class="govuk-tag phase-tag--small <%= course_content_tag_css_class(course) %>">
+  <%= course_content_tag_content(course) %>
+</div>
+
+<% if course_has_unpublished_changes(course) %>
+  <div class="govuk-body govuk-body-s govuk-!-margin-0">* Unpublished changes</div>
+<% end %>
--- a/app/helpers/course_helper.rb
+++ b/app/helpers/course_helper.rb
@@ -1,0 +1,23 @@
+module CourseHelper
+  def course_has_unpublished_changes(course)
+    course.attributes[:content_status] == "published_with_unpublished_changes"
+  end
+
+  def course_content_tag_content(course)
+    {
+      "published" => "Published",
+      "empty" => "Empty",
+      "draft" => "Draft",
+      "published_with_unpublished_changes" => "Published *"
+    }[course.attributes[:content_status]]
+  end
+
+  def course_content_tag_css_class(course)
+    {
+      "published" => "phase-tag--published",
+      "empty" => "phase-tag--no-content",
+      "draft" => "phase-tag--draft",
+      "published_with_unpublished_changes" => "phase-tag--published"
+    }[course.attributes[:content_status]]
+  end
+end
--- a/app/views/courses/_course_table_row.html.erb
+++ b/app/views/courses/_course_table_row.html.erb
@@ -9,8 +9,8 @@
   <td class="govuk-table__cell" style="border: 3px solid red">
     Running
   </td>
-  <td class="govuk-table__cell">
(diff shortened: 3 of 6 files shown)


tvararu:
[EPIC: Courses table in manage-courses-frontend](https://trello.com/c/Ju2Kfadq/1210-epic-courses-table-in-manage-courses-frontend)

tvararu:
[EPIC: Courses table in manage-courses-frontend](https://trello.com/c/Ju2Kfadq/1210-epic-courses-table-in-manage-courses-frontend)

app/views/courses/_content_status_tag.html.erb
```diff
@@ -1 +1,25 @@
-<span class="govuk-tag phase-tag--small phase-tag--published" style="border: 3px solid red">Published</span>
+<%
```
heidar: How do you feel about moving this to a helper?

tvararu: Is it likely to be used anywhere else?

tvararu: If you prefer it I'll do it though sure

dankmitchell: 100% a view helper

misaka: A part of me wants to get Draper in here to handle this kind of display logic.

defong: just use this as inspiration 

https://github.com/DFE-Digital/manage-courses-ui/blob/master/src/ui/Helpers/ViewModelHelpers.cs

😆

heidar: It's for separation of concerns and keeping the view clean, rather than reusability 🐱 Logic in templates is considered harmful, hacker news said it

tvararu: React has proven that logic in templates is alright, but, this is Rails-landia, and I'm happy to refactor this. 👍

tvararu: Refactored, check it out y'all

app/helpers/view_helper.rb
```diff
@@ -22,4 +22,28 @@ def search_ui_url(relative_path)
   def search_ui_course_page_url(provider_code:, course_code:)
     search_ui_url("/course/#{provider_code}/#{course_code}")
   end
+
+  def course_has_unpublished_changes(course)
+    course.attributes[:content_status] == "published_with_unpublished_changes"
+  end
+
+  def course_content_status(course)
```
heidar: How about making this two separate methods?

app/views/courses/_content_status_tag.html.erb
```diff
@@ -1 +1,9 @@
-<span class="govuk-tag phase-tag--small phase-tag--published" style="border: 3px solid red">Published</span>
+<% course_content_status = course_content_status(course) %>
```
heidar: If you use two different methods, we don't need this assignment.

app/helpers/view_helper.rb
```diff
@@ -22,4 +22,28 @@ def search_ui_url(relative_path)
...
+    status = course.attributes[:content_status]
```
heidar: If you only need the status, you could just pass in the status? Both works though :)

app/helpers/view_helper.rb
```diff
@@ -22,4 +22,28 @@ def search_ui_url(relative_path)
   def search_ui_course_page_url(provider_code:, course_code:)
     search_ui_url("/course/#{provider_code}/#{course_code}")
   end
+
+  def course_has_unpublished_changes(course)
```
heidar: Both of these methods are specific to courses. How about creating a new `helpers/course_helper.rb` for them to live in?

app/helpers/view_helper.rb
tvararu: I like passing in the course so that the helper does the digging, but yeah either is ok :)

PR #134, opened 2019-04-03
[1219] Change provider factory to produce JSONAPI

heidar:
### Context

The provider factory was never modified to produce JSONAPI responses in the same manner as the course and site factories. This PR is a precursor to 1219.

* Provider factory now produces JSONAPI responses
* Provider now `has_many :courses`

### DIFF
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -1,5 +1,7 @@
 class Provider < Base
+  has_many :courses, param: :provider_code
+
   def course_count
-    relationships.courses[:meta][:count]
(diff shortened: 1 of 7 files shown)


app/models/provider.rb
```diff
@@ -1,5 +1,7 @@
 class Provider < Base
+  has_many :courses, param: :provider_code
+
   def course_count
-    relationships.courses[:meta][:count]
+    courses.count
```
dankmitchell: 😍

PR #136, opened 2019-04-03
[1219] Sync course with search and compare

heidar:
### Context

After updating vacancies, call custom endpoint to cause course to by
synced with search and compare.

* Added custom endpoint
* Perform call after **ALL** site statuses have been updated

### DIFF
--- a/app/controllers/courses/vacancies_controller.rb
+++ b/app/controllers/courses/vacancies_controller.rb
@@ -26,6 +26,8 @@ module Courses
           site_status.save
         end
 
+      @course.sync_with_search_and_compare(provider_code: params[:provider_code])
+
       flash[:success] = 'Course vacancies published'
       redirect_to vacancies_provider_course_path(params[:provider_code], @course.course_code)
     end
--- a/app/models/provider.rb
+++ b/app/models/provider.rb
@@ -2,6 +2,8 @@ class Provider < Base
   has_many :courses, param: :provider_code
   has_many :sites
 
+  self.primary_key = :provider_code
+
   def course_count
     relationships.courses[:meta][:count]
   end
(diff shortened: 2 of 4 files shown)


app/models/provider.rb
```diff
@@ -1,5 +1,5 @@
 class Provider < Base
-  has_many :courses, param: :provider_code
+  self.primary_key = :provider_code
```
tvararu: Why are we removing the has_many association?

app/controllers/courses/vacancies_controller.rb
```diff
@@ -26,6 +26,8 @@ def update
           site_status.save
         end
 
+      @course.sync_with_search_and_compare(provider_code: params[:provider_code])
```
tvararu: I'm a bit lacking in context, but I assume this is already defined.

app/models/provider.rb
heidar: Sorry, it's not relevant to this PR at all, shame on me. I added it in the precursor to this PR but realised that it's not used anywhere.

spec/features/courses/vacancies/edit_spec.rb
```diff
@@ -35,6 +36,12 @@
       course
     )
 
+    stub_request(
+      :post,
+      "http://localhost:3001/api/v2/providers/AO/courses/" \
+        "#{course_attributes[:course_code]}/sync_with_search_and_compare"
+    ).to_return(status: 201, body: "")
```
misaka: Can we test that the controller action actually POSTs to this?

What about failure mode?

heidar: Handle it in the card that covers validations.

PR #141, opened 2019-04-03
Attempt to fix provider list factory

heidar:
### Context

Somehow JSON parsing the object returned by the factory returned
something vaguely similar to what we actually expected, so tests still
passed. This wasn't at all what we should be getting back from the API
though.

This commit fixes that.

### Guidance to review


heidar:
> looks reasonable

no it doesn't

PR #142, opened 2019-04-03
[1210] Add accrediting providers to the courses table

tvararu:
### Context

We need to display accrediting providers above each courses table on the courses page. Where the accrediting provider is set to the parent provider or is set to `null`, they should be displayed as self-accredited (no heading).

- pulls in the accrediting provider relation from the API
- groups them by their provider name
- sorts the providers by their names
- sorts the courses by their names
- splits out the self-accredited ones to be rendered separately
- tests to go

### Guidance to review

Paired with @heidar, we think there might be a better way to do the `sort_by` and `map(sort_by)` but we did a bit of code golfing and couldn't immediately think of something clearer.

Review without whitespace changes.

### Screenshot

Screen Shot 2019-04-05 at 16 55 17

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -3,7 +3,19 @@ class CoursesController < ApplicationController
   before_action :build_course, only: %i[show delete withdraw]
 
   def index
-    @provider = Provider.includes(:courses).find(params[:provider_code]).first
+    @provider = Provider
+      .includes(courses: [:accrediting_provider])
+      .find(params[:provider_code])
+      .first
+
+    @courses_by_accrediting_provider = @provider
+      .courses
+      .group_by { |course| course.accrediting_provider&.provider_name || @provider[:provider_name] }
+      .sort_by { |accrediting_provider, _| accrediting_provider }
+      .map { |pair| [pair[0], pair[1].sort_by(&:name)] }
+      .to_h
+
+    @self_accredited_courses = @courses_by_accrediting_provider.delete(@provider[:provider_name])
   end
 
   def show; end
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -3,16 +3,8 @@
 <% content_for :before_content do %>
   <nav class="govuk-breadcrumbs">
     <ol class="govuk-breadcrumbs__list">
-      <% if @has_multiple_providers then %>
-        <li class="govuk-breadcrumbs__list-item">
-          <%= link_to "Organisations", providers_path, class: "govuk-breadcrumbs__link" %>
-        </li>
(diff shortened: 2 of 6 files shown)


tvararu:
[EPIC: Courses table in manage-courses-frontend](https://trello.com/c/Ju2Kfadq/1210-epic-courses-table-in-manage-courses-frontend)

app/controllers/courses_controller.rb
```diff
@@ -3,6 +3,12 @@ class CoursesController < ApplicationController
 
   def index
     @provider = Provider.find(params[:code]).first.attributes
-    @courses = Course.where(provider_code: params[:provider_code])
+
+    @courses = Course
+      .includes(:accrediting_provider)
+      .where(provider_code: params[:provider_code])
+
+    @courses_by_accrediting_provider = @courses
```
fofr: Do we want to call this "accredited body" in the new code?

tvararu: Sure

heidar:
This should be able to be simplified once https://github.com/DFE-Digital/manage-courses-backend/pull/245/files#diff-7d6b888ba704f4684c5b576a1dd07676R15 is merged. :)

app/controllers/courses_controller.rb
tvararu: This is actually very annoying to keep in my head; I'll stick to `accrediting_provider` in the code.

app/views/courses/index.html.erb
```diff
@@ -27,5 +27,14 @@
     <li>copy content between courses</li>
   </ul>
 
-  <%= render partial: 'course_table', locals: { courses: @provider.courses } %>
+  <%= render partial: 'course_table', locals: { courses: @self_accredited_courses.sort_by(&:name) } %>
```
dankmitchell: Can `.sort_by(&:name)` be put in the controller, please?

tvararu: Done

tvararu:
@dankmitchell the `include_nulls` is meant to address the behaviour where in JSONAPI if you don't ask for certain metadata on an endpoint, you get this:

```json
            "relationships": {
                "accrediting_provider": {
                    "meta": {
                        "included": false
                    }
                }
            },
```

But, if you explicitly ask for the `accrediting_provider` relation, but there is nothing related to it, then it gives you `data: null`:

```json
            "relationships": {
                "accrediting_provider": {
                    "data": null
                }
            },
```

Initially these seemed interchangeable but in practice they mean different things; the first is "you didn't ask for this information," the second is "you asked, but there isn't anything here." So `include_nulls` makes the serializer aware and capable of dealing with this difference.

PR #143, opened 2019-04-03
1238 running site statuses

dankmitchell:
### Context
Running `site_statuses`

- Refactor vacancies controller
- Filter `site_statuses` to those only with a status of running

### Guidance to review

# After
Screenshot 2019-04-03 at 20 10 03

# Before
Screenshot 2019-04-03 at 20 11 36

### DIFF
--- a/app/controllers/courses/vacancies_controller.rb
+++ b/app/controllers/courses/vacancies_controller.rb
@@ -36,7 +36,7 @@ module Courses
     end
 
     def build_site_statuses
-      @site_statuses = @course.site_statuses
+      @site_statuses = @course.site_statuses.select { |site_status| site_status.status == 'running' }
     end
 
     def find_site_status(id)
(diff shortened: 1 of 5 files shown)


app/controllers/courses/vacancies_controller.rb
```diff
@@ -33,15 +32,11 @@ def update
...
+      @course = Course.where(provider_code: params[:provider_code]).find(params[:code]).first
     end
 
     def build_site_statuses
-      @site_statuses = @course.map(&:site_statuses).flatten
-    end
-
-    def build_first_course
-      @course = @course.first
+      @site_statuses = @course.site_statuses.select { |site_status| site_status.status == 'running' }
```
benilovj: You might want to put the following helper on `SiteStatus`:

```ruby
def running?
  status == 'running'
end
```

and then this can be prettified to

```ruby
@site_statuses = @course.site_statuses.select(&:running?)
```

benilovj:
Would be good to get some test coverage for this case...

app/controllers/courses/vacancies_controller.rb
dankmitchell: Great shout. Done

PR #149, opened 2019-04-04
[1257] add 'ucas' rails env

misaka:
Also minor fixes to qa and staging envs.

https://trello.com/c/0zd6xRn9/1257-add-ucas-rails-environment-to-mc-fe

### Context

The UCAS environment  is running in `development`. We want this env to run like the others with it's own `RAILS_ENV` so that we can configure parts of it in the app itself. Part of the reason we want to change this is that the ucas environment is currently broken because it needs the env var `SETTINGS__MANAGE_BACKEND__BASE_URL` but we don't have access to add that in VSTS. If we add a proper Rails env we can just set it using the standard env settings.

Add `ucas` environment.

### Guidance to review

The `qa` and `staging` environments were both missing the setting `Settings.search_ui.base_url` in the settings file. In `qa` this was being supplied in the application settings in Azure, but for `staging` this was completely missing. Aside from this fix, these changes shouldn't affect any envs other than creating the UCAS one.

### DIFF
--- a/config/environments/ucas.rb
+++ b/config/environments/ucas.rb
@@ -1,0 +1,1 @@
+require Rails.root.join("config/environments/production")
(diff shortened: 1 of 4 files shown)


misaka:
[Add ucas Rails environment to mc-fe.](https://trello.com/c/0zd6xRn9/1257-add-ucas-rails-environment-to-mc-fe)

PR #150, opened 2019-04-04
[1217] Fix relation between provider and course

heidar:
### Context

**Ignore the branch name, this is a precursor to doing that.** 

Before we were trying to map them with `provider_code`, which isn't an
attribute on course. The API returns the `provider_code` as
`institution_code` instead.

I've fixed these and fixed all the tests that failed by doing so. Most
of them were due to explicitly looking providers up by `provider_code`.

This allows us to look up a provider and do `provider.courses`, assuming
that we included courses in the provider lookup.

Likewise, we can do `course.provider`, assuming the provider was
included in the course lookup.

Change all mentions of `provider_code` to correctly call `institution_code`. The exception to this is the routes and `params`, because the route is still `/provider` and this `code` becomes `provider_code`.

### Guidance to review

This probably needs some manual testing as we're too far removed from the actual API to know if this will work or not.

### DIFF
--- a/app/controllers/courses/vacancies_controller.rb
+++ b/app/controllers/courses/vacancies_controller.rb
@@ -25,7 +25,7 @@ module Courses
           site_status.save
         end
 
-      @course.sync_with_search_and_compare(provider_code: params[:provider_code])
+      @course.sync_with_search_and_compare(institution_code: params[:provider_code])
 
       flash[:success] = 'Course vacancies published'
       redirect_to vacancies_provider_course_path(params[:provider_code], @course.course_code)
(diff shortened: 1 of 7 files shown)


defong:
Reminds me of 
https://github.com/DFE-Digital/manage-courses-api/pull/210
https://github.com/DFE-Digital/manage-courses-ui/pull/290

Whatever xxx code we call it, let it be ambiguoulyubiquitous consistent.

app/controllers/courses/vacancies_controller.rb
```diff
@@ -34,7 +34,7 @@ def update
   private
 
     def build_course
-      @course = Course.where(provider_code: params[:provider_code]).find(params[:code]).first
+      @course = Course.where(institution_code: params[:provider_code]).find(params[:code]).first
```
heidar: Forgot to add includes here for sites and site statuses, silly me.

heidar:
Closing this in favour of https://github.com/DFE-Digital/manage-courses-backend/pull/249

PR #163, opened 2019-04-08
[1267] Add location edit page

dankmitchell:
### Context
Editing sites aka training locations

Add a hidden page for editing training locations (the form doesn't do anything other than render the data)

### Guidance to review
Example - `https://localhost:3000/organisations/2DO/locations/10329711/edit`

# After
localhost_3000_organisations_2DO_locations_10329711_edit


dankmitchell:
There will be a request spec when the form gets hooked up in a later PR

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,17 @@
...
-  before_action :authenticate
+  before_action :authenticate, :build_provider
 
   def index
-    @provider = Provider.includes(:sites).find(params[:provider_code]).first
     @sites = @provider.sites.sort_by(&:location_name)
   end
+
+  def edit
+    @site = Site.where(provider_code: params[:provider_code]).find(params[:id]).first
```
heidar: `Provider.includes(:sites)` should get you all the sites, so I don't think you need another API call here, right?

app/models/site.rb
```diff
@@ -1,4 +1,19 @@
 class Site < Base
   has_one :site_status
-  belongs_to :provider
+  belongs_to :provider, param: :provider_code
+
+  REGIONS = [
+    ['', :no_region],
```
fofr: Should this be explicitly "No region" @benilovj?

app/controllers/sites_controller.rb
misaka: Wouldn't it be better to only get the site necessary for the edit action, instead of all the sites? In which case `build_provider` would be limited to only trigger for the `index` action.

spec/features/locations_spec.rb
```diff
@@ -14,27 +14,38 @@
...
+
+    expect(find('h1')).to have_content('Main site 1')
+  end
+
   context 'when the provider is opted_in' do
     let(:provider) do
       jsonapi(:provider, :opted_in, sites: sites).render
     end
 
     scenario 'should show Add Location CTA' do
```
defong: What is `CTA`?

fofr: Call to action

app/models/site.rb
benilovj: 👍 to this suggestion

app/views/sites/edit.html.erb
```diff
@@ -0,0 +1,61 @@
...
+
+      <h2 class="govuk-heading-m govuk-!-margin-top-8">Address</h2>
+
+      <div class="govuk-form-group">
+        <%= f.label :address1, "Building and street", class: "govuk-label" %>
+        <%= f.text_field :address1, class: "govuk-input govuk-!-width-full" %>
+      </div>
+
+      <div class="govuk-form-group">
+        <%= f.label :address2, "Town or city", class: "govuk-label" %>
```
fofr: Address 2 should be another line after Address1.

fofr: Screen Shot 2019-04-08 at 14 30 12

app/models/site.rb
```diff
@@ -1,4 +1,19 @@
 class Site < Base
   has_one :site_status
-  belongs_to :provider
+  belongs_to :provider, param: :provider_code
+
+  REGIONS = [
```
heidar: Does the jsonapiclient gem have enums at all?

app/controllers/sites_controller.rb
heidar: We need the provider to render the page as well, which is why I suggested the first option.

app/models/site.rb
heidar: it does not!

heidar:
Do not merge, @tvararu is working on validations for this.

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,17 @@
...
-  before_action :authenticate
+  before_action :authenticate, :build_provider
 
   def index
-    @provider = Provider.includes(:sites).find(params[:provider_code]).first
     @sites = @provider.sites.sort_by(&:location_name)
   end
+
+  def edit
+    @site = @provider.sites.find(params[:id]).first
```
tvararu: I don't think this `.find` works...

```
@provider.sites.first
=> #<Site:@attributes={"type"=>"sites", "id"=>"9576475", "code"=>"R", "location_name"=>"One Degree Academy", "address1"=>"Queensway", "address2"=>"Enfield", "address3"=>"", "address4"=>"", "postcode"=>"EN3", "region_code"=>"london"}>
@provider.sites.find(999).first
=> #<Site:@attributes={"type"=>"sites", "id"=>"9576475", "code"=>"R", "location_name"=>"One Degree Academy", "address1"=>"Queensway", "address2"=>"Enfield", "address3"=>"", "address4"=>"", "postcode"=>"EN3", "region_code"=>"london"}>
```

tvararu:
I've pushed some hacky validations that I will refine, and marked this with a WIP commit as a result.

app/controllers/sites_controller.rb
heidar: Is that in the tests? Cause webmock doesn't care what the ID is, it just returns the factory responses.

heidar:
> I've pushed some hacky validations that I will refine, and marked this with a WIP commit as a result.

I think we should do either:

1) Make a form object that handles validations and passes to our JsonApiClient models
2) Skip frontend validations and just do it server side. JsonApiClient supports this out of the box

Let's discuss tomorrow.

app/controllers/sites_controller.rb
```diff
@@ -5,8 +5,33 @@ def index
...
+    @errors = {}
+    if params[:site] then
+      site_by_name = sites.select { |site| site.location_name == params[:site][:location_name] }.first
+      name_is_in_use = !site_by_name.nil? && site_by_name.id != @site.id
+      @errors[:location_name] = 'Enter a name that is not in use' if name_is_in_use
+
+      @errors[:location_name] = 'Enter a name' if params[:site][:location_name]&.empty?
+      @errors[:address1] = 'Enter an address' if params[:site][:address1]&.empty?
+      @errors[:address3] = 'Enter a town or city' if params[:site][:address3]&.empty?
+      @errors[:postcode] = 'Enter a valid postcode (for example, SW1P 3BT)' if params[:site][:postcode]&.empty?
```
fofr: `SW1P 3BT` is a valid example but doesn't follow the standard postcode format (note the letter after the first number). Maybe use a more common postcode format like `BN1 1AA` instead?

app/views/sites/edit.html.erb
```diff
@@ -15,24 +15,51 @@
 
 <h1 class="govuk-heading-xl"><%= @site.location_name %></h1>
 
+<% if @errors.any? then %>
```
fofr: When there is an error can we also prefix the title with `Error: ` as recommended by https://design-system.service.gov.uk/components/error-summary/

app/controllers/sites_controller.rb
defong: manage courses ui uses this message

https://github.com/DFE-Digital/manage-courses-ui/blob/875ae72320369dbec9528c245abb4b6274acac51/src/ui/ViewModels/OrganisationViewModel.cs#L64

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,42 @@
...
```
fofr: https://bat-design-history.herokuapp.com/publish-teacher-training/postcodes

app/views/sites/edit.html.erb
tvararu: Done, good suggestion 👍

app/controllers/sites_controller.rb
tvararu: Done, good suggestion :+1:

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,62 @@
...
+    @errors = {}
+  end
+
+  def map_errors(message)
+    {
+      "Location name can't be blank"             => "Name is missing",
+      "Address1 can't be blank"                  => "Building and street is missing",
+      "Address3 can't be blank"                  => "Town or city is missing",
+      "Postcode can't be blank"                  => "Postcode is missing",
+      "Postcode not recognised as a UK postcode" => "Postcode is not valid (for example, SW1A 1AA)"
```
tvararu: @fofr what do you think of these messages?

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,62 @@
...
-  before_action :authenticate
+  before_action :authenticate, :build_provider, :initialise_errors
 
   def index
-    @provider = Provider.includes(:sites).find(params[:provider_code]).first
     @sites = @provider.sites.sort_by(&:location_name)
   end
+
+  def edit
+    @site = @provider.sites.find { |site| site.id == params[:id] }
```
misaka: MINOR: Feels like it'd be worth putting this into a builder hook for `edit` and `update`.

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,62 @@
...
+  end
+
+  def map_errors(message)
+    {
+      "Location name can't be blank"             => "Name is missing",
+      "Address1 can't be blank"                  => "Building and street is missing",
+      "Address3 can't be blank"                  => "Town or city is missing",
+      "Postcode can't be blank"                  => "Postcode is missing",
+      "Postcode not recognised as a UK postcode" => "Postcode is not valid (for example, BN1 1AA)"
+    }[message] || message
```
misaka: I still like the idea of having this kind of mapping in some kind of form object ... not worth it for this release but something to keep in the back of our minds maybe.

app/models/site.rb
```diff
@@ -1,4 +1,19 @@
...
+    ['Wales', :wales],
+    ['West Midlands', :west_midlands],
+    ['East Midlands', :east_midlands],
+    ['Eastern', :eastern],
+    ['Noth West', :north_west],
+    ['Yorkshire and the Humber', :yorkshire_and_the_humber],
+    ['North East', :north_east],
+    ['Scotland', :scotland],
+    ['No Region', :no_region],
+  ].freeze
```
misaka: Maybe what we should do is be requesting this list from the API once and cache it. There might be a few things we want to query the backend about on startup ...

misaka: (to be clear, not worth doing now, just something to keep in mind)

app/controllers/sites_controller.rb
```diff
@@ -1,8 +1,62 @@
...
+    @site.provider_code = @provider.provider_code
+
+    if @site.update(site_params)
+      redirect_to provider_sites_path, flash: { success: 'Your changes have been published' }
+    else
+      @errors = @site.errors.reduce({}) { |errors, (field, message)|
+        errors[field] ||= []
+        errors[field].push(map_errors(message))
+        errors
+      }
```
misaka: Let's not forget to refactor this into a method as we'll want to re-use it. Also, is this a cleaner implementation?

```
@errors = @site.errors
  .group_by { |field, _message| field }
  .transform_values { |group| group.map &:second }
```

Maybe I'm blinded by a slight aversion to using `reduce`, just a thought.

app/controllers/sites_controller.rb
tvararu: Done

app/controllers/sites_controller.rb
tvararu: I fully agree that we need to revisit how we do forms.

app/controllers/sites_controller.rb
tvararu: I think the reducer is actually easier to understand, I had to `binding.pry` to figure that one out. :)

PR #164, opened 2019-04-08
[1241] Tweaks to contact links

benilovj:
### Context
There is a bit of duplication in how we generate the contact links in the service. Also, I've noticed that we don't include in the provider code in withdraw/delete course emails

- Add a view helper for generating contact emails
- Add missing provider code to withdraw/delete course emails

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -1,6 +1,4 @@
 module ViewHelper
-  BECOMING_A_TEACHER_MAILBOX_EMAIL = 'becomingateacher@digital.education.gov.uk'.freeze
-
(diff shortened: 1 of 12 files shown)


app/helpers/view_helper.rb
```diff
@@ -1,4 +1,6 @@
 module ViewHelper
+  BECOMING_A_TEACHER_MAILBOX_EMAIL = 'becomingateacher@digital.education.gov.uk'.freeze
```
misaka: Feels like this should be in Settings.

benilovj: Yeah, good call. Fixed in a4c4cdc

PR #166, opened 2019-04-08
Suppress study mode for vacancies on some locations

fofr:
If a course is all part time or all full time, we don't need to add "Full time" and "Part time" to the end of the location label.

### Before
Screen Shot 2019-04-08 at 16 53 50

### After
Screen Shot 2019-04-08 at 16 53 57


benilovj:
@fofr tests now passing

PR #179, opened 2019-04-10
[1157] Return user to courses index after editing vacancies

tvararu:
### Guidance to review

:ship:


tvararu:
[Return user back to where they clicked 'edit vacancies'](https://trello.com/c/T7UJ3n1p/1157-return-user-back-to-where-they-clicked-edit-vacancies)

PR #180, opened 2019-04-10
[1238] Update findable column on courses table

dankmitchell:
### Context
Find column on course table (courses#index)

Expand on content when a course is not "findable"

### Guidance to review
/courses

### DIFF
--- a/app/views/courses/_course_table_row.html.erb
+++ b/app/views/courses/_course_table_row.html.erb
@@ -20,7 +20,11 @@
         search_ui_course_page_url(provider_code: @provider[:provider_code], course_code: course.attributes[:course_code])
       %>
     <% else %>
-      No
+      <% if course.ucas_status == 'new' %>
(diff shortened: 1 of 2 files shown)


app/views/courses/_course_table_row.html.erb
```diff
@@ -20,7 +20,11 @@
         search_ui_course_page_url(provider_code: @provider[:provider_code], course_code: course.attributes[:course_code])
       %>
     <% else %>
-      No
+      <% if course.ucas_status == 'new' %>
+        No – <% @provider.opted_in? ? 'still in draft' : 'must be published on UCAS' %>
```
benilovj: This should be `<%=` instead of `<%`.

It's probably worth extracting the `else` logic into a view helper so we can test it easier.

fofr: Fixed in #185

PR #190, opened 2019-04-12
[1253] Initial course page basic details

tvararu:
### Context

We need to show further information about the course.

Populates the course basic details page with info from the course object.

### Guidance to review

This is a first PR that does most of the work and I aim to come back with separate PRs for the rest of the fields. We should merge this sooner than later to prevent it from growing and becoming harder to review.

### Screenshot

Screenshot 2019-04-15 at 14 21 50

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1,9 +1,13 @@
 class CoursesController < ApplicationController
   before_action :authenticate
   before_action :build_course, only: %i[show delete withdraw]
-  before_action :build_provider, only: %i[index show]
 
   def index
+    @provider = Provider
+      .includes(courses: [:accrediting_provider])
+      .find(params[:provider_code])
+      .first
+
     # rubocop:disable Style/MultilineBlockChain
     @courses_by_accrediting_provider = @provider
       .courses
@@ -25,7 +29,9 @@ class CoursesController < ApplicationController
     @self_accredited_courses = @courses_by_accrediting_provider.delete(@provider[:provider_name])
   end
 
-  def show; end
+  def show
+    @provider = @course.provider
+  end
 
   def withdraw; end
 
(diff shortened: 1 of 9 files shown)


tvararu:
Follow-up cards have been open for the age range, apprenticeship, and applications open info.

tvararu:
[Course - basic detail tab](https://trello.com/c/izY1pEUm/1253-course-basic-detail-tab)

spec/features/courses/show_spec.rb
```diff
@@ -26,34 +26,36 @@
...
+
   scenario 'viewing the show courses page' do
     visit "/organisations/A0/courses/#{course.attributes[:course_code]}"
 
-    expect(find('.govuk-caption-xl')).to have_content(
+    expect(course_page.caption).to have_content(
       course.attributes[:description]
     )
-    expect(find('.govuk-heading-xl')).to have_content(
+    expect(course_page.title).to have_content(
```
fofr: How do you differentiate between the title element and the H1?

tvararu: `title` here can be renamed to `heading` perhaps, it's just a hook anyway. It can be anything we like, doesn't automatically mean `title` is the `<title>`

app/controllers/courses_controller.rb
```diff
@@ -29,7 +29,12 @@ def index
     @self_accredited_courses = @courses_by_accrediting_provider.delete(@provider[:provider_name])
   end
 
-  def show; end
+  def show
+    @provider = Provider
+      .includes(:sites)
+      .find(params[:provider_code])
+      .first
```
misaka: It feels to me like we should be packaging this up with the API call to get `@course` ... is the provider not already included in the response there? Can we add it?

tvararu: Good shout, let me give it a go

app/helpers/course_helper.rb
```diff
@@ -28,4 +28,18 @@ def course_ucas_status(course)
       "not_running" => "Not running"
     }[course.attributes[:ucas_status]]
   end
+
+  def course_qualifactions(course)
+     # TODO: Should this be in the API? Remove this comment if this passes code review.
```
misaka: I don't think the API should do sorting, by default, generally. There could be exceptions but I'm not sure if this qualifies?

Also, typo: `course_qualifactions` -> `course_qualifications`. :)

fofr: Should the method be called "course_outcome"?

app/controllers/courses_controller.rb
tvararu: Got it 👍

app/helpers/course_helper.rb
tvararu: Done

PR #191, opened 2019-04-12
[1306] add site prism

misaka:
### Context

SitePrism is a page object system that will help us clean up our views.

Add the SitePrism gem and a mechanism to load them up so that they're accessible in the views. Also re-enable the development error pages in development env and add `better_errors` gem to allow for better debugging of errors.

### Guidance to review

I was planning to also add more examples of how to use it which may or may not make this PR, but the mechanism as introduced in spec/support/load_site_prism_pages.rb is worth reviewing now.


misaka:
> So I don't particularly like it, but for what we're trying to achieve (auto-generated test classes) and given that this is a Ruby project with sensibilities that are not the same as mine generally, it looks okay?.. 😅

I can see your point. The idea was to automatically generate helper methods, kind of like the Rials routes helpers, or Rspec helpers are generated, partly to help standardise the naming so if we use `organisations_courses_page` we know it always refers to that page which is kind of a Singleton. But, I can see how the magic here can be disconcerting.

I've uploaded an alternative implementation which has an advantage in that it uses the autoloader to avoid possible load-order errors (which I've encountered with these pages in the past). I've included an example of how it can be used, and you can see how the page is explicitly loaded.

misaka:
[Introduce SitePrism for frontend testing](https://trello.com/c/aaJRlarF/1306-introduce-siteprism-for-frontend-testing)

PR #195, opened 2019-04-16
[1317] Display course apprenticeships and funding

tvararu:
### Context

We need to expose all the things.

Use the `funding` field which is now being passed in from the backend.

### Guidance to review

We don't have a request spec for the presence of the `apprenticeship` in the frontend and the hiding of the other two but ehhhh.

### Screenshot

Screenshot 2019-04-16 at 17 23 13


tvararu:
[Course basic details: funding (frontend)](https://trello.com/c/0VIoIz3P/1317-course-basic-details-funding-frontend)

PR #197, opened 2019-04-17
[1315] Display applications_open in courses show

tvararu:
### Context

We need to expose all the course information on the courses show page.

- Present applications_open in courses show view

### Guidance to review

When `applications_open_from` is set to null on the API, for example for courses that don't have any findable site statuses, the frontend will display the `Applications open` info but there will be an empty
string.

### Screenshot

Almost there!

Screenshot 2019-04-17 at 16 51 54

### DIFF
--- a/app/helpers/course_helper.rb
+++ b/app/helpers/course_helper.rb
@@ -55,4 +55,8 @@ module CourseHelper
       'Fee paying (no salary)'
     end
   end
+
+  def course_applications_open(course)
+    course.attributes[:applications_open_from]&.to_date&.strftime("%-d %B %Y")
+  end
 end
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -111,8 +111,8 @@
         <dt class="govuk-summary-list__key">
           Applications open
         </dt>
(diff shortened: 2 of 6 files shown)


tvararu:
[Course basic details: applications open (frontend)](https://trello.com/c/eQxrIVXf/1323-course-basic-details-applications-open-frontend)

app/helpers/course_helper.rb
```diff
@@ -55,4 +55,8 @@ def course_funding(course)
       'Fee paying (no salary)'
     end
   end
+
+  def course_applications_open(course)
+    course.attributes[:applications_open_from]&.to_date&.strftime("%-d %B %Y")
```
fofr: I wonder if `%-d %B %Y` can be put into something reusable like `:govuk_date` (see also https://github.com/alphagov/govuk_admin_template/blob/0dfd5cffe8ec6ffa2d29767531f1e64826773b7e/lib/govuk_admin_template/engine.rb#L16-L31)

tvararu: 👍 at some point I don't see why not

PR #213, opened 2019-04-23
[XXXX] Make puma auto-generate SSL certificates

tvararu:
Removes one step from the initial setup.

### DIFF
--- a/config/puma.rb
+++ b/config/puma.rb
@@ -9,7 +9,8 @@ threads threads_count, threads_count
 
 # Specifies the `environment` that Puma will run in.
 #
-environment ENV.fetch("RAILS_ENV") { "development" }
+env = ENV.fetch("RAILS_ENV", 'development')
+environment env
 
 # Specifies the number of `workers` to boot in clustered mode.
 # Workers are forked webserver processes. If using threads and workers together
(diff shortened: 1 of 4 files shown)


config/puma.rb
```diff
@@ -29,17 +30,35 @@
 # Allow puma to be restarted by `rails restart` command.
 plugin :tmp_restart
 
-# SSL in development (as DFE Signin redirects to https://localhost:3000 but
-# just listen on port 3000 elsewhere
 listen_port = ENV.fetch("PORT") { 3000 }
 
-if Rails.env.development?
+if env == "development"
```
dankmitchell: I think I prefer `if Rails.env.development?`

config/puma.rb
```diff
@@ -9,7 +9,8 @@
 
 # Specifies the `environment` that Puma will run in.
 #
-environment ENV.fetch("RAILS_ENV") { "development" }
+env = ENV.fetch("RAILS_ENV") { "development" }
+environment env
```
dankmitchell: What's the benefit of this diff?

config/puma.rb
```diff
@@ -29,17 +30,35 @@
...
 listen_port = ENV.fetch("PORT") { 3000 }
 
-if Rails.env.development?
+if env == "development"
+  cert = "#{Dir.pwd}/#{File.join('config', 'localhost', 'https', 'localhost.crt')}"
+  key = "#{Dir.pwd}/#{File.join('config', 'localhost', 'https', 'localhost.key')}"
 
-  cert = Rails.root.join('config', 'localhost', 'https', 'localhost.crt')
-  key = Rails.root.join('config', 'localhost', 'https', 'localhost.key')
+  unless File.exist?(key)
```
dankmitchell: Do we need to check for  `File.exist?(cert) && File.exist?(key)` ?

config/puma.rb
tvararu: I added it because we re-use the env variable later on

config/puma.rb
tvararu: We can't use `Rails` because we call `puma` directly in the procfile, it's weird...

config/puma.rb
tvararu: Yeah may as well

config/puma.rb
```diff
@@ -9,7 +9,8 @@
 
 # Specifies the `environment` that Puma will run in.
 #
-environment ENV.fetch("RAILS_ENV") { "development" }
+env = ENV.fetch("RAILS_ENV") { "development" }
```
misaka:
```
ENV.fetch('RAILS_ENV', 'development')
```

The block is just used when you don't want to automatically execute `"development"` every time it runs. For example what if that was `YAML.load(File.read(".config.yml"))[:environment]` ... you wouldn't want to execute that unless `RAILS_ENV` wasn't set.

config/puma.rb
```diff
@@ -29,17 +30,35 @@
...
+      root_ca.not_before = Time.now
+      root_ca.not_after = root_ca.not_before + 2 * 365 * 24 * 60 * 60 # 2 years validity
+      root_ca.sign(root_key, OpenSSL::Digest::SHA256.new)
+      root_ca
+    end
+
+    root_key = OpenSSL::PKey::RSA.new(2048)
+    file = File.new(key, "wb")
+    file.write(root_key)
+    file.close
```
misaka: Not a blocker, but more Ruby-like alternatives:

```
File.open(key, 'wb') do |file|
  file.write(root_key)
end
```

or

```
File.write(key, root_key, 'wb')
```

tvararu: Done

config/puma.rb
tvararu: Done

tvararu:
I've tested this locally by pulling a separate repo. There's difficulties with Safari sometimes caching the certificate unreasonably past browser restarts and keychain adds, but from what I can tell, it works.

PR #219, opened 2019-04-24
[1255] add publishing panel

dankmitchell:
### Context
Course published status

Add published status panel to the description page

### After
localhost_3000_organisations_2AT_courses_3CXD_description(Laptop with MDPI screen)

### DIFF
--- a/app/views/courses/_status_panel.html.erb
+++ b/app/views/courses/_status_panel.html.erb
@@ -30,6 +30,8 @@
   </p>
 
   <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
+  <p class="govuk-body">Last published:<br><%= formatted_date(@course.changed_at) %></p>
+  <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
 
   <% if @course.is_running? %>
     <h3 class="govuk-heading-m">Withdraw</h3>
--- a/app/views/courses/_status_panel.html.erb
+++ b/app/views/courses/_status_panel.html.erb
@@ -1,0 +1,45 @@
+<div class="status-box">
+  <h3 class="govuk-heading-m">Status</h3>
+  <div class="govuk-!-margin-bottom-6">
+    <%= render partial: 'content_status_tag', locals: { course: @course } %>
+  </div>
+  <h3 class="govuk-heading-s govuk-!-margin-bottom-0">Is it on Find?</h3>
+  <% if @course.findable? %>
+    <p class="govuk-body">
+      <%= govuk_link_to "Yes - view online", search_ui_course_page_url(provider_code: @provider.provider_code, course_code: @course.course_code) %>
+    </p>
+  <% else %>
+    <p class="govuk-body">
+      <% if @course.new_and_not_running? %>
+        No – <%= @provider.opted_in? ? 'still in draft' : 'must be published on UCAS' %>
+      <% else %>
+        No
+      <% end %>
+    </p>
+  <% end %>
+
+  <h3 class="govuk-heading-s govuk-!-margin-bottom-0">Applications</h3>
+  <p class="govuk-body"><%= @course.open_for_applications? ? "Open" : "Closed" %></p>
+
+  <h3 class="govuk-heading-s govuk-!-margin-0">Vacancies</h3>
+  <p class="govuk-body">
+    <%= @course.has_vacancies? ? "Yes" : "No" %>
+    <% if @provider.opted_in? %>
+      (<%= link_to 'Edit', vacancies_provider_course_path(code: @course.course_code), class: 'govuk-link' %>)
+    <% end %>
+  </p>
+
+  <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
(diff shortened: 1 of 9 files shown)


misaka:
[Course description page - publishing panel](https://trello.com/c/gD3Nh5LY/1255-course-description-page-publishing-panel)

app/views/courses/_status_panel.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <h3 class="govuk-heading-s govuk-!-margin-0">Vacancies</h3>
+  <p class="govuk-body">
+    <%= @course.has_vacancies? ? "Yes" : "No" %>
+    <% if @provider.opted_in? %>
+      (<%= link_to 'Edit', vacancies_provider_course_path(code: @course.course_code), class: 'govuk-link' %>)
+    <% end %>
+  </p>
+
+  <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
+  <p class="govuk-body">Last published:<br><%= formatted_date(@course.changed_at) %></p>
```
misaka: Why is `changed_at` the right way to tell when a course was last published? I think that there are some cases where `changed_at` might be updated but nothing has actually changed in terms of published data.

dankmitchell: What is the correct point of reference for the last time a user published this course? cc. @benilovj

benilovj: @misaka is there a more appropriatr timestamp that would reflect the publishing time? `changed_at` will incorporate new training locations being added/removed, vacancy changes etc which seem relevant to the user.

dankmitchell: They were my thoughts too Jake. I think this should remain and if it’s an issue at a later point then we address it separately.

misaka: Doesn't this get updated if a draft enrichment is changed? That's the situation I was thinking of. We originally thought that this would be No Big Deal (:tm:) since it would only mean that it would be resent to UCAS but since only the published version is sent, it would be the same as the previous version.

If I'm right (need to double-check), would this be an issue with it's use here?

benilovj: @misaka oh, I wasn't aware of that, is that happening on the C# side?

dankmitchell: Interesting. I’ll re-look at the API and Expose the timestamp from the latest “published” enrichment instead.

benilovj: Maybe if the label were something like "Last changed" instead of "Last published", it wouldn't be misleading.

Alternatively we could take the `last_published_timestamp_utc` timestamp from the latest published enrichment and display that.

It may be worth taking that widget out of this PR into a separate one, so we don't block merging the rest.

dankmitchell: Done

app/views/courses/_status_panel.html.erb
```diff
@@ -0,0 +1,43 @@
...
+    <p class="govuk-body">
+      <%= govuk_link_to "Yes - view online", search_ui_course_page_url(provider_code: @provider.provider_code, course_code: @course.course_code) %>
+    </p>
+  <% else %>
+    <p class="govuk-body">
+      <% if @course.new_and_not_running? %>
+        No – <%= @provider.opted_in? ? 'still in draft' : 'must be published on UCAS' %>
+      <% else %>
+        No
+      <% end %>
```
misaka: It's logic like this in the view that makes me want to just get the `draper` gem in here. :(

misaka:
> Looks fine but is there test for the status panel?

Agreed. There's a lot of logic in here, and if we're not going to separate that out into a decorator then we need some testing. Not to keep banging on it, but just another reason we should use ... (*blah* *blah* *blah* *ya we know*).

app/webpacker/stylesheets/application.scss
```diff
@@ -54,3 +54,7 @@ $button-colour: #00823b;
 .course-parts__fields__value--empty {
   color: govuk-colour("grey-1");
 }
+
+.govuk-link--destructive {
```
tvararu: Ehhhhh I would just call it `govuk-link--red` to be fair...

app/views/courses/_status_panel.html.erb
```diff
@@ -0,0 +1,45 @@
...
+  <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
+  <p class="govuk-body">Last published:<br><%= formatted_date(@course.last_published_at) %></p>
+  <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
+
+  <% if @course.is_running? %>
+    <h3 class="govuk-heading-m">Withdraw</h3>
+    <p class="govuk-body">Remove this course from Find and close applications.</p>
+    <%= link_to 'Withdraw this course', withdraw_provider_course_path(@provider.provider_code, @course.course_code), class: "govuk-link--destructive" %>
+  <% else %>
+    <h3 class="govuk-heading-m">Delete</h3>
```
fofr: Delete shouldn't have this header than the paragraph below, just the link.

Screen Shot 2019-05-10 at 15 27 53

PR #220, opened 2019-04-24
[XXXX] Add authenticate to users controller

tvararu:
We are not entirely sure how this works without this.


tvararu:
How this works without this: the Faraday connection object is cached between requests. We had this bug before.

PR #224, opened 2019-04-24
[1325] Redirect expired DFE sessions to signin again

tvararu:
### Context

When a session expires, the Omniauth middleware throws an exception. We should redirect users to signin when this happens.

Adds try/catch to relevant part of Omniauth initialiser.

### Guidance to review

I'm not sure if the `ActionController::InvalidAuthenticityToken` is just for CSRF issues (possibly what I was testing with locally) or for any kind of expired auth.

Have tested this locally by going to the `/transition-info` page, logging out in a separate tab, and clicking the green button.

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -16,7 +16,9 @@ module OmniAuth
 
       def callback_phase
         error = request.params['error_reason'] || request.params['error']
-        if error
+        if error == 'sessionexpired'
+          return redirect('/signin')
+        elsif error
           raise CallbackError.new(request.params['error'], request.params['error_description'] || request.params['error_reason'], request.params['error_uri'])
         elsif request.params['state'].to_s.empty? || request.params['state'] != stored_state
           # Monkey patch: Ensure a basic 401 rack response with no body or header isn't served
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -89,6 +89,10 @@ if Settings.dfe_signin.issuer.present?
       else
         @app.call(env)
       end
+    rescue ActionController::InvalidAuthenticityToken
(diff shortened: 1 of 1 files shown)


tvararu:
[Rescue expired DfE-Signin session](https://trello.com/c/3XBtne3A/1325-rescue-expired-dfe-signin-session)

dankmitchell:
This didn't work for me, screenshot attached

Something like the might solve it

https://github.com/DFE-Digital/manage-courses-frontend/blob/master/config/initializers/omniauth.rb#L20

```
if error == 'sessionexpired'
  return redirect('/signin')
else
  raise CallbackError.new(request.params['error'], request.params['error_description'] || request.params['error_reason'], request.params['error_uri'])
end
```

Screenshot 2019-04-25 at 08 08 45

misaka:
@dankmitchell I just tried this out locally and it worked for me. Are you doing something different than what Theo describes?

I like that this change is contained within something that looks like it's our concern (`class DfeSignIn`), and it seems like a more general solution in that whatever the reason is for an invalid token, send them off to the sign-in page.

config/initializers/omniauth.rb
```diff
@@ -87,7 +87,13 @@ def call(env)
         response.redirect('/dfe/sessions/new')
         response.finish
       else
-        @app.call(env)
+        begin
+          @app.call(env)
+        rescue ActionController::InvalidAuthenticityToken
+          response = Rack::Response.new
+          response.redirect('/signin')
+          response.finish
```
misaka: I'd be tempted to remove the repetition in generating a redirect here by creating a `redirect_to` method, maybe I'm used to Code Climate chirping up about it.

tvararu:
@dankmitchell updated 👍 this PR now addresses two separate issues.

PR #226, opened 2019-04-25
[1372] Handle client / server errors on accept_transition_info

tvararu:
### Context

The frontend will blow up if the API returns a 400 or a 500. The API will return 4xx when the user is already transitioned, and 500 for a variety of reasons.

Ensure users are always directed to the providers page.

### Guidance to review

The spec could be dried out, suggestions welcome!

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -2,5 +2,15 @@ class UsersController < ApplicationController
   def accept_transition_info
     User.member(current_user['user_id']).accept_transition_screen
     redirect_to providers_path
+  rescue JsonApiClient::Errors::ClientError, JsonApiClient::Errors::ServerError => e
+    e.env.body.delete("traces")
+    Raven.extra_context(e.env)
+    Raven.capture_exception(e)
+
+    if e.is_a? JsonApiClient::Errors::ClientError
+      redirect_to providers_path
+    else
+      raise e
+    end
   end
 end
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -1,0 +1,50 @@
+require 'rails_helper'
+
+describe UsersController, type: :controller do
+  let(:user) { jsonapi :user }
+
+  before do
+    stub_omniauth
+    stub_session_create
+    allow_any_instance_of(ApplicationController)
+      .to receive(:current_user)
+      .and_return('user_id' => user.id)
+    allow(Raven).to receive(:capture_exception)
+  end
+
+  describe 'GET #accept_transition_info' do
+    context "with working request" do
+      before do
+        stub_api_v2_request("/users/#{user.id}/accept_transition_screen", {}, :patch, 200)
+      end
+
+      it "redirects to providers index" do
+        get :accept_transition_info
+        expect(response).to redirect_to(providers_path)
+      end
+    end
+
+    context "with client error" do
+      before do
+        stub_api_v2_request("/users/#{user.id}/accept_transition_screen", {}, :patch, 400)
+      end
+
+      it "redirects to providers index" do
+        get :accept_transition_info
+        expect(Raven).to have_received(:capture_exception).with(instance_of(JsonApiClient::Errors::ClientError))
+        expect(response).to redirect_to(providers_path)
+      end
+    end
+
+    context "with server error" do
+      before do
+        stub_api_v2_request("/users/#{user.id}/accept_transition_screen", {}, :patch, 500)
+      end
+
+      it "redirects to providers index" do
+        expect { get :accept_transition_info }.to raise_error JsonApiClient::Errors::ServerError
+        expect(Raven).to have_received(:capture_exception).with(instance_of(JsonApiClient::Errors::ServerError))
+      end
+    end
+  end
+end


tvararu:
[Users receive 500 when attempting to transition when already transitioned](https://trello.com/c/27Kojyjg/1372-users-receive-500-when-attempting-to-transition-when-already-transitioned)

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,13 @@
 class UsersController < ApplicationController
+  rescue_from JsonApiClient::Errors::ClientError, with: :render_providers
+  rescue_from JsonApiClient::Errors::ServerError, with: :render_providers
+
   def accept_transition_info
     User.member(current_user['user_id']).accept_transition_screen
+    render_providers
+  end
+
+  def render_providers
```
benilovj: I like that we're not exposing errors to end-users, but it'd be good to send the errors to Sentry in this method, so we don't eat them

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,13 @@
 class UsersController < ApplicationController
+  rescue_from JsonApiClient::Errors::ClientError, with: :render_providers
+  rescue_from JsonApiClient::Errors::ServerError, with: :render_providers
```
misaka: I feel like this is too wide a net being cast here, I don't think we know that redirecting to providers is always going to be the right thing.

I think this should be a rescue section in `accept_transition_info` that logs the Sentry error in addition to just moving them along. Also, I worry a little bit about a circular redirect here in case the problem _isn't_ that the user has already accepted the terms, is there any way to make this more specific? Do we log an error message/condition we could use?

tvararu: Done

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,9 @@
 class UsersController < ApplicationController
   def accept_transition_info
     User.member(current_user['user_id']).accept_transition_screen
+  rescue StandardError => e
```
misaka: My only misgiving here is that any error here will turn into the user getting stuck on the `accept_transition_screen`. The only situation we know how to handle here is if the error has already accepted the transition screen, so can we find a more specific error to catch here? 

Seeing as this is an acute problem that we need to sort out sooner rather than later, I'd be happy if we used the `JsonApiClient::ClientError` or whatever we had before here until we workout a better way to do it.

app/controllers/users_controller.rb
```diff
@@ -1,6 +1,9 @@
 class UsersController < ApplicationController
   def accept_transition_info
     User.member(current_user['user_id']).accept_transition_screen
```
benilovj: Does the client library allow adding logic to/around the `accept_transition_screen`? It'd be handy to somehow encapsulate the server 422 response within the `User` model, rather than let that bleed out into the controller. Just a thought...

app/controllers/users_controller.rb
tvararu: > My only misgiving here is that any error here will turn into the user getting stuck on the accept_transition_screen.

The user won't be stuck, there's an `ensure` to redirect them below. Am I missing something?

app/controllers/users_controller.rb
tvararu: There's only the `on` and `request_method` options from what I gather: https://github.com/JsonApiClient/json_api_client/blob/e4b763f5579c4fae4b03dcc935b20627f5c14eaa/lib/json_api_client/resource.rb#L215

app/controllers/users_controller.rb
tvararu: And yeah, I can use `JsonApiClient::ClientError`, but I also want to catch `JsonApiClient::ServerError`, is this how that would work?

```
rescue JsonApiClient::ClientError => e
  Raven.capture_exception(e)
rescue JsonApiClient::ServerError => e
  Raven.capture_exception(e)
```

benilovj: I think it's possible to do

```ruby
rescue JsonApiClient::ClientError, JsonApiClient::ServerError => e
  Raven.capture_exception(e)
```

misaka: > The user won't be stuck, there's an ensure to redirect them below. Am I missing something?

@tvararu I'm imagining the situation where another error occurs that's not "The user has already accepted this screen.". A problem with the token, the backend not being happy with something ... as unlikely as this would be, the impact is pretty high. The steps I think would happen:

1. The user is shown the transition page, clicks on accept.
2. An error other than "they have already accepted" is fired, the ensure redirects them to "/providers".
3. The app tries to load providers, but sees that they haven't seen the transition info page (it was blocked above by the error), it redirects them to the transition page.
4. Go back to 1.

I know it's unlikely, but the impact is quite high (as bad as if not worse than an error that they may be able to recover from by hitting another page), and it's also relatively easy to protect against.

tvararu: @misaka done

heidar:
> The frontend will blow up if the API returns a 400 or a 500. The API will return 4xx when the user is already transitioned, and 500 for a variety of reasons.

Is this something we should be addressing on the backend rather than the frontend? At least not doing an error 40x if the user is already transitioned?

misaka:
> Is this something we should be addressing on the backend rather than the frontend? At least not doing an error 40x if the user is already transitioned?

Ya, we did discuss this ... it would probably be best, but at this point I think we're not sure it's worth the effort. But, yes, we should in general be trying to return more meaningful errors from the backend.

PR #227, opened 2019-04-25
[1337] Send DfE-SignIn ID to backend

heidar:
### Context

We need to send the DfE-SignIn ID to the backend so we're not relying on only email for user lookup.

- Send DfE-SignIn ID to the backend in the JWT payload
- Remove special token handling in development to match backend changes

### Guidance to review

Tested on MacOS Mojave 10.14.4. (For real this time, it works!)

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -69,18 +69,13 @@ private
   end
 
   def add_token_to_connection
-    if Settings.authentication.algorithm == 'plain-text'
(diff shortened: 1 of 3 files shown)


heidar:
[Users who updated their emails in DfE Signin can't use Publish](https://trello.com/c/PT3mxHeE/1337-users-who-updated-their-emails-in-dfe-signin-cant-use-publish)

PR #229, opened 2019-04-26
[1370] Add draper

dankmitchell:
### Context
add-draper-now

- Add and setup draper
- Convert `formatted_date` to draper method
- Hide `last_published_date` if nil as this doesn't seem relevant to display when `nil`

### Guidance to review
Examples
- Happy Paths  `https://localhost:3000/organisations/2AT/courses/35L5/description`
- Sad path `https://localhost:3000/organisations/2HO/courses/3B7S/description`

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -1,3 +1,7 @@
 class CourseDecorator < ApplicationDecorator
   delegate_all
+
+  def last_published_at
+    object&.last_published_at&.to_date&.strftime("%-d %B %Y")
+  end
 end
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -1,0 +1,3 @@
+class CourseDecorator < ApplicationDecorator
+  delegate_all
+end
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -1,7 +1,7 @@
 class CourseDecorator < ApplicationDecorator
   delegate_all
 
-  def last_published_at
+  def formatted_last_published_at
     object.last_published_at&.to_date&.strftime("%-d %B %Y")
   end
 end
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -2,6 +2,6 @@ class CourseDecorator < ApplicationDecorator
   delegate_all
 
   def last_published_at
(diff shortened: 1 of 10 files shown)


app/views/courses/_status_panel.html.erb
```diff
@@ -29,8 +29,11 @@
     <% end %>
   </p>
 
-  <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
-  <p class="govuk-body">Last published:<br><%= formatted_date(@course.last_published_at) %></p>
+  <% if @course.last_published_at.present? %>
+    <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
+    <p class="govuk-body">Last published:<br><%= @course.last_published_at %></p>
```
tvararu: I don't particularly like here that someone who isn't familiar with Draper or doesn't know that `@course` is the decorated course can find themselves scratching their head as to how exactly `last_published_at` is being shown to the user as `5 March 2019` instead of `2019-03-05T14:42:34Z`.

I see that Draper doesn't require that you implement a decorated method that is called exactly like the actual model attribute.

What do you think of `formatted_last_published_at`?

This is a minor suggestion.

dankmitchell: I agree and had that to begin with but didn’t like the length of it ha. I’ll change it

tvararu: I think it's long too, it's just the first thing that came to mind...

misaka: I'm not bothered by long names, and personally think that this one is Good Enough (:tm:). Also, good suggestion to make it clearer which attributes are specially formatted.

app/decorators/course_decorator.rb
```diff
@@ -0,0 +1,7 @@
+class CourseDecorator < ApplicationDecorator
+  delegate_all
+
+  def last_published_at
+    object&.last_published_at&.to_date&.strftime("%-d %B %Y")
```
misaka: I don't think we should safeguard `object` being nil. i.e. just simply `object.last_published` without the `&`.

tvararu: This was a hold-over since before `last_published_at` was implemented I believe, can be removed now.

misaka:
@dankmitchell I took the liberty of changing the two outstanding comments. Sorry but we're just hoping to get a deploy going this afternoon and this change has a bug fix we need.

PR #228, opened 2019-04-25
[1369] Pre-populate Google forms

benilovj:
### Context
We use Google forms to allow publishers to submit certain information to us. Google permits pre-populating the forms which reduces the data entry burden on the publisher, and improves the quality of the data submitted to us.

### Changes proposed in this pull request

- Add code for constructing Google form links which pre-populate provider code and publisher email on:
  - the 'add new course' form
  - the 'add new location' form
- Introduce extra settings which are used to construct the pre-population URLs
- Stub some of the feature specs at a lower level, so that the user email is available in the templates

### Guidance to review
This makes the Google form link building a bit more complex, but hopefully that's all encapsulated in the helper.

PR #230, opened 2019-04-27
[1369] Clean up feature specs with page objects

benilovj:
### Context
#228 was merged without using page objects in any of the changed code. This PR is cleaning this up.

- add some extra structure into the templates to make testing easier/more robust
- push selectors out of the feature specs and into page objects

### DIFF
--- a/app/views/courses/index.html.erb
+++ b/app/views/courses/index.html.erb
@@ -30,15 +30,21 @@
   <p class="govuk-body-s govuk-!-margin-bottom-6">You’ll be taken to a Google Form to fill in your course details.</p>
 <% end %>
 
-<%= render partial: 'course_table', locals: { courses: @self_accredited_courses } if @self_accredited_courses %>
+<% if @self_accredited_courses %>
+  <section data-qa="courses__table-section">
+    <%= render partial: 'course_table', locals: { courses: @self_accredited_courses } %>
+  </section>
+<% end %>
 
 <% @courses_by_accrediting_provider.each do |accrediting_provider, courses| %>
-  <h2 class="govuk-heading-m">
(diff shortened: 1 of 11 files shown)


app/views/courses/index.html.erb
```diff
@@ -30,15 +30,21 @@
   <p class="govuk-body-s govuk-!-margin-bottom-6">You’ll be taken to a Google Form to fill in your course details.</p>
 <% end %>
 
-<%= render partial: 'course_table', locals: { courses: @self_accredited_courses } if @self_accredited_courses %>
+<% if @self_accredited_courses %>
+  <section data-qa="courses__courses-table-section">
```
dankmitchell: Don't think we need to repeat `courses` here. `courses__table-section` is good enough

app/views/courses/index.html.erb
```diff
@@ -30,15 +30,21 @@
...
 
 <% @courses_by_accrediting_provider.each do |accrediting_provider, courses| %>
-  <h2 class="govuk-heading-m">
-    <span class="govuk-caption-m">Accredited body</span>
-    <%= accrediting_provider %>
-  </h2>
-
-  <%= render partial: 'course_table', locals: { courses: courses } %>
+  <section data-qa="courses__courses-table-section">
+    <h2 class="govuk-heading-m">
```
dankmitchell: No need to add an extra level of HTML. I'd just add the data attribute on the `<h2>`

dankmitchell: NVM, I see the value of this now 👌

app/views/courses/index.html.erb
benilovj: Fixed in 6e36e030540e6540f009da0072eba6453f02d2ec

PR #232, opened 2019-04-29
[1383] Move view logic into decorators

dankmitchell:
### Context
Refactoring helpers and view logic into decorators aka view models

- Reduce helper methods
- Reduce logic in views


app/controllers/courses_controller.rb
```diff
@@ -27,10 +27,13 @@ def index
...
     # rubocop:enable Style/MultilineBlockChain
 
-    @self_accredited_courses = @courses_by_accrediting_provider.delete(@provider[:provider_name])
+    @courses_by_accrediting_provider = CourseDecorator.decorate_collection(@courses_by_accrediting_provider)
+    @self_accredited_courses = CourseDecorator.decorate_collection(@courses_by_accrediting_provider.delete(@provider[:provider_name]))
   end
 
-  def show; end
+  def show
+    @course = @course.decorate
```
heidar: Would it be nicer to `decorate` in the `build_` methods when we can?

app/controllers/courses_controller.rb
```diff
@@ -51,6 +51,7 @@ def build_course
       .where(provider_code: @provider_code)
       .find(params[:code])
       .first
+      .decorate
```
heidar: `&.decorate`?

app/views/courses/_status_panel.html.erb
```diff
@@ -1,37 +1,24 @@
...
-    <% if @provider.opted_in? %>
-      (<%= link_to 'Edit', vacancies_provider_course_path(code: @course.course_code), class: 'govuk-link' %>)
-    <% end %>
+    <%= @course.vacancies %>
   </p>
 
   <% if @course.last_published_at.present? %>
     <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
-    <p class="govuk-body">Last published:<br><%= @course.formatted_last_published_at %></p>
+    <p class="govuk-body">Last published:<br><%= l(@course.last_published_at.to_date) %></p>
```
heidar: Should we do `&.to_date`?

app/views/courses/_course_table_row.html.erb
```diff
@@ -1,41 +1,30 @@
...
-    <span class="govuk-body-s"><%= course.attributes[:description] %></span>
+    <span class="govuk-body-s"><%= course.description %></span>
   </td>
   <td class="govuk-table__cell" data-qa="courses-table__ucas-status">
-    <%= course_ucas_status(course) %>
+    <%= course.ucas_status %>
   </td>
   <td class="govuk-table__cell" data-qa="courses-table__content-status">
-    <%= render partial: 'content_status_tag', locals: { course: course } unless course.not_running? %>
+    <% unless course.not_running? %>
```
heidar: Can we do this check in the presenter and return nil if it's not running?

app/views/courses/_basic_details_tab.html.erb
```diff
@@ -116,7 +114,7 @@
           Course starts
         </dt>
         <dd class="govuk-summary-list__value" data-qa="course__start_date">
-          <%= course_start_date(@course) %>
+          <%= l(@course.start_date.to_date, format: :short) %>
```
heidar: `&.to_date`?

app/decorators/course_decorator.rb
```diff
@@ -1,7 +1,124 @@
...
+      'Salaried'
+    when 'apprenticeship'
+      'Teaching apprenticeship (with salary)'
+    when 'fee'
+      'Fee paying (no salary)'
+    end
+  end
+
+  def apprenticeship
+    object.funding == 'apprenticeship' ? 'Yes' : 'No'
```
heidar: `@object.apprenticeship?`

app/controllers/courses_controller.rb
dankmitchell: Done

app/views/courses/_course_table_row.html.erb
dankmitchell: Done

app/decorators/course_decorator.rb
dankmitchell: Done

app/views/courses/_basic_details_tab.html.erb
dankmitchell: Done

app/views/courses/_status_panel.html.erb
dankmitchell: Done

app/controllers/courses_controller.rb
```diff
@@ -51,6 +51,7 @@ def build_course
       .where(provider_code: @provider_code)
       .find(params[:code])
       .first
+      &.decorate
```
misaka: I know this is tempting, but this means that when we start to use @course in any of the actions, we won't have a JSON API client resource object, we'll have a decorated object which will behave differently. I think decorated object should only be used in views (or anywhere else explicitly needed), and would rather have an `after_action` hook to decorate, or use `@decorated_course` (which means more changes to views, 🤢 ).

Also: https://github.com/drapergem/draper#when-to-decorate-objects

dankmitchell: Done

PR #235, opened 2019-04-30
[1383] Fix course.on_find decorated method

tvararu:
### Context

Product review.

Pass in the provider to `not_on_find` so that it can work in the courses index page, where courses do not have a provider attached to the `object`.

### Guidance to review

The existing tests already unit test this functionality; arguably this could do with an integration test.


tvararu:
[Extract some view logic into draper decorators [timebox: 1/2 day]](https://trello.com/c/QpyS6i5f/1383-extract-some-view-logic-into-draper-decorators-timebox-1-2-day)

PR #240, opened 2019-05-01
[1409] Add guard

tvararu:
### Context

This allows us to automatically run rspec on files that have changed. Uses spring so they run faster.

Add guard-rubocop, and ignore generated files when running rubocop.


defong:
[Add guard to manage-courses-frontend](https://trello.com/c/jXZhXiic/1409-add-guard-to-manage-courses-frontend)

PR #242, opened 2019-05-01
[1415] redirect users back to where they were going on signin

misaka:
https://trello.com/c/s3cyWAso/1415-dont-always-redirect-to-root-on-sign-in

### Context

On sign-in users would be redirected to `/` no matter where they were originally going.

Save where users were going before the sign-in and send them there after they are signed in. If not saved for any reason, they will still go to `/`.

### DIFF
--- a/spec/support/helpers.rb
+++ b/spec/support/helpers.rb
@@ -27,7 +27,7 @@ module Helpers
     end
   end
 
-  def stub_session_create(user: double(id: 1))
+  def stub_session_create(user: double(id: 1, 'opted_in?': false))
     allow(Session).to receive(:create).and_return(user)
   end
 
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -31,6 +31,7 @@ class ApplicationController < ActionController::Base
 
       set_user_session if current_user['user_id'].blank?
     else
+      session[:redirect_back_to] = request.path
       redirect_to '/signin'
     end
   end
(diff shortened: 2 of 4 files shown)


spec/support/helpers.rb
```diff
@@ -27,7 +27,7 @@ def stub_omniauth(disable_completely: true, user: nil)
     end
   end
 
-  def stub_session_create(user: double(id: 1))
+  def stub_session_create(user: double(id: 1, 'opted_in?': false))
```
benilovj: Everyone's opted in now, I think this should default to `true`

tvararu:
[Authorization failure when updating site statuses](https://trello.com/c/OnBKJMq6/1302-authorization-failure-when-updating-site-statuses)

tvararu:
[Don't always redirect to root on sign-in](https://trello.com/c/s3cyWAso/1415-dont-always-redirect-to-root-on-sign-in)

dankmitchell:
If we are doing this then we should also remove the `disable_completely` option from the `stub_omniauth` method here https://github.com/DFE-Digital/manage-courses-frontend/blob/master/spec/support/helpers.rb#L24

spec/support/helpers.rb
dankmitchell: Agreed. I'm making the bigger change here - https://github.com/DFE-Digital/manage-courses-frontend/pull/244

dankmitchell:
> If we are doing this then we should also remove the `disable_completely` option from the `stub_omniauth` method here https://github.com/DFE-Digital/manage-courses-frontend/blob/master/spec/support/helpers.rb#L24

Created a follow up story for this work.

PR #243, opened 2019-05-01
[1302] fix authorization errors

misaka:
https://trello.com/c/OnBKJMq6/1302-authorization-failure-when-updating-site-statuses

### Context

Apparently Faraday is not thread-safe and the way we were setting the
authorization token on the connection was opening us up to a race condition.

This plugs that hole by using the Thread key-value context to save the user
token, and using that when we generate the connection to mcb.

### Guidance to review

This PR requires #242 and is based off of that one.

### DIFF
--- a/app/models/base.rb
+++ b/app/models/base.rb
@@ -4,7 +4,7 @@ class Base < JsonApiClient::Resource
   # the user was first authenticated.
   class MCBConnection < JsonApiClient::Connection
     def run(request_method, path, params: nil, headers: {}, body: nil)
-      authorization = "Bearer #{Thread.current[:manage_courses_backend_token]}"
+      authorization = "Bearer #{Thread.current.fetch(:manage_courses_backend_token)}"
       super(
         request_method,
         path,
--- a/app/models/base.rb
+++ b/app/models/base.rb
@@ -1,5 +1,22 @@
 class Base < JsonApiClient::Resource
+  # This connection class grabs the bearer token from the thread key-value
+  # store and adds it to our connection. This token was saved there for us when
+  # the user was first authenticated.
+  class MCBConnection < JsonApiClient::Connection
+    def run(request_method, path, params: nil, headers: {}, body: nil)
+      authorization = "Bearer #{Thread.current[:manage_courses_backend_token]}"
+      super(
+        request_method,
+        path,
+        params: params,
+        headers: headers.update('Authorization' => authorization),
+        body: body
+      )
+    end
+  end
+
   include Draper::Decoratable
 
   self.site = "#{Settings.manage_backend.base_url}/api/v2/"
+  self.connection_class = MCBConnection
 end
--- a/spec/features/backend_requests_spec.rb
+++ b/spec/features/backend_requests_spec.rb
@@ -6,8 +6,46 @@ describe 'requests made to mc-be' do
     let(:provider2) { jsonapi :provider }
 
     it 'is thread-safe' do
+      # The api calls in this test should be completely isolated:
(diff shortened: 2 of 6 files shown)


tvararu:
[Authorization failure when updating site statuses](https://trello.com/c/OnBKJMq6/1302-authorization-failure-when-updating-site-statuses)

app/models/base.rb
```diff
@@ -1,5 +1,22 @@
 class Base < JsonApiClient::Resource
+  # This connection class grabs the bearer token from the thread key-value
+  # store and adds it to our connection. This token was saved there for us when
+  # the user was first authenticated.
+  class MCBConnection < JsonApiClient::Connection
+    def run(request_method, path, params: nil, headers: {}, body: nil)
+      authorization = "Bearer #{Thread.current[:manage_courses_backend_token]}"
```
tvararu: Should we throw something up to Sentry if this lookup finds nil?

misaka: Ya, good call! We should be using `.fetch`.

misaka:
@tvararu 

> Should we throw something up to Sentry if this lookup finds nil?

Added a `.fetch`. I couldn't think of a way to test it outside of a unit test as I don't see how this situation would occur, so I'm not sure we need to test it.

defong:
> @tvararu
> 
> > Should we throw something up to Sentry if this lookup finds nil?
> 
> Added a `.fetch`. I couldn't think of a way to test it outside of a unit test as I don't see how this situation would occur, so I'm not sure we need to test it.

passing thought/no idea

```ruby
Thread.current[:manage_courses_bac
kend_token] == nil raise hell
```
?

spec/features/backend_requests_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+                           .and_wrap_original do |method, *args|
+        # This simulates a bug we found in the wild:
+        #
+        #   - Request 1: start and set the authorisation token for user2 on the
+        #                Faraday connection
+        #   - Request 1: invoke the controller action which takes a little time
+        #                opening it up to a race condition error
+        #   - Request 2: start and set the authorisation token for user2 on the
+        #                Faraday connection
+        #   - Request 1: connect to manage-courses backend with user1's token
```
timabell: I'm assuming at this point there was now a case of the wrong token being picked up. I think it would be worth adding another line to this comment explaining the failure state that was seen as I can't immediately see what that would be from the test code.

misaka:
> passing thought/no idea
> 
> ```ruby
> Thread.current[:manage_courses_bac
> kend_token] == nil raise hell
> ```
> 
> ?

This is what `.fetch` does, it allows a default to be provided, but if not it raises an exception.

spec/features/backend_requests_spec.rb
```diff
@@ -0,0 +1,60 @@
...
+          allow(JWT).to receive(:encode).and_return('tokenUser2')
+          visit("/organisations/#{provider2.provider_code}")
+        end
+        thread.join
+        method.call(*args)
+      end
+
+      allow(Provider).to receive(:find)
+                           .with(provider2.provider_code)
+                           .and_call_original
```
timabell: is this required? (open question)

misaka: Nope, looks left-over code.

misaka: Whoops. Nope, wrong. This is required because of the `allow` above ... we need to allow `find` to be called with either `provider1.provider_code` or `provide2.provider_code`.

PR #244, opened 2019-05-01
[1418] Remove opted_in

dankmitchell:
### Context
All providers are now `opted_in`

Remove `opted_in`

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -5,9 +5,9 @@ class CourseDecorator < ApplicationDecorator
     "#{object.name} (#{object.course_code})"
   end
 
-  def vacancies(provider = object.provider)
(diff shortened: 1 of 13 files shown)


app/decorators/course_decorator.rb
```diff
@@ -112,7 +112,7 @@ def unpublished_status_hint
 
   def not_on_find(provider)
```
benilovj: You can probably remove the `provider` parameter here.

app/decorators/course_decorator.rb
```diff
@@ -7,7 +7,7 @@ def name_and_code
 
   def vacancies(provider = object.provider)
```
benilovj: You can probably remove the `provider` parameter here.

PR #249, opened 2019-05-03
[1431] Render the 404 page when provider not found

dankmitchell:
### Context
Showing the user nice a 404 page when record not found - https://github.com/DFE-Digital/manage-courses-backend/pull/332

- Rescue `JsonApiClient::Errors::NotFound` and render our nice 404 page

### Guidance to review
Example `https://localhost:3000/organisations/T9`

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -1,4 +1,6 @@
 class ProvidersController < ApplicationController
+  rescue_from JsonApiClient::Errors::NotFound, with: :not_found
+
   def index
     @providers = Provider.all
     render_manage_ui if @providers.empty?
--- a/spec/requests/providers_spec.rb
+++ b/spec/requests/providers_spec.rb
@@ -1,0 +1,25 @@
+require 'rails_helper'
+
+describe 'Providers', type: :request do
+  before do
+    stub_omniauth
+    stub_session_create
+  end
+
+  describe 'GET show' do
+    it 'render providers show' do
+      provider = jsonapi(:provider)
+      stub_api_v2_request("/providers/#{provider.provider_code}", provider.render)
+      get(provider_path(provider.provider_code))
+      expect(response.body).to include(provider.provider_name)
+    end
+
+    context 'provider does not exist' do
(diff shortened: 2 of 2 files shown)


heidar:
[Handle not found errors with 404 responses](https://trello.com/c/CZFmryNP/1431-handle-not-found-errors-with-404-responses)

dankmitchell:
[Fix provider page causing a 500](https://trello.com/c/65wXqnCL/1424-fix-provider-page-causing-a-500)

spec/support/helpers.rb
```diff
@@ -24,6 +24,7 @@ def stub_omniauth(disable_completely: true, user: nil)
     # Temp solution until we implement `return_url` for DfE sign-in
     if disable_completely
       allow_any_instance_of(ApplicationController).to receive(:authenticate)
+      Thread.current[:manage_courses_backend_token] = 'foobar'
```
dankmitchell: Temporary change to get the specs green. But this will be removed in the next PR that removes this method completely

app/controllers/providers_controller.rb
```diff
@@ -1,11 +1,13 @@
...
+
   def index
     @providers = Provider.all
     render_manage_ui if @providers.empty?
     redirect_to provider_path(@providers.first.provider_code) if @providers.size == 1
   end
 
   def show
-    @provider = Provider.find(params[:code]).first.attributes
+    @provider = Provider.find(params[:code]).first
```
dankmitchell: The exception wasn’t being caught in the action so had todo it at controller level.

app/controllers/providers_controller.rb
```diff
@@ -1,4 +1,6 @@
 class ProvidersController < ApplicationController
+  rescue_from JsonApiClient::Errors::NotFound, with: :not_found
```
misaka: Similar to mc-be, I think we should have a before-hook setter that finds `@provider` and returns an appropriate error when it's not found.

dankmitchell: We tried this and it didnt work so we put it at controller level

PR #251, opened 2019-05-06
1422 redirect to root if timed out user attempts to signout

dankmitchell:
### Context
If a user session has expired making `current_user` is nil then a user attempts to signout from DfE-Signin a server error is thrown because DfE-Signin requires the users `id_token` which is stored in the `current_user` object

Check for the presence of `current_user` and redirect to root as their session has already expired

Do we need to think about a more robust strategy for this i.e. persist the `id_token` on the user when the signin?

### Guidance to review
Delete the session in your browser and hit `signout`

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -19,7 +19,11 @@ class SessionsController < ApplicationController
   end
 
   def signout
-    redirect_to "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=#{current_user['credentials']['id_token']}&post_logout_redirect_uri=#{Settings.dfe_signin.base_url}/auth/dfe/signout"
+    if current_user.present?
+      redirect_to "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=#{current_user['credentials']['id_token']}&post_logout_redirect_uri=#{Settings.dfe_signin.base_url}/auth/dfe/signout"
+    else
+      redirect_to root_path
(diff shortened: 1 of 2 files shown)


dankmitchell:
[SessionsController timeout error](https://trello.com/c/h0oXxCZI/1422-sessionscontroller-timeout-error)

app/controllers/sessions_controller.rb
```diff
@@ -19,7 +19,11 @@ def create
   end
 
   def signout
-    redirect_to "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=#{current_user['credentials']['id_token']}&post_logout_redirect_uri=#{Settings.dfe_signin.base_url}/auth/dfe/signout"
+    if current_user.present?
+      redirect_to "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=#{current_user['credentials']['id_token']}&post_logout_redirect_uri=#{Settings.dfe_signin.base_url}/auth/dfe/signout"
+    else
+      redirect_to root_path
+    end
```
misaka: Is this testable?

misaka:
> Do we need to think about a more robust strategy for this i.e. persist the id_token on the user when the signin?

Why would we bother to do this? Just to guarantee that they're signed-out of DFE Sign-In? I'm not sure I see the value in this, offhand.

misaka:
This approach looks Good Enough (:tm:) to me.

PR #254, opened 2019-05-07
[1442] return 404 when a course does not exist

misaka:
### Context

Users are getting 500 errors when a course does not exist.

Give the users a 404 biscuit instead of 500s.


misaka:
[Return 404 when course cannot be found](https://trello.com/c/XkcNZiwe/1442-return-404-when-course-cannot-be-found)

misaka:
@dankmitchell This was what I was thinking of for handling not found errors in the `build_*` methods.

PR #255, opened 2019-05-07
[1444] Rescue from 500s by showing the generic 500 page

tvararu:
### Context

Currently, the users get a blank page. This is an improvement.

### Guidance to review

The test I wrote resides in `view_pages_spec.rb ` which is probably not the best place?

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception, except: :not_found
+  rescue_from StandardError, with: :internal_server_error
   rescue_from JsonApiClient::Errors::NotAuthorized, with: :render_manage_ui
   rescue_from JsonApiClient::Errors::AccessDenied, with: :render_manage_ui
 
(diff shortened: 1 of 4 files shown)


tvararu:
[Rescue from generic 500s with the existing 500 page](https://trello.com/c/buZhyJt3/1444-rescue-from-generic-500s-with-the-existing-500-page)

misaka:
> Is there enough logging to denote where the user came from and where they going, are they logged in etc.

This should be done with request logging. Would be nice if AI could work.

app/controllers/application_controller.rb
```diff
@@ -1,5 +1,6 @@
 class ApplicationController < ActionController::Base
   protect_from_forgery with: :exception, except: :not_found
+  rescue_from StandardError, with: :internal_server_error
```
misaka: Does this still capture to Raven?

misaka:
> The test I wrote resides in pages_controller_spec which is probably not the best place?

`spec/features/view_pages_spec.rb` where it is now seems reasonable enough.

app/controllers/application_controller.rb
tvararu: It does not, I have addressed it in the method below, great spot!

tvararu:
@defong this doesn't seek to improve the existing logging, just make sure that users see an appropriate error message while maintaining the same level of logging we do currently have. :)

PR #257, opened 2019-05-08
[1439] set user id and sign-in user id in Raven contexts

misaka:
### Context

Having the user id and sign-in user id helps us better debug errors.

Add the user id and sign-in user id to what's captured by Raven.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -25,10 +25,14 @@ class ApplicationController < ActionController::Base
 
   def authenticate
     if current_user
+      assign_sentry_contexts
       add_token_to_connection
       set_has_multiple_providers
 
-      set_user_session if current_user['user_id'].blank?
+      if current_user['user_id'].blank?
+        set_user_session
+        Raven.user_context(id: current_user['user_id'])
+      end
     else
       session[:redirect_back_to] = request.path
       redirect_to '/signin'
@@ -55,6 +59,7 @@ class ApplicationController < ActionController::Base
 private
 
   def set_user_session
+    # TODO: we should return a session object here with a 'user' attached to id.
     user = Session.create(first_name: current_user_info[:first_name],
                           last_name: current_user_info[:last_name])
     session[:auth_user]['user_id'] = user.id
@@ -82,5 +87,10 @@ private
                        Settings.manage_backend.algorithm)
 
     Thread.current[:manage_courses_backend_token] = token
+  end
+
+  def assign_sentry_contexts
+    Raven.user_context(id: current_user['user_id'])
+    Raven.tags_context(sign_in_user_id: current_user.fetch('uid'))
   end
 end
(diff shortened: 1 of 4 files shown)


app/controllers/application_controller.rb
```diff
@@ -25,10 +25,14 @@ def current_user
...
   def authenticate
     if current_user
+      assign_sentry_contexts
       add_token_to_connection
       set_has_multiple_providers
 
-      set_user_session if current_user['user_id'].blank?
+      if current_user['user_id'].blank?
+        set_user_session
+        Raven.user_context(id: current_user['user_id'])
```
heidar: Hmm, can we just run `assign_sentry_contexts` on line 35 after the `if`, so it'll always get set when we have data?

app/controllers/application_controller.rb
```diff
@@ -83,4 +88,9 @@ def add_token_to_connection
 
     Thread.current[:manage_courses_backend_token] = token
   end
+
+  def assign_sentry_contexts
+    Raven.user_context(id: current_user['user_id'])
+    Raven.tags_context(sign_in_user_id: current_user.fetch('uid'))
```
heidar: Should probably use safe navigation here since I'm pretty sure `current_user` could be `nil`.

app/controllers/sessions_controller.rb
```diff
@@ -8,8 +8,11 @@ def new
   def create
     session[:auth_user] = auth_hash
 
+    Raven.tags_context(sign_in_user_id: current_user.fetch('uid'))
```
heidar: `skip_before_action :authenticate` causes `current_user` to not get set here, so I don't think this does anything but set these to `nil`. I'm surprised this doesn't blow up actually.

app/controllers/application_controller.rb
misaka: I wanted to have the context set ASAP ... with the idea being that if `add_token_to_connection` or `set_has_multiple_providers` blow up, we still have _some_ context.

timabell:
[Allow mc-support to grant access requests through mc-be](https://trello.com/c/VdmjSTgi/1363-allow-mc-support-to-grant-access-requests-through-mc-be)

app/controllers/sessions_controller.rb
misaka: If I'm not mistaken, `current_user` just pulls out `session[:auth_user]` (https://github.com/DFE-Digital/manage-courses-frontend/pull/257/files#diff-55c5b7aecfb519d0e4880eaf2788eb6eR23), so once that is set it's save to access `current_user`.

app/controllers/application_controller.rb
misaka: I haven't seen how yet ... considering `assign_sentry_contexts` is called in both places after `session[:auth_user]` is set. Can we test this?

PR #262, opened 2019-05-10
Copy changes around Apply refresh timescales.  

avinhurry:
### Context 
UCAS now refresh from our API more frequently.

### DIFF
--- a/app/views/courses/vacancies/edit.html.erb
+++ b/app/views/courses/vacancies/edit.html.erb
@@ -52,7 +52,7 @@
             </div>
           </div>
         </fieldset>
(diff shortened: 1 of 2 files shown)


app/views/courses/vacancies/edit.html.erb
```diff
@@ -52,7 +52,7 @@
             </div>
           </div>
         </fieldset>
-        <p class="govuk-body">Changes will appear on Find straight away. UCAS Apply will be updated overnight.</p>
+        <p class="govuk-body">Changes will appear on Find straight away. UCAS Apply will be updated within 2 hours.</p>
```
fofr: Awesome 🎉

PR #270, opened 2019-05-13
1463 add courses fees page

dankmitchell:
### Context
Re-creating `/organisation/2cf/course/self/2yn6/fees` in rails

Add `/fees` to a course

c# | rails
:--:|:--:
c#|rails

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -1,7 +1,7 @@
 class CoursesController < ApplicationController
   decorates_assigned :course
-  before_action :build_course, only: %i[show description about requirements delete withdraw publish]
-  before_action :build_provider, only: %i[show description about requirements publish]
+  before_action :build_course, except: :index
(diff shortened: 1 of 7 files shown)


dankmitchell:
[Add edit course length and fees page](https://trello.com/c/qrMfyr7Q/1463-add-edit-course-length-and-fees-page)

benilovj:
In the Rails screenshot, the fee is "9000.0" (extra zero)... will that be resolved with the backend work?

dankmitchell:
@misaka do you think I should edit the backend to return a rounded number amount? Or make the frontend do it?

benilovj:
> @misaka do you think I should edit the backend to return a rounded number amount? Or make the frontend do it?

Just discussed it with @misaka IRL, we've come to the conclusion that the frontend should explicitly define how it formats the number, even if it comes through in the right format from the backend.

PR #272, opened 2019-05-14
[XXXX] Only display the 500 page in production mode

tvararu:
Better solution, turn on the relevant Rails flag.

Tests removed because there's no point in testing Rails itself.


misaka:
Ya, this gets the job done, or should ... but it smells wrong.

defong:
Just wondering?

https://github.com/DFE-Digital/manage-courses-backend/blob/47ad7ceac528681ca0b53c99d735cfd1d984adfb/config/application.rb#L31

tvararu:
@defong I think that's for something else, but we could add some on the frontend too if necessary

PR #274, opened 2019-05-14
[1458] Switch factories to use the JSONAPI renderer

heidar:
### Context

Using the `jsonapi-rb` gems to render JSONAPI responses based on our factories. 

A few lines of metaprogramming magic to create serializers and make our models able to turn themselves into JSONAPI hashes.

### Guidance to review

TODO:

* Still need to do this for the rest of the factories (should be easy)
* Relations (easy to add in)
* Custom meta attributes (might be easier to declare the serializers manually for these)
* Need to support the `includes` parameter (should be easy)
* Need to figure out how to make this work with `build_list` (a bit of a pain)

### DIFF
--- a/spec/support/define_factory_jsonapi_render_methods.rb
+++ b/spec/support/define_factory_jsonapi_render_methods.rb
@@ -1,0 +1,25 @@
+FactoryBot.define do
+  # This defines an after build callback that goes on ALL factories.
+  after :build do |record|
+    # This defines an instance method called to_jsonapi on the object we just
+    # created with the factory.
+    record.define_singleton_method(:to_jsonapi) do
+      # Create a string that holds the name of the class, for example
+      # "UserSerializer"
+      serializer_class = "#{record.class}Serializer"
+
+      renderer = JSONAPI::Serializable::Renderer.new
+      renderer.render(
+        record,
+        class: {
+          # This tells the renderer what serializers to use. The key is going
+          # to be the name of the class as a symbol, and the value is the
+          # serializer class.
+          #
+          # For example: User: UserSerializer
+          record.class.name.to_sym => serializer_class.constantize
+        }
+      )
+    end
+  end
+end
--- a/spec/support/initialize_factory_serializers.rb
+++ b/spec/support/initialize_factory_serializers.rb
@@ -1,0 +1,24 @@
+FactoryBot.factories.each do |factory|
+  next unless factory.name.in? []
+
+  # Create a string that holds the name of the class, for example
+  # "UserSerializer"
+  serializer_class_name = "#{factory.name.capitalize}Serializer"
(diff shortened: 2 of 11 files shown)


spec/support/initialize_factory_serializers.rb
```diff
@@ -0,0 +1,14 @@
+FactoryBot.factories.each do |factory|
+  next unless factory.name.in? %i[user]
+
+  serializer_klass = "#{factory.name.capitalize}Serializer"
+
+  Object.const_set(
```
defong: Magic code :heart_eyes:

misaka: Ya, can we document what's happening here please?

misaka: I _think_ that you're defining a new class on-the-fly here ... is there a clearer way to do it? Even moving the `Class.new` out of the `Object.const_set` might help imo.

spec/support/define_factory_jsonapi_render_methods.rb
```diff
@@ -0,0 +1,15 @@
+FactoryBot.define do
+  after :build do |record|
```
misaka: This little gem could also use a little explanation.

spec/support/initialize_factory_serializers.rb
```diff
@@ -0,0 +1,24 @@
+FactoryBot.factories.each do |factory|
+  next unless factory.name.in? %i[user]
+
+  # Create a string that holds the name of the class, for example
+  # "UserSerializer"
+  serializer_class_name = "#{factory.name.capitalize}Serializer"
+
+  # Create a new class (of type Class) and make it inherit from the JSONAPI
+  # serializer resource. Everything in the block is realted to the serializer.
```
defong: realted => related

```suggestion
  # serializer resource. Everything in the block is related to the serializer.
```

spec/support/initialize_factory_serializers.rb
```diff
@@ -0,0 +1,24 @@
...
+  # "users".
+  # We set the attributes to render out to the attributes we define in the
+  # factory.
+  serializer_class      = Class.new(JSONAPI::Serializable::Resource) do
+    type(factory.name.to_s.pluralize)
+    attributes(*FactoryBot.attributes_for(factory.name).keys)
+  end
+
+  # Create a new constant with the class name and make it point to the class
+  # we created. For example UserSerializer, and it would point to the class we
```
defong:
```suggestion
  # we created. For example "UserSerializer", and it would point to the class we
```

heidar:
[Fix tests in mc-fe](https://trello.com/c/wBdaVsDJ/1458-fix-tests-in-mc-fe)

PR #275, opened 2019-05-14
Use live search URL

fofr:
Use https://find-postgraduate-teacher-training.education.gov.uk instead of https://www.find-postgraduate-teacher-training.service.gov.uk.

`https://www.find-postgraduate-teacher-training.service.gov.uk` works but isn't the official URL yet.

The [start page](https://www.gov.uk/guidance/find-postgraduate-teacher-training-courses-in-england) on GOV.UK links to https://find-postgraduate-teacher-training.education.gov.uk

### DIFF
--- a/config/settings/production.yml
+++ b/config/settings/production.yml
@@ -8,4 +8,4 @@ manage_backend:
 manage_ui:
   base_url: https://www.publish-teacher-training-courses.service.gov.uk
 search_ui:
-  base_url: https://www.find-postgraduate-teacher-training.service.gov.uk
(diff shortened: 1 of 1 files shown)


config/settings/production.yml
```diff
@@ -8,4 +8,4 @@ manage_backend:
 manage_ui:
   base_url: https://www.publish-teacher-training-courses.service.gov.uk
```
defong: @fofr is this one still correct then?

benilovj: I think this one's fine – basically we are using the service domain for Publish.

PR #276, opened 2019-05-14
[1485] Reduce cookie size

heidar:
### Context
Our cookies were getting too big. With all the auth info from DSI (DfE Sign In) + encryption + some extra stuff, it overflows.

We only use the DSI ID, email, first and last name, and the ID token, the rest of the data we get from DSI we don't use

Only save the data we need from DSI, throw the rest away. This makes the cookie much smaller.

### Guidance to review

I want to refactor this badly, but I'm holding off so I can finish the factories. I'm hoping we can replace much of this code with a shared DSI gem.

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -6,7 +6,17 @@ class SessionsController < ApplicationController
   end
 
   def create
-    session[:auth_user] = auth_hash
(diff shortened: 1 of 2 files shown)


heidar:
[Reduce cookie size](https://trello.com/c/gbH4hFl8/1485-reduce-cookie-size)

PR #287, opened 2019-05-19
[1500] Move copy content warning into partial

dankmitchell:
### Context
Cleaner code

Move copy content warning into partial to avoid duplicated code

### Guidance to review
No visual or functional change, just DRYer code.


dankmitchell:
[DRYup Refactor copy_form warning message](https://trello.com/c/EvyOkZO1/1500-dryup-refactor-copyform-warning-message)

benilovj:
The template DRYing is definitely a good improvement, but now `_copy_content_warning.html.erb` has some logic belonging to the individual pages.

How about using [`content_for`](https://guides.rubyonrails.org/layouts_and_rendering.html#using-the-content-for-method) to keep the page-specific bits (probably the list items) in the original template files?

dankmitchell:
Nice idea @benilovj. I deffo don't like that logic but it was an improvement on the previous code. Another iteration to come :)

PR #291, opened 2019-05-21
Implement layer caching

smneal:
### Context
Reduce build time by implementing layer caching

Add task to push master builds as latest tags
Add task to pull latest tag for each build


smneal:
> Thinking about this a bit more, what I've implemented in the past is tagging images with `$(IMAGE_NAME):$(BRANCH_NAME)` and then using that for the image cache. Might be worth considering doing in the future.

Thanks Misha. That could be something to look at if you've used it in the past and it's worked well. For short lived branches how did you keep the registry tidy?

misaka:
> For short lived branches how did you keep the registry tidy?

That's a good question. I wonder if we could integrate with an on-delete hook on GitHub.

PR #296, opened 2019-05-22
1483 add course preview

dankmitchell:
### Context
One step closer to deprecating manage courses ui

Add `courses#preview`

### Guidance to review
# After
localhost_3000_organisations_1CS_courses_2DHM_preview(Laptop with MDPI screen)

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,6 +1,17 @@
 module ApplicationHelper
   def markdown(source)
-    markdown = RDiscount.new(source)
-    markdown.to_html.html_safe
+    options = {
+      filter_html:     true,
+      link_attributes: { rel: 'nofollow', target: "_blank", class: "govuk-link" },
+    }
+
+    extensions = {
+      lax_spacing: true
+    }
+
+    renderer = Redcarpet::Render::HTML.new(options)
+    markdown = Redcarpet::Markdown.new(renderer, extensions)
+
+    markdown.render(source).html_safe
   end
 end
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -1,0 +1,6 @@
+module ApplicationHelper
+  def markdown(source)
+    markdown = RDiscount.new(source)
+    markdown.to_html.html_safe
+  end
+end
--- a/app/views/courses/preview.html.erb
+++ b/app/views/courses/preview.html.erb
@@ -1,0 +1,94 @@
+<%= content_for :page_title, "#{course.name_and_code} - Courses" %>
+
+<% content_for :before_content do %>
+  <%= link_to "Back to course", provider_course_path, class: "govuk-back-link" %>
+
+  <p class="govuk-body preview-banner">
+    This is a preview of a course: <%= course.name_and_code %>.
+  </p>
+<% end %>
+
+<h1 class="govuk-heading-xl">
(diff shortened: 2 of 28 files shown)


dankmitchell:
[[M] Add course previewing](https://trello.com/c/srALqCie/1483-m-add-course-previewing)

app/views/courses/preview/_apply.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="govuk-!-margin-bottom-8">
+  <h2 class="govuk-heading-l" id="section-apply">Apply</h2>
+  <% if course.has_vacancies? %>
+    <p class="govuk-body">
+      <a href="https://2019.teachertraining.apply.ucas.com/apply/student/login.do">Apply on the UCAS website</a>. You’ll need to register with UCAS before you can apply.
```
benilovj: Could this be in Settings? It will change between recruitment cycles.

app/views/courses/preview/_qualification.html.erb
```diff
@@ -0,0 +1,70 @@
+<% if course.outcome == 'QTS' %>
+  <details class="govuk-details">
```
benilovj: I wonder if this template can be DRYed up with draper?

app/helpers/application_helper.rb
```diff
@@ -0,0 +1,6 @@
+module ApplicationHelper
+  def markdown(source)
+    markdown = RDiscount.new(source)
```
benilovj: Can we lock Markdown down to just the features we are using (lists and links)?

dankmitchell: 100%

app/views/courses/preview.html.erb
```diff
@@ -0,0 +1,94 @@
...
+  </dd>
+  <% if course.length.present? %>
+    <dt class="govuk-list--description__label">Course length</dt>
+    <dd><%= course.length %></dd>
+  <% end %>
+  <% if course.applications_open_from.present? %>
+    <dt class="govuk-list--description__label">Date you can apply from</dt>
+    <dd><%= l(course.applications_open_from&.to_date) %></dd>
+  <% end %>
+  <% if course.applications_open_from.present? %>
```
misaka: Should this and the section above be separate? They're testing the same thing ...

app/views/courses/preview/_apply.html.erb
misaka: I think this should probably be a call to a decorator or helper method which should return the right url based on the recruitment cycle that the course is in. Settings would be too static for this.

app/views/courses/preview/_fees.html.erb
```diff
@@ -0,0 +1,35 @@
+<div class="govuk-!-margin-bottom-8">
+  <h2 class="govuk-heading-l" id="section-fees">Fees</h2>
+
+  <% if course.fee_uk_eu.present? %>
+    <div class="body-text">
+      <table class="govuk-table">
+        <caption class="govuk-table__caption govuk-!-font-weight-regular govuk-!-margin-bottom-4">The course fees for 2019/2020 are as follows:</caption>
```
misaka: Similar to the above comment about the apply URL, this could benefit from being a helper/decorator method that will, later, sort out the correct text to insert for the applicable recruitment cycle for this course.

PR #299, opened 2019-05-23
1483 course preview about course

dankmitchell:
### Context
Course preview

- Setup course preview contents nav
- Add about course content
- Add markdown and a nice parser from [GDS RE team](https://github.com/alphagov/re-rotas/blob/master/app/lib/rotas/markdown_renderer.rb)

### Guidance to review
Example - `/organisations/2AT/courses/3CXF/preview`

# After
localhost_3000_organisations_2AT_courses_3CXF_preview(Laptop with MDPI screen)

### DIFF
--- a/spec/lib/govuk/markdown_renderer_spec.rb
+++ b/spec/lib/govuk/markdown_renderer_spec.rb
@@ -1,0 +1,180 @@
+require 'rails_helper'
+
+# Ignore whitespace
+def expect_equal_ignoring_ws(first, second)
+  expect(first.lines.map(&:strip).join('')).to eq(second.lines.map(&:strip).join(''))
+end
+
+describe Govuk::MarkdownRenderer do
+  let(:html) { Govuk::MarkdownRenderer.render(markdown) }
+
+  describe 'ul' do
+    let(:markdown) do
+      <<~MD
+        - item
+        - item
+        - item
+      MD
+    end
+
+    it 'renders correct HTML' do
+      expected_html = <<~HTML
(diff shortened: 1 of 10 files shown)


spec/lib/govuk/markdown_renderer_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+
+# Ignore whitespace
+def expect_equal_ignoring_ws(first, second)
+  expect(first.lines.map(&:strip).join('')).to eq(second.lines.map(&:strip).join(''))
+end
+
+describe Govuk::MarkdownRenderer do
+  let(:html) { Govuk::MarkdownRenderer.render(markdown) }
+
+  context 'ul' do
```
misaka: Could we make these `describe` please? As-in, this is the subject of the test.

spec/lib/govuk/markdown_renderer_spec.rb
```diff
@@ -0,0 +1,180 @@
...
+    let(:markdown) do
+      <<~MD
+        - item
+        - item
+        - item
+      MD
```
misaka: The specific example of a list we saw today was a list that followed a paragraph with no line-break. Can we test that we render that correctly here too?

fofr: When I spoke to @dankmitchell he said he would look at this in a follow-up PR.

spec/lib/govuk/markdown_renderer_spec.rb
dankmitchell: Done

PR #318, opened 2019-05-30
1438 improve course preview page

dankmitchell:
### Context
Course preview page

Minor improvements based on current implementation

### Guidance to review
| old  | new  |
|---|---|
| bat-dev-manage-courses-ui-app azurewebsites net_organisation_13e_course_self_3bqq_preview(Laptop with MDPI screen) | localhost_3000_organisations_13e_courses_3bqq_preview(Laptop with MDPI screen) |

### DIFF
--- a/app/lib/govuk/markdown_renderer.rb
+++ b/app/lib/govuk/markdown_renderer.rb
@@ -26,11 +26,7 @@ module Govuk
     end
 
     def link(link, title, content)
-      <<~HTML
-        <a href="#{link}" title="#{title}" class="govuk-link">
-          #{content}
-        </a>
-      HTML
+      %(<a href="#{link}" class="govuk-link">#{content}</a>)
     end
 
     def autolink(link, link_type)
--- a/app/lib/govuk/markdown_renderer.rb
+++ b/app/lib/govuk/markdown_renderer.rb
@@ -33,6 +33,10 @@ module Govuk
       HTML
     end
 
+    def autolink(link, link_type)
+      %(<a href="#{link}" class="govuk-link">#{link}</a>)
+    end
+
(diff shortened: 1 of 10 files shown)


app/lib/govuk/markdown_renderer.rb
```diff
@@ -33,6 +33,10 @@ def link(link, title, content)
       HTML
     end
 
+    def autolink(link, link_type)
+      %(<a href="#{link}" class="govuk-link">#{link}</a>)
```
fofr: Do we need to worry about mailto links here?

app/webpacker/stylesheets/application.scss
```diff
@@ -12,6 +12,10 @@ $button-colour: #00823b;
 @import "warning-summary";
 @import "definition-list";
 
+.govuk-list li .govuk-body:last-child {
```
fofr:
```suggestion
.govuk-list li > .govuk-body:last-child {
```

Should we only apply this to paragraphs that are direct descendants of an li element?

app/lib/govuk/markdown_renderer.rb
```diff
@@ -25,14 +25,18 @@ def list(content, list_type)
       end
     end
 
-    def link(link, title, content)
+    def link(link, _title, content)
       <<~HTML
-        <a href="#{link}" title="#{title}" class="govuk-link">
+        <a href="#{link}" class="govuk-link">
```
fofr: I think the line break here is causing an extra space to appear in the markup:

This markdown:
Screen Shot 2019-05-30 at 16 31 04

Comes with a space after it:
Screen Shot 2019-05-30 at 16 30 45

PR #321, opened 2019-05-30
[1513] Hook up edit locations form

tvararu:
### Context

This allows users to add locations to a course.

Creates the `courses/sites#update` action and wires up the form to `PUT` to it. This assigns the selected sites to the payload and then `saves` it.

### Guidance to review

Validation errors:

Screen Shot 2019-05-31 at 14 27 05

Success:

Screen Shot 2019-05-31 at 14 12 09

### DIFF
--- a/app/controllers/courses/sites_controller.rb
+++ b/app/controllers/courses/sites_controller.rb
@@ -34,7 +34,7 @@ module Courses
     def build_course
       @provider_code = params[:provider_code]
       @course = Course
-        .includes(site_statuses: [:site])
+        .includes(:sites)
         .includes(provider: [:sites])
         .where(provider_code: @provider_code)
         .find(params[:code])
--- a/app/controllers/courses/sites_controller.rb
+++ b/app/controllers/courses/sites_controller.rb
@@ -8,6 +8,27 @@ module Courses
 
     def edit; end
 
+    def update
+      @course.provider_code = @provider.provider_code
+      selected_site_ids = params.dig(:course, :site_statuses_attributes)
+        .values
+        .select { |f| f["selected"] == "1" }
+        .map { |f| f["id"].to_i }
+
+      @course.sites = @provider.sites.select { |site| selected_site_ids.include?(site.id) }
+
+      if @course.save
+        @course.sync_with_search_and_compare(provider_code: params[:provider_code])
+
+        redirect_to provider_course_path(params[:provider_code], params[:code]), flash: { success: 'Course locations saved' }
+      else
+        # TODO: Change this to be @course.errors when the API is updated.
+        flash[:error] = "You must choose at least one location"
+
+        render :edit
+      end
+    end
+
   private
 
     def build_course
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -78,7 +78,7 @@ private
   def build_course
     @provider_code = params[:provider_code]
     @course = Course
-      .includes(site_statuses: [:site])
+      .includes(:sites)
       .includes(provider: [:sites])
       .includes(:accrediting_provider)
       .where(provider_code: @provider_code)
@@ -128,7 +128,7 @@ private
   end
 
   def build_copy_course
-    @source_course = Course.includes(site_statuses: [:site])
+    @source_course = Course.includes(:sites)
                            .includes(provider: [:sites])
                            .includes(:accrediting_provider)
                            .where(provider_code: @provider_code)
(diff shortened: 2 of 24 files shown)


tvararu:
[[M] Course basic details - Edit locations](https://trello.com/c/qGZ1yDBA/1513-m-course-basic-details-edit-locations)

app/controllers/courses/sites_controller.rb
```diff
@@ -8,6 +8,28 @@ class SitesController < ApplicationController
 
     def edit; end
 
+    def update
+      @course.provider_code = @provider.provider_code
+      @course.sites = params.dig(:course, :site_statuses_attributes)
+        &.values
+        &.select { |field| field["selected"] == "1" }
+        &.map { |field| @provider.sites.find { |site| site.id == field["id"] } }
```
tvararu: Paging @misaka for a way to make this more 整頓

app/controllers/courses/sites_controller.rb
```diff
@@ -8,6 +8,24 @@ class SitesController < ApplicationController
...
+
+      if @course.save
+        @course.sync_with_search_and_compare(provider_code: params[:provider_code])
+
+        redirect_to provider_course_path(params[:provider_code], params[:code]), flash: { success: 'Course locations published' }
```
fofr: We should only say Published if the course is published, otherwise we should say Saved

app/controllers/courses/sites_controller.rb
```diff
@@ -8,6 +8,30 @@ class SitesController < ApplicationController
...
+      @course.provider_code = @provider.provider_code
+      @course.sites = params.dig(:course, :site_statuses_attributes)
+        &.values
+        &.select { |field| field["selected"] == "1" }
+        &.map { |field| @provider.sites.find { |site| site.id == field["id"] } }
+
+      # If I call @course.save directly, and the API returns an error,
+      # it seems to mutate the @course object, removing `site_statuses`
+      # and causing `render :edit` to fail.
+      no_idea_why_i_have_to_dup_this_course = @course.dup
```
tvararu: I don't want to merge this, obviously, but I can't figure out why I can't call `@course.save` directly...

tvararu:
Marking this as good for review; suggestions welcome on how to address the `.dup` hack...

app/controllers/courses/sites_controller.rb
heidar: This is probably because the API doesn't allow attributes it isn't expecting?

app/controllers/courses/sites_controller.rb
misaka: Ya, not sure. For one thing, I'm not sure about the safeties ... do you really want to assign `@course.sites` to nil? Aside from that, it might be nice to use the `where` with a list of ids.

```ruby
      selected_site_ids = params.dig(:course, :site_statuses_attributes).values
        .select { |f| f['selected'] == '1' }
        .map    { |f| f['id'].to_i }
      @course.sites = @provider.sites.where(id: selected_site_ids)
```

app/controllers/courses/sites_controller.rb
misaka: It sounds like the question is why is `@course` changing at all when an error is returned.

Shot in the dark, but could this be related to the through relationship? I can see how it might be a little be confused if you modify `sites` here and then when it tries to save it to the backend it gets an error ... then is `site_statuses` still valid since you've just tried to save a change to it and it's come back with an error? I'd have to play around with it to understand it better.

misaka: Is this related to the work you were doing recently to remove the `site_statuses` from the response sent by the backend?

app/controllers/courses/sites_controller.rb
tvararu: No `.where` because `@provider.sites` is a plain array, not an AR model. Have improved it with the rest

app/controllers/courses/sites_controller.rb
```diff
@@ -8,12 +8,36 @@ class SitesController < ApplicationController
 
     def edit; end
 
+    def update
+      @course.provider_code = @provider.provider_code
```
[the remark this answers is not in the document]
tvararu: Separate PR?

tvararu:
Reviewing commit by commit is easiest.

app/controllers/courses/sites_controller.rb
tvararu: Sorted it out, seems to work without this hack now.

PR #332, opened 2019-06-04
[1505] Refactor copy content warning

fofr:
Move copied fields out of partial and into parent calling the partial to better split the responsibility between the templates.

_copy_content_warning.html.erb now doesn't have logic that's specific to the about/fees/req's/salary forms.

### DIFF
--- a/app/views/courses/_copy_content_warning.html.erb
+++ b/app/views/courses/_copy_content_warning.html.erb
@@ -7,24 +7,8 @@
       We’ve copied these fields from <%= @source_course.name %> (<%= @source_course.course_code %>):
     </p>
     <ul class="govuk-list">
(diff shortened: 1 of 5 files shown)


fofr:
[[XS] Follow up to copy_form warning message refactor](https://trello.com/c/FQTNtNiX/1505-xs-follow-up-to-copyform-warning-message-refactor)

fofr:
While doing this I realised that this implementation is actually a regression:
https://trello.com/c/BMyfmPrn/1578-rails-copy-content-messaging-is-a-regression-on-live

We should be hardcoding which fields were copied, it should be dependent on the actual content being copied.

fofr:
Merging this anyway for now.

PR #334, opened 2019-06-04
[1577] Fix error on enrichment pages when there are no self accredited courses

fofr:
### Context

We were seeing an error when accessing the enrichment pages for a course offered by a school direct provider (ie one with no self-accredited courses):

* /requirements
* /about
* /salary
* /fees

```
undefined method `reject' for nil:NilClass
```

Example path:
`/organisations/1XO/courses/2FPL/salary`

### Changes

- Added a failing test
- Fixed with a safe navigation operator

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -124,7 +124,7 @@ private
 
   def filter_courses
     @courses_by_accrediting_provider = @courses_by_accrediting_provider.reject { |c| c == course.id }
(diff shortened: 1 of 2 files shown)


fofr:
[Error on course enrichment pages for providers without self accredited courses](https://trello.com/c/IDZUA3Lh/1577-error-on-course-enrichment-pages-for-providers-without-self-accredited-courses)

PR #335, opened 2019-06-04
[1578] Fix copy content summary message

dankmitchell:
### Context
Follow up to https://github.com/DFE-Digital/manage-courses-frontend/pull/332

- Send "course name" and "course code" into copy warning summary partial
- Only render the summary box when there is actually content being copied
- Only render deep links for content that's been copied

### Guidance to review
Example - `https://localhost:3000/organisations/2AT/courses/35L7/about?utf8=%E2%9C%93&copy_from=35L6` Message in the summary box should have course code `35L6`

### DIFF
--- a/app/views/courses/about.html.erb
+++ b/app/views/courses/about.html.erb
@@ -7,9 +7,9 @@
 <%= render partial: 'courses/copy_content_warning',
                     locals: {
                       copied_fields: [
-                        ['About the course', '#about_course_wrapper'],
-                        ['Interview process', '#interview_process_wrapper'],
-                        ['How school placements work', '#how_school_placements_work_wrapper']
+                        ['About the course', 'about_course'],
+                        ['Interview process', 'interview_process'],
+                        ['How school placements work', 'how_school_placements_work']
                       ]} if params[:copy_from].present? %>
 
 <h1 class="govuk-heading-xl">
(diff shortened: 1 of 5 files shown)


dankmitchell:
[Rails Copy content messaging is a regression on Live](https://trello.com/c/BMyfmPrn/1578-rails-copy-content-messaging-is-a-regression-on-live)

fofr:
This fixes a different bug to https://trello.com/c/BMyfmPrn/1578-rails-copy-content-messaging-is-a-regression-on-live.

app/views/courses/about.html.erb
```diff
@@ -4,15 +4,34 @@
...
-                    locals: {
-                      source_course_name: @source_course.name,
-                      source_course_code: @source_course.course_code,
-                      copied_fields: [
-                        ['About the course', '#about_course_wrapper'],
-                        ['Interview process', '#interview_process_wrapper'],
-                        ['How school placements work', '#how_school_placements_work_wrapper']
-                      ]} if params[:copy_from].present? %>
+<% if params[:copy_from].present? %>
+  <% if @source_course.about_course.present? || @source_course.interview_process.present? || @source_course.how_school_placements_work.present? %>
```
misaka: Just trying to understand the issue being addressed in this change ... is this the key here? That it uses `@source_course.about_course.present?` to decide whether to display this message?

If so, then we should still be able to use the partial and perform the check there. We could pass in `'about_course'`, instead of `'#about_course_wrapper'` for `copied_fields`, and then in the partial:

```
    <% if copied_fields.any? { |_, attr| @source_course.__send__(attr).present? } %>
```

to test whether to print any messages, and use `"##{field[1]}_wrapper"` when passing it to `link_to`.

dankmitchell: Nice suggestion. Thanks

dankmitchell: All done. Thanks @misaka

app/views/courses/about.html.erb
```diff
@@ -6,10 +6,12 @@
 
 <%= render partial: 'courses/copy_content_warning',
                     locals: {
+                      source_course_name: @source_course.name,
```
defong: dont think these should be here

PR #338, opened 2019-06-05
[1513] Don't .to_i selected_site_ids in course locations update

tvararu:
### Context

Updating locations on courses is on dev but broken.

I switched from using `.to_i` in both places to only doing it on the `selected_site_ids` in a previous refactor, and did not notice it broke functionality. This adds more tests to make sure we send the right payload to the API.

### DIFF
--- a/app/controllers/courses/sites_controller.rb
+++ b/app/controllers/courses/sites_controller.rb
@@ -13,7 +13,7 @@ module Courses
       selected_site_ids = params.dig(:course, :site_statuses_attributes)
         .values
         .select { |f| f["selected"] == "1" }
-        .map { |f| f["id"].to_i }
+        .map { |f| f["id"] }
 
       @course.sites = @provider.sites.select { |site| selected_site_ids.include?(site.id) }
 
--- a/spec/features/courses/sites/edit_spec.rb
+++ b/spec/features/courses/sites/edit_spec.rb
@@ -55,7 +55,21 @@ feature 'Edit course sites', type: :feature do
       stub_api_v2_request(
         "/providers/#{provider.provider_code}/courses/#{course.course_code}",
         {}, :patch, 200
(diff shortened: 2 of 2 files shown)


tvararu:
[[M] Course basic details - Edit locations](https://trello.com/c/qGZ1yDBA/1513-m-course-basic-details-edit-locations)

app/controllers/courses/sites_controller.rb
```diff
@@ -13,7 +13,7 @@ def update
       selected_site_ids = params.dig(:course, :site_statuses_attributes)
         .values
         .select { |f| f["selected"] == "1" }
-        .map { |f| f["id"].to_i }
+        .map { |f| f["id"] }
```
misaka: Ok, now I'm confused. You _have_ to leave them as strings? So this means that object ids are being returned as strings for these JSONAPI resources?

tvararu: Y E S

misaka: I just had to see for myself.

```
I, [2019-06-05T16:48:43.960678 #75522]  INFO -- omniauth: (dfe) Request phase initiated.
I, [2019-06-05T16:49:07.534751 #75522]  INFO -- omniauth: (dfe) Callback phase initiated.

From: /Users/misaka/Code/DfE/manage-courses-frontend/app/controllers/providers_controller.rb @ line 13 ProvidersController#show:

    10: def show
    11:   @provider = Provider.find(params[:code]).first
    12:   binding.pry
 => 13: end

[1] pry(#<ProvidersController>)> @provider.id
=> "12035"
```

Ok, not what I expected.

misaka: I wonder if this is something we should be deserialising by hand.

PR #341, opened 2019-06-06
[1588] Indicate if locations changes will be published

fofr:
If a course is running - ie it's on Find – then changes to the locations will go live immediately. We previously based this on the enrichment status.


fofr:
[Button and success text when editing course locations based on wrong status](https://trello.com/c/9NNvpY7Q/1588-button-and-success-text-when-editing-course-locations-based-on-wrong-status)

PR #344, opened 2019-06-07
[1584] When copying course fields, don't blank a field if the source was blank

fofr:
### Context

Previously, in the live rails version, if a field was empty in the course being copied, it would cause that field to be emptied in the new course.

In the C# version if the field was empty in the course being copied, then that field would be left alone.

The C# version was correct.

* Move the logic for checking if a value is present out of view and into controller
* Only set values on the destination course if present in the source course
* Add a test to illustrate

### Guidance to review

* Test by copying from courses where only some data is available, but into courses that are otherwise complete (eg add something to an empty course and copy that to a published one).

* Seems like the copy course features are in the `spec/requests` when I'd expect to find them using SitePrism in `spec/features`. I've avoided adding further specs because I think this should be refactored after this PR.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -20,34 +20,42 @@ class CoursesController < ApplicationController
 
   def about
     if params[:copy_from].present?
-      course.about_course = @source_course.about_course
-      course.interview_process = @source_course.interview_process
-      course.how_school_placements_work = @source_course.how_school_placements_work
+      @copied_fields = [
+        ['About the course', 'about_course'],
+        ['Interview process', 'interview_process'],
+        ['How school placements work', 'how_school_placements_work']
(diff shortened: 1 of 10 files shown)


fofr:
[Copying from one course to another shouldn't blank fields if they were blank in the copied course](https://trello.com/c/gHEE1xCo/1584-copying-from-one-course-to-another-shouldnt-blank-fields-if-they-were-blank-in-the-copied-course)

app/controllers/courses_controller.rb
```diff
@@ -135,4 +143,9 @@ def build_copy_course
                            .find(params[:copy_from])
                            .first
   end
+
+  def copy_field_if_present_in_source_course(field)
```
benilovj: alternative (slightly more concise) way of doing this, I think:

```ruby
def copy_field_if_present_in_source_course(field)
  course[field] = @source_course[field] || course[field]
end
```

PR #345, opened 2019-06-07
[1580] Update courses description and locations edit titles

tvararu:
They should contain "Error:" in case of validation errors, for better accessibility.

### DIFF
--- a/spec/features/courses/description_spec.rb
+++ b/spec/features/courses/description_spec.rb
@@ -184,6 +184,7 @@ feature 'Course description', type: :feature do
           scenario 'it shows the description page with validation errors' do
             course_page.publish.click
 
+            expect(page.title).to have_content('Error:')
             expect(course_page).to be_displayed
             expect(course_page.error_summary).to have_content("About course can't be blank")
           end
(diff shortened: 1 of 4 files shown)


tvararu:
[[M] Course basic details - Edit locations validations](https://trello.com/c/QkdCYsJL/1580-m-course-basic-details-edit-locations-validations)

spec/features/courses/sites/edit_spec.rb
```diff
@@ -113,6 +113,7 @@
 
       locations_page.save.click
 
+      expect(page.title).to have_content('Error:')
```
benilovj:
```suggestion
      expect(course_page.title).to have_content('Error:')
```

spec/features/courses/description_spec.rb
```diff
@@ -184,6 +184,7 @@
           scenario 'it shows the description page with validation errors' do
             course_page.publish.click
 
+            expect(page.title).to have_content('Error:')
```
benilovj:
```suggestion
            expect(course_page.title).to have_content('Error:')
```

spec/features/courses/sites/edit_spec.rb
tvararu: The `.title` is already being used for the `h1` on the page.

I don't think it's wrong in this situation to use the `page.title` which already selects the correct element vs having to update all our page objects with a new `title` selector that is actually the `css('head title')`

benilovj: oh fair enough

PR #347, opened 2019-06-07
Add new_staging rails env

smneal:
### Context
Add rails env and settings for new staging environment

Addition of env, setting files and update to webpacker


config/settings/new_staging.yml
```diff
@@ -0,0 +1,11 @@
+dfe_signin:
+  issuer: https://signin-test-oidc-as.azurewebsites.net
+  profile: https://signin-test-pfl-as.azurewebsites.net
+  secret: please_change_me # Override with SETTINGS__DFE_SIGNIN__SECRET
```
defong: remember to change it

smneal: Thanks. It's already set in the new staging env :) 👍

PR #349, opened 2019-06-07
[1480] Wire up course length and fees form to update course

heidar:
### Context

This is a precursor PR to this whole story. Although it does implement most of the functionality, it's still missing a couple of things that will be done separately.

This will unblock similar work that needs to be done on a few other forms.

The fee form was wired up to post to the publish endpoint. We added an
update action to the course controller and wired up the fees form to use
that instead.

Paired with Sau and Tim.

### Guidance to review

Validations are blocked by the backend implementing validation on update. Right now they only fire on publish.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -10,6 +10,20 @@ class CoursesController < ApplicationController
 
   def show; end
 
+  def update
+    @course.provider_code = @provider.provider_code
+
+    if @course.save
+      flash[:success] = 'Your changes have been saved'
+      redirect_to(
+        description_provider_course_path(
+          @provider.provider_code,
+          @course.course_code
+        )
+      )
+    end
+  end
+
   def description
     @published = flash[:success]
     flash.delete(:success)
@@ -75,7 +89,7 @@ class CoursesController < ApplicationController
         ]
       }.to_h
     else
-      flash[:success] = 'Your changes have been published'
+      flash[:success] = "Your course has been published. The link for this course is: #{Settings.search_ui.base_url}/course/#{@provider.provider_code}/#{@course.course_code}"
     end
 
     redirect_to description_provider_course_path(@provider.provider_code, @course.course_code)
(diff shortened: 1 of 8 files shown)


app/controllers/courses_controller.rb
```diff
@@ -75,7 +90,7 @@ def publish
         ]
       }.to_h
     else
-      flash[:success] = 'Your changes have been published'
+      flash[:success] = "Your course has been published. The link for this course is: https://localhost:5000/course/#{@provider.provider_code}/#{@course.course_code}"
```
benilovj: You’ll want to parameterise the URL with the right env.

dankmitchell:
```suggestion
      flash[:success] = "Your course has been published. The link for this course is: #{Settings.search_ui.base_url}/course/#{@provider.provider_code}/#{@course.course_code}"
```

fofr: This flash message ought to have "Your course has been published" as the title, and the following sentence as a paragraph. But this can come in a follow up PR as it'll involve refactoring our flash messages somewhat.

app/views/courses/description.html.erb
```diff
@@ -42,7 +42,7 @@
 <% if @published.present? %>
   <div class="govuk-success-summary" aria-labelledby="success-summary-title" role="alert" tabindex="-1" data-module="error-summary">
     <h2 class="govuk-success-summary__title" id="success-summary-title">
-      Your course has been published
+      <%= @published %>
```
fofr: I wonder if this should be `@success_message` as it is sometimes a save message and sometimes a publish one.

app/views/courses/fees.html.erb
```diff
@@ -14,7 +14,7 @@
 
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
-    <%= form_for course, url: publish_provider_course_path(course.provider.provider_code, course.course_code), method: :post do |f| %>
+    <%= form_for course, url: provider_course_path(course.provider.provider_code, course.course_code) do |f| %>
```
dankmitchell: If we making this change should we change `form_for` to `form_with`?

https://m.patrikonrails.com/rails-5-1s-form-with-vs-old-form-helpers-3a5f72a8c78a

app/views/courses/description.html.erb
heidar: I want to do another PR to refactor this to just use the flash

PR #352, opened 2019-06-10
[1608] Don't sync when editing locations or vacancies

tvararu:
### Context

The frontend doesn't necessarily know when a course is due to be synced with search-and-compare, the concern for that knowledge should be in the backend.


tvararu:
[Don't sync courses when editing location or vacancies.](https://trello.com/c/71q1Bfh7/1608-dont-sync-courses-when-editing-location-or-vacancies)

misaka:
@timabell has merged the PR that rolls back his breaking change, fwiw, so dev is a clean slate to test with.

PR #355, opened 2019-06-10
[1610] Add provider code to course factory

heidar:
### Context

Due to the way we use the factories, it's difficult to set the
provider_code on the course without causing an endless loop.

Also, setting provider_code on the course factory gets overwritten by
the provider being set to nil in the factory, and removing that breaks
things.

This is further broken by a bug in the JSONAPI client.

Added workaround for broken factories/jsonapi client.
Removed workaround for `provider_code` not being in responses.

### Guidance to review

I cleaned up the test affected, some of that is unrelated to these changes, sorry.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -11,8 +11,6 @@ class CoursesController < ApplicationController
   def show; end
 
   def update
-    @course.provider_code = @provider.provider_code
-
     if @course.save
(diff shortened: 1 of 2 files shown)


heidar:
[Expose provider code in courses endpoint](https://trello.com/c/VKYT3Xbo/1610-expose-provider-code-in-courses-endpoint)

PR #359, opened 2019-06-11
[1608] Sync courses when updating vacancies

tvararu:
### Context

We removed this because we were planning on implementing it on the backend, but we're thinking about including it back in temporarily.


tvararu:
[Don't sync courses when editing location or vacancies.](https://trello.com/c/71q1Bfh7/1608-dont-sync-courses-when-editing-location-or-vacancies)

PR #361, opened 2019-06-11
[xxxx] Add feature specs for copying each type of enrichment

fofr:
### Context

We had some request specs but no end to end tests for copying enrichments.

* Add interactions for selecting a course and copying it
* Checks the warning message shown
* Checks the correct values were copied
* Fixes existing checks by looking for `.value` rather than `.have_content` on an `input`

Follow up to #344

### DIFF
--- a/spec/features/courses/about_spec.rb
+++ b/spec/features/courses/about_spec.rb
@@ -73,6 +73,83 @@ feature 'About course', type: :feature do
     end
   end
 
+  context 'when copying course requirements from another course' do
+    let(:course_2) {
+      jsonapi(
+        :course,
+        name: 'Biology',
+        provider: provider,
+        about_course: 'Course 2 - About course',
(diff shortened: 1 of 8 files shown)


spec/features/courses/about_spec.rb
```diff
@@ -81,4 +72,88 @@
...
+        expect(about_course_page.warning_message).not_to have_content(name)
+      end
+
+      expect(about_course_page.about_textarea.value).to eq(course_3.about_course)
+      expect(about_course_page.interview_process_textarea.value).to eq(course_2.interview_process)
+      expect(about_course_page.how_school_placements_work_textarea.value).to eq(course_2.how_school_placements_work)
+    end
+  end
+
+  def copy_fees(from:, to:)
```
benilovj: It'd be lovely to have this functionality on the SitePrism page object, rather than in the spec. Is that tricky to do?

PR #366, opened 2019-06-12
[1681] Fix enrichment updates

dankmitchell:
### Context
```
Follow-up story to #1480, #1478, #1482, #1481. The forms were all wired up correctly but we're not actually passing the data over to the backend so it's failing to save since the params are `nil`
```

Pass course params over to the backend

### Guidance to review
Make a change to a course enrichment and hit save


dankmitchell:
[Updating enrichments does not save](https://trello.com/c/uPSKNHTV/1618-updating-enrichments-does-not-save)

PR #368, opened 2019-06-13
[1626] Fix for course enrichments saving no changes

dankmitchell:
### Context
Course enrichments

Force all changes to be `dirty` so they are sent to the backend regardless if a user changes something or not

### Guidance to review
🤷‍♂

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -11,6 +11,7 @@ class CoursesController < ApplicationController
   def show; end
 
   def update
+    @course.set_all_attributes_dirty
     if @course.update(course_params)
       flash[:success] = 'Your changes have been saved'
       redirect_to(


dankmitchell:
[Cannot update a course enrichment is nothing changes](https://trello.com/c/0J2v5VmW/1626-cannot-update-a-course-enrichment-is-nothing-changes)

app/controllers/courses_controller.rb
```diff
@@ -11,6 +11,7 @@ def index; end
   def show; end
 
   def update
+    @course.set_all_attributes_dirty
```
misaka: Just to ask, why not just suppress the update completely if there are no changes?

misaka: @heidar and I just did a spike of a solution in the backend to allow for empty attributes using `fetch`. We think this would be a cleaner approach than this ... I'll cut a PR for it.

PR #373, opened 2019-06-14
[1623] Update uk and international fee data type

dankmitchell:
### Context
UK and International Fees

Specifical set UK and International fees to integer

### Guidance to review
Post should not be a string

### DIFF
--- a/spec/requests/courses/fees_spec.rb
+++ b/spec/requests/courses/fees_spec.rb
@@ -14,8 +14,8 @@ describe 'Courses', type: :request do
               include_nulls: [:accrediting_provider],
               course_length: 'TwoYears',
               provider: provider,
-              fee_uk_eu: '9500',
-              fee_international: '1200',
+              fee_uk_eu: 9500,
+              fee_international: 1200,
               fee_details: 'Some information about the fees',
               financial_support: 'Some information about the finance support'
     end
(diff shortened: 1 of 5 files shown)


dankmitchell:
I don't think we actually need this change 🤔

benilovj:
It seems like this change will improve our model of the underlying data?

tvararu:
@dankmitchell is it because we have frontend validations on the input? We should disable those I think.

dankmitchell:
[Add missing validations to enrichments](https://trello.com/c/e7wGJYfZ/1623-add-missing-validations-to-enrichments)

spec/requests/courses/fees_spec.rb
```diff
@@ -105,8 +105,8 @@
 
         original_course_details = [
                                     course_2.course_length,
-                                    course_2.fee_uk_eu,
-                                    course_2.fee_international
+                                    course_2.fee_uk_eu.to_s,
```
defong: you sure you need to to string it?

dankmitchell: JSON API Client 🤷‍♂

fofr:
I think this has introduced this bug:
https://trello.com/c/weZDU1qy/1671-saving-an-empty-fee-gets-converted-to-0

PR #375, opened 2019-06-14
[1633] Fail the build when tests fail

heidar:
### Context

Pipeline says a build is successful even when tests fail.

Make the build fail when tests fail.

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -106,3 +106,4 @@ steps:
   inputs:
     testRunner: JUnit
     testResultsFiles: '*.xml'
+    failedTaskOnFailedTest: true


heidar:
[Fail pipeline build when tests fail](https://trello.com/c/uc12PoYH/1633-fail-pipeline-build-when-tests-fail)

heidar:
> What was it doing before? Not failing?

lol yeah...

We could have had it always fail with the old stderr hack, always pass (without it), and with this actually fail when tests fail. Derp.

heidar:
> Still, wat? y?

I have no idea, Azure is weird, weird weird weird

misaka:
> We could have had it always fail with the old stderr hack, always pass (without it), and with this actually fail when tests fail. Derp.

🤦‍♂

heidar:
Insane default ftw!

PR #379, opened 2019-06-18
[1624] Don't strip latest_enrichment__ from enrichment errors

tvararu:
### Context

This is being removed from the API, as it causes pointers to not be associated correctly.

Depends on:

- https://github.com/DFE-Digital/manage-courses-backend/pull/497

Remove the `latest_enrichment__` prefix from course enrichment errors.


tvararu:

PR #382, opened 2019-06-18
[1624] Simplify site validation errors

tvararu:
### Context

These are now being sent from the API with correct formatting and can be directly displayed in the frontend.

Remove `map_errors`.


tvararu:

PR #383, opened 2019-06-18
[1629] Fix up course length

dankmitchell:
### Context
Course length wasn't being posted to the backend properly so it the data wasn't being saved

Re-factor form to handle 1year, 2years and "Other"

### Guidance to review
No visual change.

Note: We need to address the backend to save the "other" value correctly.


dankmitchell:
[[M] Updating course length doesnt work](https://trello.com/c/eEF5TILv/1629-m-updating-course-length-doesnt-work)

PR #385, opened 2019-06-18
[1624] Refer to @provider instead of course.provider in html

tvararu:
### Context

Refactoring to prepare for validations.

The necessity of this is because when you do a `@course.update`, and it fails, the `@course.provider` is blatted away because the response from the backend API does not contain a provider, only course info. The provider info is still available in the `@provider` though.

This is also terser.

### Guidance to review

I have not manually tested this and am relying on the fact that none of these pages are live yet and mostly have coverage.


tvararu:

PR #386, opened 2019-06-18
[1624] Move site errors and error_messages partials to views/shared

tvararu:
### Context

These will be reused for the course validations.

Also:

- Removes redundant `@errors.any?` on the `render` call
- Moves error partials to render above the `h1`, as is recommended in the design system
- Updates the content to the one that is used on `manage-courses-ui` for enrichment validations

### After (content change)

Screenshot 2019-06-18 at 17 31 50


tvararu:

PR #387, opened 2019-06-19
[1629] Fix other course length

dankmitchell:
### Context
Setting "other" course length

Pass across the "other" length into `course_length` when the user has selected it

### Guidance to review
Try setting "other" for course length

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -11,6 +11,12 @@ class CoursesController < ApplicationController
   def show; end
 
   def update
+    # Course length should be saved as `course_length` so if "other" is selected then pass that text value into `course_length`
+    if params[:course][:course_length_other_length].present? && params[:course][:course_length] == 'Other'
+      params[:course][:course_length] = params[:course][:course_length_other_length]
+    end
+    params[:course].delete(:course_length_other_length)
+
     if @course.update(course_params)
       flash[:success] = 'Your changes have been saved'
       redirect_to(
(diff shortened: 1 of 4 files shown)


dankmitchell:
[[M] Updating course length doesnt work](https://trello.com/c/eEF5TILv/1629-m-updating-course-length-doesnt-work)

app/controllers/courses_controller.rb
```diff
@@ -11,6 +11,11 @@ def index; end
   def show; end
 
   def update
+    # Course length should be saved as `course_length` so if "other" is selected then pass that text value into `course_length`
+    if params[:course][:course_length_other_length].present? && params[:course][:course_length] == 'Other'
```
defong: can we use dig

```ruby
params.dig(:course, :course_length_other_length).present? && params.dig(:course, :course_length) == 'Other'
```

spec/features/courses/fees_spec.rb
```diff
@@ -64,6 +64,28 @@
     expect(current_path).to eq description_provider_course_path('AO', course_1.course_code)
   end
 
+  context 'with course_length_other selected' do
+    let(:course_1) do
+      jsonapi(
+        :course,
+        :with_fees,
+        course_length: '6months',
```
defong:
```suggestion
        course_length: '6 months',
```

closer to real life example

fofr:
When you first select "Other", it should be empty, it shouldn't show our internal representation of the value:

other-field-not-blank

PR #388, opened 2019-06-19
[1624] Display error summary at top of enrichment pages

tvararu:
### Context

We need to display validation errors when users save course description fields.

Renders the shared errors template at the top of the about, fees, salary, and requirements pages.

The forms are updated to each contain a hidden field, for the `page` that they correspond to, to allow the `courses#update` action to render the correct page when responding with validation errors.

4 new `patch` routes are created, one for each page. This is because on the regular `update` endpoint, the URL is `/organisations/ORG/courses/CRS`. If a user takes this URL and submits it in the same browser tab, they'll be taken to the basic details, which is confusing.

### Guidance to review

There will be a follow-up PR to implement the red highlights on the form fields.

### Screenshots

Screenshot 2019-06-19 at 14 31 50
Screenshot 2019-06-19 at 14 33 07
Screenshot 2019-06-19 at 14 33 26

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -119,6 +119,9 @@ group :test do
 
   # Page object for Capybara
   gem 'site_prism'
+
+  # Allows assert_template in request specs
+  gem 'rails-controller-testing'
 end
 
 # Windows does not include zoneinfo files, so bundle the tzinfo-data gem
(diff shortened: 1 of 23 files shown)


tvararu:

PR #389, opened 2019-06-20
[1624] Abstract site form field logic into partial

tvararu:
### Context

This uses the [capture](https://api.rubyonrails.org/classes/ActionView/Helpers/CaptureHelper.html) feature I recently discovered to create a reusable partial for the site form fields. This will be reused in the enrichment validations.

Simplifies `site/_form` by creating a reusable `_field_with_error` partial / component.

### Guidance to review

There should already be adequate test coverage for this, this is just a refactor.

If we're happy with this pattern of doing a reusable frontend template, I would like to extend it to other parts of the frontend. In the future, we can switch to [ActionView::Component](https://github.com/rails/rails/pull/36388).

### DIFF
--- a/app/views/sites/_form.html.erb
+++ b/app/views/sites/_form.html.erb
@@ -1,48 +1,36 @@
-<div class="govuk-form-group<%= " govuk-form-group--error" if @errors[:location_name]&.any? %>">
-  <%= f.label :location_name, class: "govuk-label" do %>
-    <h2 class="govuk-heading-m<%= " govuk-!-margin-bottom-0" if @errors[:location_name]&.any? %>">Name</h2>
-  <% end %>
-  <%= render "shared/error_messages", field: :location_name %>
-  <%= f.text_field :location_name, class: "govuk-input govuk-!-width-full", aria: { describedby: @errors[:location_name] ? "location_name-error" : '' }, data: { qa: 'location__name' } %>
-</div>
+<%= render "shared/form_field",
+  form: f, field: :location_name, qa: 'location__name', label: capture { %>
+    <h2 class="<%= cns("govuk-heading-m", "govuk-!-margin-bottom-0": @errors[:location_name]&.any?) %>">Name</h2>
+  <% }
+%>
 
 <h2 class="govuk-heading-m govuk-!-margin-top-8">Address</h2>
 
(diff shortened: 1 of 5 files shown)


app/views/shared/_field_with_error.html.erb
```diff
@@ -0,0 +1,31 @@
+<%
+  label ||= ''
+  type ||= :text_field
+  qa ||= ''
+  select_options ||= []
+  input_options = {
```
fofr: I'd prefer if we were passing in an input to this partial. Probably need some sort of helper for the aria-describedby attributes.

app/views/sites/_form.html.erb
```diff
@@ -1,48 +1,36 @@
...
-</div>
+  <% }
+%>
 
-<div class="govuk-form-group<%= " govuk-form-group--error" if @errors[:address3]&.any? %>">
-  <%= f.label :address3, "Town or city", class: "govuk-label" %>
-  <%= render "shared/error_messages", field: :address3 %>
-  <%= f.text_field :address3, class: "govuk-input govuk-!-width-full", aria: { describedby: @errors[:address3] ? "address3-error" : '' } %>
-</div>
+<%= render "shared/field_with_error",
```
fofr: I think the partial should be named form_field, rather than field_with_error

app/views/shared/_field_with_error.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    class: type == :select ? "govuk-select" : "govuk-input govuk-!-width-full",
```
fofr: These classes need to be customisable

app/views/shared/_field_with_error.html.erb
```diff
@@ -0,0 +1,31 @@
...
+    },
+    data: {
+      qa: qa
+    }
+  }
+%>
+
+<div class="govuk-form-group<%= " govuk-form-group--error" if @errors[field]&.any? %>">
+  <% if label.blank? %>
+    <%= f.label field, class: "govuk-label" %>
```
fofr: I wonder if we should be explicit about this Rails behaviour, ie:
```
label ||= field.humanize
```

tvararu: `.to_s.humanize` probably 👍

tvararu:

PR #390, opened 2019-06-20
Fix success summary and enrichment back links

fofr:
### Context

- Back links on enrichment pages linked to the Basic details tab rather than the Description tab
- The success summary on publishing duplicated the publish message

### Before
Screen Shot 2019-06-20 at 12 08 36


fofr:
The routes we use for the description tab and the details tab are the wrong way around:
https://trello.com/c/cOUm6DTu/1662-course-route-path-should-be-description-tab-not-the-basic-details-one

Description should be the default.

PR #391, opened 2019-06-20
[1624] Wrap all enrichment fields with form component

tvararu:
### Context

This allows them to display the validation error red highlight. `course_length` is treated as a special case.


tvararu:

PR #392, opened 2019-06-21
[1669] Iterate vacancies design for single site/single study mode courses

fofr:
### Context

Update the code to match the design:
https://bat-design-history.herokuapp.com/publish-teacher-training/vacancies-iteration-14-jan

We built the complex version of vacancies when we launched the feature, but we didn't return to it to create the simpler toggle. The toggle should be used when there is only 1 site and the course has one study mode.

- Add check for if course is single site/single study mode
- Show a checkbox for toggling vacancies
- Use different update logic and template for single site/single study mode
- Refactor vacancies tests to:
   - use SitePrism
   - be explicit about the study mode and number of sites when testing a course

### Guidance to review

| State | No error | With error |
|--|--|--|
| With vacancies |Screen Shot 2019-06-21 at 14 05 45|Screen Shot 2019-06-21 at 14 05 53|
| No vacancies |Screen Shot 2019-06-21 at 14 06 13 |Screen Shot 2019-06-21 at 14 06 17|

### DIFF
--- a/app/controllers/courses/vacancies_controller.rb
+++ b/app/controllers/courses/vacancies_controller.rb
@@ -8,6 +8,39 @@ module Courses
     def edit; end
 
     def update
+      unless params[:change_vacancies_confirmation]
+        error_message = if @course.has_vacancies?
+                          "Please confirm there are no vacancies to close applications"
+                        else
+                          "Please confirm there are vacancies to reopen applications"
+                        end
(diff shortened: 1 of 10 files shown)


fofr:
[Iterate vacancies design for single site courses](https://trello.com/c/Bb9KBLVP/1669-iterate-vacancies-design-for-single-site-courses)

PR #396, opened 2019-06-24
Update copy specs to include custom course length

fofr:
Make sure that copying custom course lengths is tested and works as expected.

### DIFF
--- a/spec/features/courses/fees_spec.rb
+++ b/spec/features/courses/fees_spec.rb
@@ -114,7 +114,7 @@ feature 'Course fees', type: :feature do
         :course,
         name: 'Biology',
         provider: provider,
-        course_length: 'TwoYears',
+        course_length: 'Something custom',
         fee_uk_eu: 9500,
         fee_international: 1200,
         fee_details: 'Some information about the fees',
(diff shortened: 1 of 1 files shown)


spec/features/courses/fees_spec.rb
```diff
@@ -156,7 +156,8 @@
       end
 
       expect(course_fees_page.course_length_one_year).to_not be_checked
-      expect(course_fees_page.course_length_two_years).to be_checked
+      expect(course_fees_page.course_length_other).to be_checked
```
defong: expect(course_fees_page.course_length_two_years).to_not be_checked

can we add that line in please

spec/features/courses/fees_spec.rb
```diff
@@ -182,7 +183,8 @@
...
```
defong: ```expect(course_fees_page.course_length_two_years).to_not be_checked```

can we add that line in please

spec/features/courses/fees_spec.rb
```diff
@@ -114,7 +114,7 @@
         :course,
         name: 'Biology',
         provider: provider,
-        course_length: 'TwoYears',
+        course_length: 'Something custom',
```
defong: ``` context 'when copying course fees from another course with custom course length' do```

can we change that as well please

PR #401, opened 2019-06-25
[1484] Make accredited body sorting case insensitive

fofr:
### Context

A couple of providers begin with a lowercase letter, eg e-Qualitas. Courses associated with this accredited body were appearing in a block at the bottom of the page instead of the expected alphabetical position.

| Before | After |
|-|-|
| Screen Shot 2019-06-25 at 16 16 52 | Screen Shot 2019-06-25 at 16 16 08 |

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -161,7 +161,7 @@ private
           @provider.provider_name
         end
       }
-      .sort_by { |accrediting_provider, _| accrediting_provider }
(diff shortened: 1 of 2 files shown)


fofr:
[Sorting shouldn’t put words beginning with a lowercase letter at the end](https://trello.com/c/GzswrLn2/1484-sorting-shouldnt-put-words-beginning-with-a-lowercase-letter-at-the-end)

PR #407, opened 2019-06-26
[1298] 404 if site not found

fofr:
We shouldn't throw a 500 if the provider doesn't have the specified site ID.

### Context

When editing a site it throws a 500 if you mess with an edit location URL and put in a bad ID.

Because it can’t find a site you get:
```undefined method 'location_name' for nil:NilClass```

For example, a bad URL: `/organisations/135/locations/1000000000000/edit`

- Stop the 500
- Show a 404

### DIFF
--- a/app/controllers/sites_controller.rb
+++ b/app/controllers/sites_controller.rb
@@ -43,7 +43,12 @@ private
 
   def build_site
     @site = @provider.sites.find { |site| site.id == params[:id] }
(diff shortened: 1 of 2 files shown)


fofr:
[Edit location 500s if location not found](https://trello.com/c/02Vk32Vc/1298-edit-location-500s-if-location-not-found)

PR #408, opened 2019-06-26
[1664] Retain enrichment error messages on edit

dankmitchell:
- Update validation links to navigate to the rails frontend app
- Display validation messages on the individual enrichment pages

### Guidance to review
- Try and publish an invalid course then click one of the validation message links

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -44,12 +44,7 @@ class CoursesController < ApplicationController
   end
 
   def about
-    if params[:display_errors].present?
-      attributes = %i[about_course interview_process how_school_placements_work]
-
-      @course.publish
-      @errors = @course.errors.messages.select { |key| attributes.include? key }
-    end
+    show_deep_linked_errors(%i[about_course interview_process how_school_placements_work])
 
     if params[:copy_from].present?
       @copied_fields = [
@@ -61,12 +56,7 @@ class CoursesController < ApplicationController
   end
 
   def requirements
-    if params[:display_errors].present?
-      attributes = %i[required_qualifications personal_qualities other_requirements]
-
-      @course.publish
-      @errors = @course.errors.messages.select { |key| attributes.include? key }
-    end
+    show_deep_linked_errors(%i[required_qualifications personal_qualities other_requirements])
 
     if params[:copy_from].present?
       @copied_fields = [
@@ -78,12 +68,7 @@ class CoursesController < ApplicationController
   end
 
   def fees
-    if params[:display_errors].present?
-      attributes = %i[course_length fee_uk_eu fee_international fee_details financial_support]
-
-      @course.publish
-      @errors = @course.errors.messages.select { |key| attributes.include? key }
-    end
+    show_deep_linked_errors(%i[course_length fee_uk_eu fee_international fee_details financial_support])
 
     if params[:copy_from].present?
       @copied_fields = [
@@ -97,12 +82,7 @@ class CoursesController < ApplicationController
   end
 
   def salary
-    if params[:display_errors].present?
-      attributes = %i[course_length salary_details]
-
-      @course.publish
-      @errors = @course.errors.messages.select { |key| attributes.include? key }
-    end
+    show_deep_linked_errors(%i[course_length salary_details])
 
     if params[:copy_from].present?
       @copied_fields = [
@@ -230,5 +210,12 @@ private
 
   def initialise_errors
     @errors = {}
+  end
+
+  def show_deep_linked_errors(attributes)
+    return if params[:display_errors].blank?
+
+    @course.publishable?
+    @errors = @course.errors.messages.select { |key| attributes.include? key }
   end
 end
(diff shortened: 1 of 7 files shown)


tvararu:

app/controllers/courses_controller.rb
```diff
@@ -42,6 +42,22 @@ def description
   end
 
   def about
+    if params[:display_errors].present?
+      attributes = %w[about_course interview_process how_school_placements_work]
+
+      errors = @course.publish(provider_code: @provider.provider_code).errors
```
fofr: If I shared a link with `display_errors` in the URL, but the errors had been fixed, would this accidentally publish the course?

tvararu: Yes, good point, which is why we should do a custom endpoint instead...

dankmitchell: No, because we hit the publish endpoint which would return no errors if there are no errors

tvararu:
Have simplified this with the new `.publish` action but still feels like we need a `publishable?` action instead that just validates on the server... going to take a stab at that.

tvararu:
This needs to be merged/deployed AFTER https://github.com/DFE-Digital/manage-courses-backend/pull/578

tvararu:
Added `[DO NOT MERGE]` until backend is deployed.

fofr:
Testing this locally, the error propagates as expected through to About this course and Fees, but not Requirements. I'll investigate.

fofr:
The inconsistency is because in the backend the qualifications attribute is called `qualifications` (https://github.com/DFE-Digital/manage-courses-backend/blob/master/app/deserializers/api/v2/deserializable_course.rb#L19) vs `required_qualifications` in frontend (https://github.com/DFE-Digital/manage-courses-frontend/blame/1664-retain-enrichment-error-messages-on-edit/app/helpers/view_helper.rb#L45)

I suggest we fix this separately.

app/helpers/view_helper.rb
```diff
@@ -37,12 +37,12 @@ def enrichment_error_url(provider_code:, course:, field:)
...
-      'fee_uk_eu' => base + '/fees#fee_uk_eu_wrapper',
-      'course_length' => base + (course.has_fees? ? '/fees' : '/salary') + '#course_length_wrapper',
-      'salary_details' => base + '/salary#salary_details_wrapper',
-      'qualifications' => base + '/requirements#required_qualifications_wrapper'
+      'about_course' => base + '/about?display_errors=true#about_course_wrapper',
+      'how_school_placements_work' => base + '/about?display_errors=true#how_school_placements_work_wrapper',
+      'fee_uk_eu' => base + '/fees?display_errors=true#fee_uk_eu_wrapper',
+      'course_length' => base + (course.has_fees? ? '/fees' : '/salary') + '?display_errors=true#course_length_wrapper',
+      'salary_details' => base + '/salary?display_errors=true#salary_details_wrapper',
+      'qualifications' => base + '/requirements?display_errors=true#required_qualifications_wrapper'
```
misaka: When would `display_errors` ever be missing/false? And would `@course.errors` ever be populated in that case?

fofr: It's not there when you visit `/fees` directly

misaka: I can't quite think of how `@course.errors` would have any errors in that case, though.

app/models/course.rb
```diff
@@ -73,4 +61,26 @@ def running_site_statuses
   def has_multiple_running_sites_or_study_modes?
     running_site_statuses.length > 1 || full_time_or_part_time?
   end
+
+private
+
+  def post_request(path)
+    base_url = "#{Course.site}providers/#{provider.provider_code}/courses/#{course_code}"
```
misaka:
```suggestion
    base_url = "#{Course.site}#{Course.path}/%{course_code}" % path_attributes
```

JSONAPIClient provides information about how to build this URL. There may be better ways to pull this path out, but with this change at least we'd be DRYing up how we generate the path here. Here's what these components look like in pry:

```
[4] pry(#<Course>)> Course.path
=> "providers/%{provider_code}/courses"
[5] pry(#<Course>)> path_attributes
=> {:provider_code=>"1MN", :course_code=>"27XB"}
```

app/helpers/view_helper.rb
misaka: @dankmitchell Do you know if we need this added bit machinery `display_errors`? I'd be keen to strip out anything we don't need.

dankmitchell: It's the only way we found to trigger re-validation in the controller for each page...

misaka: I see now. `display_errors` is used to avoid posting to the backend needlessly. This makes complete sense. :shipit:

app/models/course.rb
dankmitchell: Sorted

fofr:
Bug with requirements that was found while testing this is being tracked separately:
https://trello.com/c/BnKmktuz/1747-fix-inconsistent-qualifications-requiredqualifications-field-names

PR #411, opened 2019-06-27
[1620] It's grade 4 not level 4

fofr:
### Context

Content should always read 'grade 4' not 'level 4' as GCSEs are not qualification level 4. Highlighted in a show and tell.


fofr:
[Change 'level 4' to 'grade 4'](https://trello.com/c/iTlgKH6A/1620-change-level-4-to-grade-4)

PR #414, opened 2019-06-28
[1671] Course fee improvements

tvararu:
### Context

Users are currently unable to save the empty string value on the course fee fields.

#### Mark fee attributes on course as strings

This prevents `json_api_client` from typecasting a valid value of '' (empty string) to `0`, which are two separate values in our system. It also removes some weird behaviour, like how `9,000.00` is cast to `9`.

We should revisit our own (or rather, now lack of any) casting and validations to make sure we're doing as much work as we can so that users don't constantly bump into "X is not a number" type errors when they type reasonable values like `£9,000.00`.

#### Use text_fields instead of number_fields

[The Design System encourages using input type='text' for currency values](https://design-system.service.gov.uk/components/text-input/). While it is not explicitly stated, it is because type='number' introduces potential accessibility issues.

### Guidance to review

No tests as it feels like those would be testing the web platform and `json_api_client` respectively.

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -6,8 +6,8 @@ class Course < Base
   custom_endpoint :sync_with_search_and_compare, on: :member, request_method: :post
   custom_endpoint :publish, on: :member, request_method: :post
 
-  property :fee_international, type: :integer
-  property :fee_uk_eu, type: :integer
+  property :fee_international, type: :string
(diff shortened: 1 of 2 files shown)


tvararu:
[Saving an empty fee gets converted to 0](https://trello.com/c/weZDU1qy/1671-saving-an-empty-fee-gets-converted-to-0)

PR #417, opened 2019-07-01
1706 setup recruitment cycles

dankmitchell:
### Context
Supporting multiple recruitment years

Add `recruitment_cycle` scope to courses and locations

### Guidance to review
- No visual changes
- All legacy routes will be redirected in a subsequent PR

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -22,8 +22,9 @@ class CoursesController < ApplicationController
     if @course.update(course_params)
       flash[:success] = 'Your changes have been saved'
       redirect_to(
-        provider_course_path(
+        provider_recruitment_cycle_course_path(
           @course.provider_code,
+          @course.recruitment_cycle_year,
           @course.course_code
         )
       )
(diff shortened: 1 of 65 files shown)


dankmitchell:
[Setup recruitment cycles](https://trello.com/c/sb1Hggbk/1706-setup-recruitment-cycles)

fofr:
We'll need to be making the template changes conditional – ie in October when we go back to having just one cycle it should revert back to the current design.

app/controllers/courses_controller.rb
```diff
@@ -22,8 +22,9 @@ def update
     if @course.update(course_params)
       flash[:success] = 'Your changes have been saved'
       redirect_to(
-        provider_course_path(
+        provider_recruitment_cycle_course_path(
           @course.provider_code,
+          @course.recruitment_cycle,
```
misaka: I know this is a pain, but considering at some point we're possibly going to actually want the `recruitment_cycle` object in mcfe, should we be using `@course.recruitment_cycle_year` here?

PR #422, opened 2019-07-03
[1664] Replace course publish method

tvararu:
### Context

Rewriting publish validations so they are not as hacky on the frontend and can be displayed on multiple pages.

Write a custom method for the `course.publish` action that allows us to manually pass in an empty `attributes` hash. This is to enable upcoming changes to the backend that will make it associate `pointer`s correctly on the json-api response when publishing.

### Guidance to review

Tests already exist for this. :shipit:


tvararu:

PR #424, opened 2019-07-03
[1664] Handle publish errors better

tvararu:
### Context

Making validation errors consistent across the course.

This makes the `@course.publish` method behave more similarly to `.save`, where it automatically populates `@course.errors` with the correct hash of errors. This is only possible now that the backend gives us back `pointer`s on the error object returned from `publish`.

There will be follow-up work to change this from using a flash to doing a `render` instead.

### Guidance to review

This will break if it's deployed without https://github.com/DFE-Digital/manage-courses-backend/pull/572. We should deploy backend end to end first.

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -90,16 +90,10 @@ class CoursesController < ApplicationController
   def preview; end
 
   def publish
-    errors = @course.publish(provider_code: @provider.provider_code).errors
-    if errors.present?
-      flash[:error_summary] = errors.map { |error|
-        [
-          error[:title].last(error[:title].length - 'Invalid '.length),
-          error[:detail]
-        ]
-      }.to_h
-    else
+    if @course.publish
       flash[:success] = "Your course has been published."
+    else
+      flash[:error_summary] = @course.errors.messages
     end
 
     redirect_to provider_course_path(@provider.provider_code, @course.course_code)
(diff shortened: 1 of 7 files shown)


tvararu:

app/controllers/courses_controller.rb
```diff
@@ -90,16 +90,10 @@ def delete; end
...
-        [
-          error[:title].last(error[:title].length - 'Invalid '.length),
-          error[:detail]
-        ]
-      }.to_h
-    else
+    if @course.publish
       flash[:success] = "Your course has been published."
+    else
+      flash[:error_summary] = @course.errors.messages
```
tvararu: (3 PRs to get this wonderful syntax finally working!)

app/helpers/course_helper.rb
```diff
@@ -5,15 +5,17 @@ def course_manage_error_link(field, error)
...
   end
 
   def course_summary_label(key, field)
     if @errors&.key? field
-      error = @errors[field]
       content_tag :dt, class: 'govuk-summary-list__key course-parts__fields__label--error' do
-        content_tag(:span, key) + course_manage_error_link(field, error)
+        [
+          content_tag(:span, key),
+          *@errors[field].map { |error| course_manage_error_link(field, error) }
```
fofr: What does the `*` do here?

tvararu: It's the [Ruby splat operator](https://www.honeybadger.io/blog/ruby-splat-array-manipulation-destructuring/), which is similar to the JavaScript spread operator `...`, which depending on the context will do a few things but mostly means "take this array / hash and turn it into individual things".

tvararu: This code is generating an array of error links and then joining them up into an array with the above `content_tag`, which then get added up together using `.reduce(:+)`.

PR #427, opened 2019-07-04
Don't warn users about unsaved changes on save

fofr:
### Context

We shouldn't show the "You have unsaved changes" warning when a user tries to save their changes.

### Guidance to review

- Copy course content on an enrichment page
- Try to click save

### DIFF
--- a/app/webpacker/packs/application.js
+++ b/app/webpacker/packs/application.js
@@ -9,6 +9,14 @@ new CookieMessage($cookieMessage).init();
 
 var $copyWarningMessage = document.querySelector('[data-copy-course="warning"]');
 if ($copyWarningMessage) {
+  var $enrichmentForm = document.querySelector('[data-qa="enrichment-form"]');
+
+  if ($enrichmentForm) {
+    $enrichmentForm.addEventListener("submit", function() {
+      window.onbeforeunload = null;
+    });
+  }
+
   window.onbeforeunload = function() {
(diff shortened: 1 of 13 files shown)


app/webpacker/packs/application.js
```diff
@@ -9,6 +9,14 @@ new CookieMessage($cookieMessage).init();
 
 var $copyWarningMessage = document.querySelector('[data-copy-course="warning"]');
 if ($copyWarningMessage) {
+  var $enrichmentForm = document.querySelector('[data-qa="enrichment-form"]');
```
tvararu: `data-qa` is for testing hooks, this is for hooking in a behaviour. We could use an `id` instead, but it's not a big deal.

app/webpacker/packs/application.js
```diff
@@ -9,6 +9,14 @@ new CookieMessage($cookieMessage).init();
...
+
+  if ($enrichmentForm) {
```
tvararu: This seems weird... would it not work if we moved the call below to `window.onbeforeunload` to inside an `if` instead of setting it to null on submit?

```js
if (!$enrichmentForm) {
  window.onbeforeunload = function() {
	  return 'You have unsaved changes, are you sure you want to leave?'
	}
}

tvararu: I've realised the logic is slightly different after it was pointed out irl, disregard 👍

PR #430, opened 2019-07-04
[1387] Add rollover screen

fofr:
### Context

- Show rollover screen to transitioned users
- Continue showing transition screen to users who haven't seen it yet, they'll see this first (still about 60 per week seeing it)
- Only show the rollover screen once per user
- Introduce a new "rolled_over" state

Goes with https://github.com/DFE-Digital/manage-courses-backend/pull/577

Screen Shot 2019-07-04 at 15 09 57

### Guidance to review

We probably want to disable the redirect to the rollover screen until we are ready for it. What's the best way to do that?

### DIFF
--- a/spec/controllers/users_controller_spec.rb
+++ b/spec/controllers/users_controller_spec.rb
@@ -22,9 +22,9 @@ describe UsersController, type: :controller do
         stub_api_v2_request("/users/#{user.id}/accept_transition_screen", {}, :patch, 200)
       end
 
-      it "redirects to providers index" do
+      it "redirects to the rollover screen" do
         get :accept_transition_info
-        expect(response).to redirect_to(providers_path)
+        expect(response).to redirect_to(rollover_path)
       end
     end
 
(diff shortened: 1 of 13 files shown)


fofr:
[[S] Rollover Screen - first log-in for users when next cycle is available](https://trello.com/c/CqD9fZa4/1387-s-rollover-screen-first-log-in-for-users-when-next-cycle-is-available)

spec/controllers/users_controller_spec.rb
```diff
@@ -22,7 +22,40 @@
...
+        get :accept_transition_info
+        expect(Raven).to have_received(:capture_exception).with(instance_of(JsonApiClient::Errors::ClientError))
+        expect(response).to redirect_to(rollover_path)
+      end
+    end
+  end
+
+  describe 'GET #accept_transition_info when rollover is disabled' do
+    before do
+      Settings.rollover = false
```
fofr: Must be a better way to do this.

benilovj: https://github.com/rspec/rspec-mocks#method-stubs

Something like

```ruby
allow(Settings).to receive(:rollover).and_return(false)
```

PR #432, opened 2019-07-05
[1394] Courses table for next and current cycles during rollover

fofr:
### Context

https://bat-design-history.herokuapp.com/publish-teacher-training/publishing-during-rollover

- Filter courses by recruitment_cycle_year
- During a rollover period, indicate if the courses shown are for the current or next cycle
- Add helpers to determine which cycle is being shown
- Show a slightly different table for the next cycle
- Show a simplified courses table for all cycles after 2019

What's intentionally missing from this PR:
- Breadcrumbs
- A link to a different "Add course" Google form
- Handling of the "Applications" column (depends on what we intend to do with rolled over courses)
- Handling of the courses page when the current cycle becomes a previous cycle

### Guidance to review
- Modify some local courses to be in the next recruitment cycle in backend, for example:
```ruby
next_cycle = RecruitmentCycle.all[1]
courses[0..3].each { |c| c.recruitment_cycle = next_cycle; c.save }
```
- Change `2019` to `2020` in the URL

#### Next cycle
Screen Shot 2019-07-05 at 14 52 58

(note: The draft course here shouldn't say it'll be on Find, this is because the copied course has a UCAS status of running, which needs to be reset to new)

#### Current cycle during rollover
Screen Shot 2019-07-05 at 14 46 36

#### Current cycle before we enable rollover
Screen Shot 2019-07-05 at 14 53 48

### DIFF
--- a/app/helpers/recruitment_cycle_helper.rb
+++ b/app/helpers/recruitment_cycle_helper.rb
@@ -1,0 +1,31 @@
+module RecruitmentCycleHelper
+  def rollover?
+    Settings.rollover
+  end
+
+  def current_cycle?
+    cycle_year == Settings.current_cycle
+  end
+
+  def next_cycle?
+    cycle_year == Settings.current_cycle + 1
+  end
+
+  def recruitment_cycle_title
+    year_range = "#{cycle_year} – #{cycle_year + 1}"
+
+    if current_cycle?
+      "Current cycle (#{year_range})"
+    elsif next_cycle?
+      "Next cycle (#{year_range})"
+    else
+      year_range
+    end
+  end
+
+private
+
+  def cycle_year
+    params[:recruitment_cycle_year].to_i
+  end
+end
--- a/app/views/courses/_course_table_row.html.erb
+++ b/app/views/courses/_course_table_row.html.erb
@@ -6,10 +6,12 @@
     %>
     <span class="govuk-body-s"><%= course.description %></span>
   </td>
-  <td class="govuk-table__cell" data-qa="courses-table__ucas-status">
-    <%= course.ucas_status %>
-  </td>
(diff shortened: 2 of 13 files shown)


fofr:
[[S] Courses Table - Next cycle](https://trello.com/c/4gdcwkIg/1394-s-courses-table-next-cycle)

app/views/courses/_course_table_row.html.erb
```diff
@@ -6,10 +6,12 @@
     %>
     <span class="govuk-body-s"><%= course.description %></span>
   </td>
+  <% if show_legacy_courses_table? %>
   <td class="govuk-table__cell" data-qa="courses-table__ucas-status">
```
benilovj: should indent here for readability

app/helpers/recruitment_cycle_helper.rb
```diff
@@ -0,0 +1,31 @@
+module RecruitmentCycleHelper
```
benilovj: Having a fair amount of logic in a helper is a bit smelly. One way to fix that would be to move a lot of these methods onto a `RecruitmentCycle` model (which is a plain Ruby object). Its initializer could take the cycle year and maybe Settings.current_cycle.

dankmitchell: We can leave `recruitment_cycle_title` here but move the rest into a model

app/views/courses/_course_table_row.html.erb
fofr: Fixed

app/helpers/recruitment_cycle_helper.rb
fofr: Will address in a subsequent PR.

dankmitchell:
Sorry @fofr. Merge conflicts. I can fix up 💪

PR #436, opened 2019-07-08
[1722] Warn user when they leave unsaved/changed form

markmead:
### Context

Setup check on forms with `data-module="form"` attribute that prompts the user that they have unsaved changes when attempting to leave the page. 

JS module to handle form checks when user leaves with unsaved changes.

### Notes

This might seem excessive but was recommended to cover all browsers/any deprecations.

```
event.preventDefault();
event.returnValue = "You have unsaved changes, are you sure you want to leave?";
return "You have unsaved changes, are you sure you want to leave?";
```


app/webpacker/scripts/form-check-leave.js
```diff
@@ -0,0 +1,33 @@
...
+
+FormCheckLeave.prototype.action = function($form, $data) {
+  const $changes = this.data();
+  $form.addEventListener("submit", () => (window.onbeforeunload = null));
+
+  window.onbeforeunload = function() {
+    event.preventDefault();
+    event.returnValue =
+      "You have unsaved changes, are you sure you want to leave?";
+    return "You have unsaved changes, are you sure you want to leave?";
```
dankmitchell: Worth putting a comment in here explaining why we have two return values?

dankmitchell:
Also, please squash the commits on merge

app/webpacker/packs/application.js
```diff
@@ -18,6 +21,9 @@ if ($copyWarningMessage) {
   }
 
   window.onbeforeunload = function() {
-    return 'You have unsaved changes, are you sure you want to leave?'
-  }
+    return "You have unsaved changes, are you sure you want to leave?";
+  };
 }
+
+const $form = document.querySelector('[data-module="form"]');
```
fofr: Could we use `form-check-leave` as module value?

app/webpacker/scripts/form-check-leave.js
```diff
@@ -0,0 +1,33 @@
...
+  if (!$form) return;
+  const $data = this.data();
+  this.action($form, $data);
+};
+
+FormCheckLeave.prototype.action = function($form, $data) {
+  const $changes = this.data();
+  $form.addEventListener("submit", () => (window.onbeforeunload = null));
+
+  window.onbeforeunload = function() {
```
fofr: Do we need to pass in event here?
`window.onbeforeunload = function(event) {`

fofr: Do we need some comparison of original data and new data here?

markmead: `event` is always present but I will add it in for clarification. 

Have a fix for the comparison, not sure when that disappeared.

app/webpacker/scripts/form-check-leave.js
```diff
@@ -0,0 +1,30 @@
...
+  const $form = this.$module;
+
+  if ($form) {
+    const $originalFormContent = encodeURIComponent(
+      Array.from(new FormData($form))
+    );
+
+    window.onbeforeunload = function(event) {
+      const $updatedFormContent = encodeURIComponent(
+        Array.from(new FormData($form))
```
fofr: I think `Array.from` will break on IE11 (which is about 17% of users):
https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Browser_compatibility

markmead: What's your recommended fix? It was this vs a giant for loop which looked fragile.

For this type of function I'd recommend jQuery `serialize` but that's not my call to make.

dankmitchell: I think something like this would be fine; https://gomakethings.com/how-to-serialize-form-data-with-vanilla-js/. The jQuery method is convenient but we don't want to add jQuery.

markmead: Requires all fields in the array to have a `name` attribute. They all should have so reusability shouldn't be an issue but just worry it's easy to break. I'll have a play around.

fofr:
[Warn user when leaving an enrichment page with unsaved changes](https://trello.com/c/3fBK0dge/1722-warn-user-when-leaving-an-enrichment-page-with-unsaved-changes)

app/webpacker/scripts/form-check-leave.js
fofr: We could use an Array.from polyfill? https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Array/from#Polyfill

markmead: Got what Dan sent working but would just like to trial what I've pushed up. The use of `change` in the event listener is a cleaner solution that's worked for me.

app/webpacker/scripts/form-check-leave.js
```diff
@@ -0,0 +1,31 @@
...
+  const $form = this.$module;
+  let $change = false;
+
+  if ($form) {
+    $form.addEventListener("submit", function() {
+      window.onbeforeunload = null;
+    });
+
+    $form.addEventListener("change", function() {
+      $change = true;
```
fofr: Nice :+1: This works for me when testing locally.

I've found one small bug – when a user copies content from another course we add an onbeforeunload too:
https://github.com/DFE-Digital/manage-courses-frontend/blob/4d58255a5b3ba3c87d9f96b4ce16773480ab7751/app/webpacker/packs/application.js#L23-L25

This gets replaced with an empty function if nothing is changed now.

Steps to reproduce:
- Got to About this course
- Copy from another course (an orange box will show)
- When leaving this page (without making further changes), we need to warn of unsaved changes – this is because what was copied is not yet saved

PR #442, opened 2019-07-09
[1742] Refractor CSS class 

markmead:
### Context

Add `app-` to classes that aren't from the GOV design system


markmead:
[Update CSS class names to have "app" if not using GOV class](https://trello.com/c/8p48XP8O/1742-update-css-class-names-to-have-app-if-not-using-gov-class)

PR #443, opened 2019-07-10
[1748] Record form errors/success on publish

markmead:
### Context

Record form errors/success on publish and send to Google Analytics.

Added the `data-ga-event-form` to error element and pass in values "error" or "success"

### DIFF
--- a/app/webpacker/scripts/form-error-tracking.js
+++ b/app/webpacker/scripts/form-error-tracking.js
@@ -1,0 +1,7 @@
+export function triggerFormAnalytics(category, action, label) {
+  ga("send", "event", {
+    eventCategory: category,
+    eventAction: action,
+    eventLabel: label
+  });
+}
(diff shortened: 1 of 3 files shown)


app/webpacker/scripts/form-error-tracking.js
```diff
@@ -0,0 +1,7 @@
+export function triggerFormAnalytics(category, action, label) {
```
fofr: This method looks generic enough to be something like `sendEventToGa`

markmead: This is consistent with **search-and-compare-ui** but happy to change.

fofr:
[Record each success and error message to Google Analytics as a GA Event](https://trello.com/c/vS2xhvvx/1748-record-each-success-and-error-message-to-google-analytics-as-a-ga-event)

PR #448, opened 2019-07-10
[1678] Added back link to guidance page

markmead:
### Context

Add a button to return to the referrer page so that the page no longer feels like a dead end.

Button to return back to the page the user came from no matter where they are.

### Guidance to review

Should this become a component/partial that we can reuse on other pages? Such as cookies, terms etc.

### DIFF
--- a/app/views/pages/guidance.html.erb
+++ b/app/views/pages/guidance.html.erb
@@ -1,5 +1,13 @@
 <%= content_for :page_title, "Guidance" %>
 
+<% content_for :before_content do %>
+  <% if request.referrer.present? %>
+    <%= govuk_back_link_to(:back) %>
+  <% else %>
+    <%= govuk_back_link_to(providers_path) %>
+  <% end %>
+<% end %>
+
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-two-thirds">
     <h1 class="govuk-heading-xl">Guidance for Publish teacher training courses</h1>


markmead:
[Add 'back' link to publishing guidance page](https://trello.com/c/rDEirWaq/1678-add-back-link-to-publishing-guidance-page)

app/views/pages/guidance.html.erb
```diff
@@ -1,5 +1,9 @@
 <%= content_for :page_title, "Guidance" %>
 
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(:back) %>
```
dankmitchell: What if the `referrer` is nil?

I think it should take them back to `provider#index` which will force them to sign in...

fofr: If it's nil I think it defaults to `history.back`

markmead: Gonna test this now.

fofr: What happens to `providers_path` if the user isn't logged in?

dankmitchell: The get taken to DfE-Sign

fofr: What happens if the user has only 1 organisation?

dankmitchell: They are re-directed to the `org#show` view

PR #452, opened 2019-07-11
[1758] Add URLs and entries for next cycle course forms

fofr:
### Context

Now the new forms are set up, add links to them and pre-fill the correct fields.
Link to the right form based on current or next cycle and accredited or unaccredited providers.

### DIFF
--- a/app/helpers/provider_helper.rb
+++ b/app/helpers/provider_helper.rb
@@ -1,6 +1,6 @@
 module ProviderHelper
-  def add_course_url(email, provider, current)
-    cycle_key = current ? 'current_cycle' : 'next_cycle'
+  def add_course_url(email, provider, is_current_cycle:)
+    cycle_key = is_current_cycle ? 'current_cycle' : 'next_cycle'
 
     if provider.accredited_body?
       google_form_url_for(Settings.google_forms[cycle_key].new_course_for_accredited_bodies, email, provider)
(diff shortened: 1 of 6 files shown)


fofr:
[Iterate 'Add new Course' Google sheet](https://trello.com/c/wdxRrNDy/1758-iterate-add-new-course-google-sheet)

app/helpers/provider_helper.rb
```diff
@@ -1,14 +1,16 @@
...
+      google_form_url_for(Settings.google_forms[cycle_key].new_course_for_accredited_bodies, email, provider)
     else
-      google_form_url_for(Settings.google_forms.new_course_for_unaccredited_bodies, email, provider)
+      google_form_url_for(Settings.google_forms[cycle_key].new_course_for_unaccredited_bodies, email, provider)
     end
   end
 
-  def add_course_link(email, provider)
-    link_to "Add a new course", add_course_url(email, provider), class: "govuk-button govuk-!-margin-bottom-2", rel: "noopener noreferrer", target: "_blank"
+  def add_course_link(email, provider, current)
```
benilovj: You might make this method signature clearer by:
- using a named parameter
- expanding the variable name (`current` is a bit ambiguous)
- naming it in a way that makes it clear that it's a boolean

eg.
```suggestion
  def add_course_link(email, provider, is_current_cycle:)
```

fofr: Added in c4ef56d5553c66e8c58788f5bfe2437f546cc35a

PR #457, opened 2019-07-12
[1764] Accessibility testing and semantic updates

markmead:
### Context

Fixed accessibility issues and tidy up semantic HTML

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -73,18 +73,18 @@
 <div class="course-tabs govuk-tabs">
   <h2 class="govuk-tabs__title">Contents</h2>
 
-  <ul class="govuk-tabs__list">
+  <ul class="govuk-tabs__list" role="tablist">
     <li class="govuk-tabs__list-item" role="presentation">
       <%=
         link_to "#description",
-        class: "govuk-tabs__tab govuk-tabs__tab--selected", id: "description", role: 'tab', tabindex: '-1', aria: { selected: true } do
+        class: "govuk-tabs__tab govuk-tabs__tab--selected", role: 'tab', tabindex: '-1', aria: { selected: true } do
       %>
         Description<br>
         <span class="govuk-body-s govuk-!-font-weight-regular pointer-events-none">Content, fees and eligibility</span>
       <% end %>
     </li>
 
-    <li class="govuk-tabs__list-item">
+    <li class="govuk-tabs__list-item" role="presentation">
       <%=
         link_to details_provider_recruitment_cycle_course_path(@provider.provider_code, course.recruitment_cycle_year, course.course_code),
         class: "govuk-tabs__tab" do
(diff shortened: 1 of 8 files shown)


app/views/courses/show.html.erb
```diff
@@ -73,18 +73,18 @@
...
-        class: "govuk-tabs__tab govuk-tabs__tab--selected", id: "description", role: 'tab', tabindex: '-1', aria: { selected: true } do
+        class: "govuk-tabs__tab govuk-tabs__tab--selected", role: 'tab', tabindex: '-1', aria: { selected: true } do
       %>
         Description<br>
         <span class="govuk-body-s govuk-!-font-weight-regular pointer-events-none">Content, fees and eligibility</span>
       <% end %>
     </li>
 
-    <li class="govuk-tabs__list-item">
+    <li class="govuk-tabs__list-item" role="presentation">
```
fofr: Why is this role needed?

It's not in the design system component, which will have been rigorously tested for accessibility: https://design-system.service.gov.uk/components/tabs/

app/views/courses/show.html.erb
```diff
@@ -101,9 +101,9 @@
         <%= render partial: 'courses/description_content' %>
       </div>
       <% unless course.not_running? %>
-        <div class="govuk-grid-column-one-third" data-qa="course__status_panel">
+        <aside class="govuk-grid-column-one-third" data-qa="course__status_panel">
```
fofr: I don't think we should be switching from `div` to `aside`.

Aside:
> a portion of a document whose content is only indirectly related to the document's main content

I would not say that the content of these side columns is indirectly related to the document. In the case of the status panel it's crucial.

app/views/courses/show.html.erb
```diff
@@ -73,18 +73,18 @@
 <div class="course-tabs govuk-tabs">
   <h2 class="govuk-tabs__title">Contents</h2>
 
-  <ul class="govuk-tabs__list">
+  <ul class="govuk-tabs__list" role="tablist">
```
fofr: I don't think this role is needed, it's not included in the design system, see also: https://simplyaccessible.com/article/danger-aria-tabs/

app/views/courses/show.html.erb
markmead: [I wasn't quite sure if it was necessary but it existed on the sibling `li`](https://github.com/DFE-Digital/manage-courses-frontend/blob/b209a35930b2ea2f499d7ec5db9c99ef70ed948d/app/views/courses/show.html.erb#L77)

fofr: Probably best to remove them all

fofr: Ah, that’s probably because it’s on the currently selected tab, where the link isn't particularly useful.

markmead: So leave there and remove on the other?

fofr:
[Accessibility testing](https://trello.com/c/wSLsEX5E/1764-accessibility-testing)

PR #461, opened 2019-07-16
Certificate can't be created, Time doesn't have method zone

NyWeb:
### Context
Rails is meant to extend Time with zone, but hasn't.

#### My assumption
Rails isn't running as `config/puma.rb` is run.

### Proposed solution:
Since we're just creating a certificate for dev/local, let's use Time.now.

### Guidance to review
None


NyWeb:
[Changing how local certificates are generated](https://trello.com/c/5D8ixl2H/1785-changing-how-local-certificates-are-generated)

defong:
does this mean 

https://github.com/DFE-Digital/manage-courses-backend/blob/b873f1600388dd4a8173100fcfd66ac5d965c8b0/app/models/site_status.rb#L100

is broken?

NyWeb:
Nope, Rails would undoubtedly be running at that time!

PR #466, opened 2019-07-18
[1796] Don't show publish button on rolled over courses

tvararu:
Publishing will trigger a sync to `Find` which will currently fail on the server. Until we fix this, we should display some content instead.

Paired with @fofr.

### After

Screenshot 2019-07-18 at 11 15 27


tvararu:
[Publishing button for rollover](https://trello.com/c/wouXquQ9/1796-publishing-button-for-rollover)

PR #469, opened 2019-07-18
Remove "link for this course" message

fofr:
### Context

This message incorrectly shows for all saved changes, not just publishing.

This means it shows when editing an unpublished draft, or when editing a course in the new cycle, which is confusing.

As a quick fix, remove the message.

### Before
Screen Shot 2019-07-18 at 16 42 15

### After
Screen Shot 2019-07-18 at 16 41 53


misaka:
KILL ITTT!!

PR #472, opened 2019-07-19
[1795] A new_status for a site on a course should be allowed to assign vacancies

defong:
### Context
Misalignment, a `site_status` with `new_status` can be added to a course.

But afterwards the when editing vacancies for the `site_status`, the `site_status` does not show up.

A new_status for a site on a course should show up in order for it to be allowed to assign vacancies

### Guidance to review
A `site_status` with `new_status` regardless of `vac_status` should default to not having any `vac_status` set to allow for the users to set it themselves 

I believe that this is legacy UCAS data.

### DIFF
--- a/spec/helpers/vacancy_helper_spec.rb
+++ b/spec/helpers/vacancy_helper_spec.rb
@@ -1,8 +1,9 @@
 require 'rails_helper'
 
-feature 'Vacancy helpers', type: :helper do
+ffeature 'Vacancy helpers', type: :helper do
   describe '#vacancy_available_for_course_site_status' do
     let(:vacancy_study_mode) { nil }
+    let(:is_new) { false }
 
     subject do
       helper.vacancy_available_for_course_site_status?(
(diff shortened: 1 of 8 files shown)


spec/helpers/vacancy_helper_spec.rb
```diff
@@ -1,8 +1,9 @@
 require 'rails_helper'
 
-feature 'Vacancy helpers', type: :helper do
+ffeature 'Vacancy helpers', type: :helper do
```
fofr: `ffeature` 👀

fofr:
Not sure about this. Makes sense that this is a bug caused by legacy UCAS data.

I assume that for some courses that are running, their are some sites with status running and some that are new.

If the site_status is new, it's not showing up on Find, so what's the value in editing vacancies for it?

fofr:
As discussed on Slack, my take on this is that it's a data clean up exercise.

We should change any site_status that's "new" but on a running course to "suspended".

benilovj:
> As discussed on Slack, my take on this is that it's a data clean up exercise.
> 
> We should change any site_status that's "new" but on a running course to "suspended".

I second this.

defong:
cool net result no longer having to look at frontend :sweat_smile:

PR #473, opened 2019-07-19
[1692] Restrict entry requirement options based on course level

fofr:
### Context

- Lock down the editing options based on the legal rules
- This can only be merged when we've cleaned up the data:
  - All secondary course science requirements should be not_required or not_set
  - All FE requirements should be not_required

- Don't allow any editing of FE course entry requirements
- Don't allow editing of Science entry requirements on Secondary courses

Screen Shot 2019-07-19 at 13 55 04

### DIFF
--- a/app/controllers/courses/entry_requirements_controller.rb
+++ b/app/controllers/courses/entry_requirements_controller.rb
@@ -2,6 +2,7 @@ module Courses
   class EntryRequirementsController < ApplicationController
     decorates_assigned :course
     before_action :build_course
+    before_action :not_found_if_further_education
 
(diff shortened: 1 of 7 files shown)


fofr:
Opening as a discussion point.

benilovj:
I'm not sold on the frontend containing so much of the domain logic. How about:

- the course API returns a list of required equivalency subjects e.g.:

```
required_equivalency_subjects: ['maths', 'english']
```

- the frontend then doesn't need to worry about primary/secondary etc

fofr:
Closed in favour of #473

PR #474, opened 2019-07-19
[1781] Create provider about page

tvararu:
### Context

Provider enrichments in Rails instead of C#.

Updates the provider factory with actual values for `train_with_us` which uncovered a discrepancy in previous specs that requires two more data-qa attrs.

### Guidance to review

🚢


tvararu:
[[S] Create provider 'About your organisation' (move to Rails)](https://trello.com/c/LlvUdxMv/1781-s-create-provider-about-your-organisation-move-to-rails)

PR #478, opened 2019-07-23
[1692] Show only required GCSE subjects, includes validation

fofr:
### Context

A cleaner approach to https://github.com/DFE-Digital/manage-courses-frontend/pull/473.
Lock down the editing options based on the legal rules

- Base all logic on gcse_required_subjects – DFE-Digital/manage-courses-backend#639
- Only show the required subjects for each course when editing entry requirements
- Don't show any entry requirements if there are none needed for the course and make the edit page 404 if the page is requested
- Remove the "No requirement" option, what is shown is always required

When the entry requirements form is submitted without providing an answer to one of the required subjects, show an error.

Doesn’t yet cover the case where an invalid option is passed in. That'll be a separate PR with a backend one.

### Screenshots

localhost_3000_organisations_2FR_2019_courses_3CVZ_entry-requirements

localhost_3000_organisations_2FR_2019_courses_3CW3_entry-requirements

### DIFF
--- a/app/controllers/courses/entry_requirements_controller.rb
+++ b/app/controllers/courses/entry_requirements_controller.rb
@@ -7,6 +7,9 @@ module Courses
     def edit; end
 
     def update
+      @errors = missing_subject_errors
+      return render :edit if @errors.any?
+
       if @course.update(course_params)
         flash[:success] = 'Your changes have been saved'
         redirect_to(
@@ -23,6 +26,13 @@ module Courses
     end
 
   private
+
+    def missing_subject_errors
+      course.gcse_subjects_required
+        .reject { |subject| params.dig(:course, subject) }
+        .map { |subject| [subject.to_sym, ["Pick an option for #{subject.titleize}"]] }
+        .to_h
+    end
 
     def course_params
       params.require(:course).permit(
--- a/app/controllers/courses/entry_requirements_controller.rb
+++ b/app/controllers/courses/entry_requirements_controller.rb
@@ -2,6 +2,7 @@ module Courses
   class EntryRequirementsController < ApplicationController
     decorates_assigned :course
     before_action :build_course
+    before_action :not_found_if_no_gcse_subjects_required
 
     def edit; end
 
@@ -37,6 +38,10 @@ module Courses
         .where(provider_code: params[:provider_code])
         .find(params[:code])
         .first
+    end
+
+    def not_found_if_no_gcse_subjects_required
+      render file: 'errors/not_found', status: :not_found if course.gcse_subjects_required.empty?
     end
   end
 end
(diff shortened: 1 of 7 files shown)


fofr:
[Edit UCAS equivalency](https://trello.com/c/3c0qUpdf/1691-edit-ucas-equivalency)

app/controllers/courses/entry_requirements_controller.rb
```diff
@@ -2,10 +2,14 @@ module Courses
   class EntryRequirementsController < ApplicationController
     decorates_assigned :course
     before_action :build_course
+    before_action :not_found_if_no_gcse_subjects_required
 
     def edit; end
 
     def update
+      @errors = set_missing_subject_errors
+      return render :edit if @errors.any?
```
tvararu: Maybe?

```ruby
return render :edit if missing_subject_errors.any?
```

app/controllers/courses/entry_requirements_controller.rb
```diff
@@ -23,6 +27,17 @@ def update
 
   private
 
+    def set_missing_subject_errors
```
tvararu: `missing_subject_errors` feels nicer to me, nbd if you disagree

app/controllers/courses/entry_requirements_controller.rb
```diff
@@ -23,6 +27,17 @@ def update
...
+      errors = {}
+      course.gcse_subjects_required.each do |subject|
+        if !params.dig(:course, subject)
+          errors[subject] = ["Pick an option for #{subject.titleize}"]
+        end
+      end
+
+      errors
+    end
```
tvararu: This logic can be simplified using `select` + `map`; we filter away all the subjects which do exist in the params, then each subject becomes a tuple of `[subject, "msg"]`. An array of these tuples becomes the hash we want with `.to_h`:

```ruby
def missing_subject_errors
  course.gcse_subjects_required
    .select { |subject| !params.dig(:course, subject) }
    .map { |subject| [subject, "Pick an option for #{subject.titleize}"] }
    .to_h
end
```

defong:
Should the word `Science` vs `science` be case  sensitive?

app/controllers/courses/entry_requirements_controller.rb
tvararu: Nevermind, that wouldn't work at all!

app/controllers/courses/entry_requirements_controller.rb
fofr: Updated.
`[subject, "Pick an option for #{subject.titleize}"]` needed to be `[subject, ["Pick an option for #{subject.titleize}"]]`

fofr:
Can't be deployed until https://github.com/DFE-Digital/manage-courses-backend/pull/639 is deployed.

PR #481, opened 2019-07-25
Use entry requirement options from backend

fofr:
### Context

- Pull in options from backend rather than hardcoding
- Use keys to determine labels and help text

### Guidance to review

Goes alongside https://github.com/DFE-Digital/manage-courses-backend/pull/651


fofr:
Needs https://github.com/DFE-Digital/manage-courses-backend/pull/651 to be deployed before this can get merged.

PR #494, opened 2019-07-30
[1846] Allow editing of course age range in years

dankmitchell:
### Context
Edit course age range

Enable editing of course age range

### Guidance to review
`/organisations/<PROVIDER_CODE>/2019/courses/<COURSE_CODE>/age-range`

### DIFF
--- a/app/controllers/concerns/edit_basic_detail.rb
+++ b/app/controllers/concerns/edit_basic_detail.rb
@@ -12,6 +12,18 @@ module EditBasicDetail
     @errors = errors
     return render :edit if @errors.any?
 
+    # Age range 'other' override
+    course = params.dig(:course)
+    is_other = course.dig(:age_range_in_years) == "Other"
+
+    if course.dig(:course_age_range_in_years_other_from).present? &&
+        course.dig(:course_age_range_in_years_other_to).present? &&
+        is_other
+      params[:course][:age_range_in_years] = "#{course.dig(:course_age_range_in_years_other_from)}_to_#{course.dig(:course_age_range_in_years_other_to)}"
+    elsif is_other
+      params[:course][:age_range_in_years] = nil
+    end
+
     if @course.update(course_params)
       flash[:success] = 'Your changes have been saved'
       redirect_to(
--- a/app/controllers/courses/age_range_controller.rb
+++ b/app/controllers/courses/age_range_controller.rb
@@ -1,0 +1,15 @@
+module Courses
+  class AgeRangeController < ApplicationController
+    include EditBasicDetail
+
+  private
+
+    def errors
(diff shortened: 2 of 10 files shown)


app/controllers/courses/age_range_controller.rb
```diff
@@ -0,0 +1,19 @@
+module Courses
+  class AgeRangeController < ApplicationController
+    include EditBasicDetail
+
+    def edit
+      @options = @course.meta['edit_options']['age_range_in_years'].dup
```
fofr: Do we need to check that `course_age_range_in_years_other_to` and `course_age_range_in_years_other_from` are numbers?

app/controllers/courses/age_range_controller.rb
```diff
@@ -0,0 +1,19 @@
...
+    end
+
+  private
+
+    def errors
+      params.dig(:course, :age_range_in_years) ? {} : { age_range_in_years: ["Pick an age range"] }
```
fofr: Do we need to check that `course_age_range_in_years_other_to` and `course_age_range_in_years_other_from` are numbers?

app/decorators/course_decorator.rb
```diff
@@ -69,6 +69,16 @@ def other_course_length?
     %w[OneYear TwoYears].exclude?(course.course_length)
   end
 
+  def other_age_range?
+    %w[3_to_7
+       5_to_11
+       7_to_11
+       7_to_14
```
fofr: Can this list come from the edit_options?
eg If someone has a primary course but enters "11_to_18" as a custom age range (this might be valid for some SEND courses), we want it to show Other – this logic would give a false negative.

app/controllers/concerns/edit_basic_detail.rb
```diff
@@ -12,6 +12,15 @@ def update
     @errors = errors
     return render :edit if @errors.any?
 
+    # Age range 'other' override
+    if params[:course][:course_age_range_in_years_other_to].present? &&
```
defong: maybe, less boxy

``` ruby
course = params.dig(:course)

is_other = course.dig(:age_range_in_years) == "Other"

if is_other && course.dig(:course_age_range_in_years_other_to) && course.dig(:course_age_range_in_years_other_from)
 params[:course][:age_range_in_years] = "#{course.dig(:course_age_range_in_years_other_to)}_to_#{course.dig(:course_age_range_in_years_other_from)}"    
elsif is_other
  params[:course][:age_range_in_years] = nil

```

app/controllers/courses/age_range_controller.rb
dankmitchell: Validations being done as a separate story

app/decorators/course_decorator.rb
dankmitchell: Good spot. Done

PR #500, opened 2019-08-02
[1886] move validations out of controller

dankmitchell:
### Context
Course age range

- Ensure `age_range_in_years` is present when splitting 
- Add validations for "other"

### Guidance to review
- Try and save the age range without choosing an option

### DIFF
--- a/app/controllers/courses/age_range_controller.rb
+++ b/app/controllers/courses/age_range_controller.rb
@@ -1,15 +1,57 @@
 module Courses
   class AgeRangeController < ApplicationController
-    include EditBasicDetail
+    decorates_assigned :course
+    before_action :build_course
+
+    def update
+      # Age range 'other' override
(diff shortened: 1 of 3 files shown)


app/controllers/courses/age_range_controller.rb
```diff
@@ -1,11 +1,46 @@
...
-    include EditBasicDetail
+    decorates_assigned :course
+    before_action :build_course
+
+    def edit; end
+
+    def update
+      # Age range 'other' overrides
+      course = params.dig(:course)
+      is_other = course.dig(:age_range_in_years) == "Other"
```
fofr: You could put this logic into the `course_params` method in `age_range_controller.rb`, which is what's responsible for  what gets passed to the update action.

Then you could continue using `include EditBasicDetail`.

PR #509, opened 2019-08-08
[1868] show all contact errors on provider editing

markmead:
### Context

Display all contact address errors.

- `"field"` to `"fields"` 
- `fields` is now an array rather than a single string


tvararu:
[Provider editing: Contact address should get highlighted when it has an error](https://trello.com/c/1VKVKoKM/1868-provider-editing-contact-address-should-get-highlighted-when-it-has-an-error)

PR #508, opened 2019-08-08
[1807] Bugfix - hide published-on if never published

timabell:
Was falling over with a nil error for new records.

PR #511, opened 2019-08-08
[1807] Bugfix - hide published-on if never published

timabell:
### Context

Was falling over with a nil error for new records.

(replaces #508 which had a typo in the branch name)

Handle providers with no published enrichment

Without the patch the updated test now fails

```
$ be rspec spec/features/providers/details_spec.rb -fd
2019-08-08 16:11:34 WARN Selenium [DEPRECATION] Selenium::WebDriver::Chrome#driver_path= is deprecated. Use Selenium::WebDriver::Chrom
e::Service#driver_path= instead.
Run options: include {:focus=>true}

All examples were filtered out; ignoring {:focus=>true}

View provider
  with empty provider details
    renders correctly (FAILED - 1)
  with draft provider details
Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.
    renders correctly
  with published provider details
Checking for expected text of nil is confusing and/or pointless since it will always match. Please specify a string or regexp instead.
    renders correctly
```
### Guidance to review

:eye:

### DIFF
--- a/app/views/providers/details.html.erb
+++ b/app/views/providers/details.html.erb
@@ -71,7 +71,9 @@
       <div class="govuk-!-margin-bottom-6" data-qa="provider__content-status">
         <%= provider.status_tag %>
       </div>
-      <p class="govuk-body">Last published on <%= l(@provider&.last_published_at.to_date) %></p>
+      <% if @provider&.last_published_at %>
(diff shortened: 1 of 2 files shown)


timabell:
[[S] accredited body (pull onto Rails), frontend](https://trello.com/c/ATAPQ9c6/1807-s-accredited-body-pull-onto-rails-frontend)

PR #521, opened 2019-08-12
[1929] Switch request access URL from ui to frontend

danielburnley:
### Context

Requesting access has been ported from the old C# application to the new Rails application, but the frontend still links to the old one and needs to be updated.

- Update the request access link to the manage-courses-frontend implementation rather than the C# implementation


tvararu:
I added a do not merge to this but then removed it because I think we can safely ship this now; I've tested access requests just now and they work great in the new frontend.

@defong Nah no need to test this

PR #531, opened 2019-08-15
[1854] Move about your org link to recruitment cycle page

danielburnley:
### Context
Move "About your Organisation" link into the "Current Cycle" or "Next Cycle" page for organisations so that providers can edit the content or information about their organisation in advance of the next cycle without affecting the content or information presented on the current cycle.

- Move "About your organisation" link to recruitment cycle pages
- Add new breadcrumb helper for organisation details page
- Replace `back` link with breadcrumbs
- Fixes "may be confusing" error for details spec

### DIFF
--- a/app/views/recruitment_cycles/show.html.erb
+++ b/app/views/recruitment_cycles/show.html.erb
@@ -18,6 +18,7 @@
     <% end %>
     <hr class="govuk-section-break govuk-section-break--m">
 
+    <%= render partial: 'recruitment_cycles/about_organisation', locals: { provider: @provider, year: params[:year] } %>
     <%= render partial: 'recruitment_cycles/courses_info', locals: { provider: @provider, year: params[:year] } %>
     <%= render partial: 'recruitment_cycles/locations_info', locals: { provider: @provider, year: params[:year] } %>
   </div>
--- a/spec/features/providers/details_spec.rb
+++ b/spec/features/providers/details_spec.rb
@@ -4,7 +4,6 @@ feature 'View provider', type: :feature do
   let(:org_detail_page) { PageObjects::Page::Organisations::OrganisationDetails.new }
 
   before do
-    allow(Settings).to receive(:rollover).and_return(false)
     stub_omniauth
   end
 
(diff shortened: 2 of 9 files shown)


danielburnley:
Wonder if it's best, for now, to hide if it's the next recruitment cycle? If more work needs to be done that is

tvararu:
@danielburnley I think that would be more confusing than the current implementation 🤔 @fofr?

app/views/recruitment_cycles/show.html.erb
```diff
@@ -20,5 +20,6 @@
 
     <%= render partial: 'recruitment_cycles/courses_info', locals: { provider: @provider, year: params[:year] } %>
     <%= render partial: 'recruitment_cycles/locations_info', locals: { provider: @provider, year: params[:year] } %>
+    <%= render partial: 'recruitment_cycles/about_organisation', locals: { provider: @provider, year: params[:year] } %>
```
fofr: Can this go at the top please. That's where it used to live before rollover.

spec/features/providers/details_spec.rb
```diff
@@ -56,8 +55,11 @@ def test_details_page(expected_status)
       provider.to_jsonapi
     )
 
+
```
fofr: Rogue whitespace

fofr:
We should only make/merge this change if all of a provider’s current cycle "About your org" and contact details have already been copied to the next cycle. ie Users won't see blank fields. @danielburnley @misaka

misaka:
We currently have no provider enrichments for the next recruitment cycle:

```
irb(main):005:0> ProviderEnrichment.where(provider: RecruitmentCycle.current_recruitment_cycle.providers).count
=> 2069
irb(main):006:0> ProviderEnrichment.where(provider: RecruitmentCycle.next_recruitment_cycle.providers).count
=> 0
```

misaka:
Seems like we don't have a backend ticket to support this, is that right?

Options:

1. copy-on-demand: we copy the last provider enrichment from the previous cycle. This could be tricky to figure out _when_ to do the copy, is it when we try to serialise the provider (with enrichments) in the backend and discover that they have no enrichments for the requested cycle?
2. data migration: I think we need to make changes to mcbe to pull up the correct cycle's enrichments anyway, so we could also add a migration to ensure that the next cycle's enrichments all have a copy of the latest enrichment from the current cycle

In general, 2 seems like a simpler approach, but one side-effect would be that the next cycle's enrichments would be a snapshot of what the current ones looked like at the time the migration runs. 1, on the other hand, would mean that providers could edit the enrichment for the current cycle and only when they venture into the next cycle would it be pulled in. This might be more intuitive for users, thoughts @fofr ?

fofr:
@misaka I'm happy with the simpler option 2. About your org editing isn't a frequent activity.

dankmitchell:
@fofr can you remove the reject?

PR #533, opened 2019-08-16
[1925] Update logic for applications open for next cycle

tvararu:
### Context

Editing applications open.

The logic was flawed; we should be comparing against the known current
cycle instead of the one being passed in. Updated the test as well.

#### Current cycle

Screenshot 2019-08-16 at 15 19 33

#### Next cycle

Screenshot 2019-08-16 at 15 19 38


tvararu:
[[S] Edit 'Course Applications Open' - (MCFE)](https://trello.com/c/CCq1R1QO/1925-s-edit-course-applications-open-mcfe)

PR #534, opened 2019-08-18
[1938] add new page for course outcomes

misaka:
### Context

So users can set the outcome on their new course.
- renders same form fields as the edit page
- has a "Continue" button (not wired up yet)

Add "new" page for course outcomes.

### Guidance to review

This turned into more work than anticipated. I'm willing to put some time in the evenings to address any comments/questions on this PR, but pls keep in mind I'm trying to be on holiday so I'd greatly appreciate pro-active comments and, as always, encourage additional commits to help  get this over the line. :)

Some notes:
- I've cut up the PR into commits that should stand on their own, and that can (hopefully) be reviewed in order as a kind of narrative.
- This change relies on a `.../courses/new` endpoint on backend to return the possible outcomes for a course. This wasn't called out on the ticket and is the bulk of the extra work that ate away into my holiday time this weekend. :P
- The PR DFE-Digital/manage-courses-backend#723 adds the `.../courses/new` endpoint and is necessary for this change to work.
- Editing the outcome for a new course that doesn't have a level set isn't well defined ... it happens get the "with qts" outcomes displayed, but really this journey doesn't work without having the level set.
- The success criteria `any existing course params should be rendered as hidden input fields, to allow for editing of this param from the confirmation page` was ignored. This was already enough work without adding this bit.

### DIFF
--- a/app/controllers/concerns/course_basic_detail_concern.rb
+++ b/app/controllers/concerns/course_basic_detail_concern.rb
@@ -1,4 +1,4 @@
-module EditBasicDetail
+module CourseBasicDetailConcern
   extend ActiveSupport::Concern
 
   included do
--- a/app/controllers/concerns/course_basic_detail_concern.rb
+++ b/app/controllers/concerns/course_basic_detail_concern.rb
@@ -8,17 +8,10 @@ module CourseBasicDetailConcern
   end
 
   def new
-    # Using .find(:new) here is a little bit hacky, but this is the only way I
-    # could find to construct the URL with `.../courses/new` at the end, and at
-    # the end of the day jsonapi defines ids as being strings so it's in no way
-    # invalid. If we can find a way to use custom enpoints or other to improve
-    # this we should.
-    @course = Course
-                .where(recruitment_cycle_year: params[:recruitment_cycle_year],
-                       provider_code: params[:provider_code])
-                .find(:new)
-                .first
-    nil
+    @course = Course.fetch_new(
+      recruitment_cycle_year: @provider.recruitment_cycle_year,
+      provider_code: @provider.provider_code
+    )
   end
 
   def edit; end
--- a/app/controllers/concerns/course_basic_detail_concern.rb
+++ b/app/controllers/concerns/course_basic_detail_concern.rb
@@ -3,7 +3,22 @@ module CourseBasicDetailConcern
 
   included do
     decorates_assigned :course
-    before_action :build_course
+    before_action :build_provider, only: :new
+    before_action :build_course, only: %i[edit update]
+  end
+
+  def new
+    # Using .find(:new) here is a little bit hacky, but this is the only way I
+    # could find to construct the URL with `.../courses/new` at the end, and at
+    # the end of the day jsonapi defines ids as being strings so it's in no way
+    # invalid. If we can find a way to use custom enpoints or other to improve
+    # this we should.
+    @course = Course
+                .where(recruitment_cycle_year: params[:recruitment_cycle_year],
+                       provider_code: params[:provider_code])
+                .find(:new)
+                .first
+    nil
   end
 
   def edit; end
@@ -28,6 +43,13 @@ module CourseBasicDetailConcern
   end
 
 private
+
+  def build_provider
+    @provider = Provider
+                  .where(recruitment_cycle_year: params[:recruitment_cycle_year])
+                  .find(params[:provider_code])
+                  .first
+  end
 
   def build_course
     @course = Course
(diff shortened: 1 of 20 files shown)


misaka:
[[M] Add "new" page for course outcomes](https://trello.com/c/4HGV6r3S/1938-m-add-new-page-for-course-outcomes)

app/views/courses/outcome/new.html.erb
```diff
@@ -0,0 +1,32 @@
...
+                  ),
+                  method: :put do |form| %>
+
+      <%= render 'form', form: form, course_name_and_code: nil %>
+
+      <%= form.submit "Continue",
+                      class: "govuk-button", data: { qa: 'course__save' } %>
+
+      <p class="govuk-body">
+        <%= link_to 'Cancel changes',
```
fofr: No need for a cancel link in the new course wizard.

spec/support/helpers.rb
```diff
@@ -1,4 +1,31 @@
 module Helpers
+  def url_for_resource_collection(resource)
```
danielburnley: I found these methods quite cumbersome to parse through in my head when reading them due to the number of things each one can do based on the resource passed in. 

IMO I'd find these easier to read/find how to use if they were more explicit and less generic, e.g. `url_for_recruitment_cycle_collection` and `url_for_provider_collection`

danielburnley: ah I see further down there's more methods :thinking: okay i can see why having these generic for now works

spec/features/courses/vacancies/edit_spec.rb
```diff
@@ -1,9 +1,11 @@
+# coding: utf-8
```
danielburnley: :eyes: should this be in there?

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -0,0 +1,61 @@
...
+    before_action :build_course, only: %i[edit update]
+  end
+
+  def new
+    # Using .find(:new) here is a little bit hacky, but this is the only way I
+    # could find to construct the URL with `.../courses/new` at the end, and at
+    # the end of the day jsonapi defines ids as being strings so it's in no way
+    # invalid. If we can find a way to use custom enpoints or other to improve
+    # this we should.
+    @course = Course
```
danielburnley: There might be some value in refactoring the `Course.where...` out into a method that explains what it's doing a bit more clearly imo

spec/features/courses/vacancies/edit_spec.rb
misaka: Ya, so this is related to [my comment in the other PR](https://github.com/DFE-Digital/manage-courses-backend/pull/723#discussion_r315004676) about replacing apostrophes with single-quotes. Emacs has detected UTF-8 characters and this is the standard way to signal that a Ruby source file is UTF-8 encoded. There is the argument that Ruby 2.x assumes UTF-8, but it doesn't hurt here and I didn't feel like hunting down the UTF-8 character.

spec/support/helpers.rb
misaka: That's a fair point. I've moved them into `private`, for now at least, since they're not really meant for external use. Maybe it'll better explain the intention that way.

app/controllers/concerns/course_basic_detail_concern.rb
misaka: Ya, I think you're right, as evidenced by the explanatory note here. I've moved it to `Course.fetch_new`.

app/models/course.rb
```diff
@@ -11,6 +11,19 @@ class Course < Base
 
   self.primary_key = :course_code
 
+  # Fetch a "new" record from the backend API.
+  def self.fetch_new(recruitment_cycle_year:, provider_code:)
+    # Using .find(:new) here is a little bit hacky, but this is the only way I
+    # could find to construct the URL with `.../courses/new` at the end, and at
+    # the end of the day jsonapi defines ids as being strings so it's in no way
+    # invalid. If we can find a way to use custom enpoints or other to improve
```
defong:
```suggestion
    # invalid. If we can find a way to use custom endpoints or other to improve
```

spec/features/courses/new_spec.rb
```diff
@@ -0,0 +1,38 @@
+require "rails_helper"
```
defong: shouldn't this file sit with 

spec/features/courses/vacancies/edit_spec.rb

instead of 

spec/features/courses/new_spec.rb

misaka: @defong This feature spec is for the course creation overall ... these steps should be tested as whole journey through the create course, and we'll probably have multiple feature specs for course creation to test out different scenarios.

The request spec exercises this functionality in isolation.

PR #536, opened 2019-08-19
Added prod.yml file

rizzkhan1:
Context
To ensure that after migration all micro service urls are pointing to service domain.
new_prod.yml and production.yml can be deleted after successful migration to vsts.

Changes proposed in this pull request
Created prod.yml file with settings copied from production.yml but with service domain instead of education domain url.

### DIFF
--- a/config/settings/prod.yml
+++ b/config/settings/prod.yml
@@ -1,0 +1,14 @@
+dfe_signin:
+  issuer: https://oidc.signin.education.gov.uk
+  secret: please_change_me # Override with SETTINGS__DFE_SIGNIN__SECRET
+  profile: https://profile.signin.education.gov.uk
+  base_url: https://www2.publish-teacher-training-courses.service.gov.uk
+manage_backend:
+  base_url: https://api2.publish-teacher-training-courses.service.gov.uk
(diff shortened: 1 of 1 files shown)


config/settings/prod.yml
```diff
@@ -0,0 +1,14 @@
...
+  base_url: https://www2.publish-teacher-training-courses.service.gov.uk
+manage_backend:
+  base_url: https://api2.publish-teacher-training-courses.service.gov.uk
+manage_ui:
+  base_url: https://www.publish-teacher-training-courses.service.gov.uk
+search_ui:
+  base_url: https://www.find-postgraduate-teacher-training.service.gov.uk
+environment:
+  label: "Beta"
+  selector_name: "beta"
```
defong: https://github.com/DFE-Digital/manage-courses-frontend/blob/151d12ed7aa4918aef1c87e6ba6bfb2cab48d5d6/config/settings/new_prod.yml#L14

does casing matters?

rizzkhan1: Don't know if case matters. But it should work as is because I copied it from production.yml file which is currently in use.

PR #537, opened 2019-08-19
[1983] Add unauthorized route, action and page

dankmitchell:
### Context
Unauthorized users

Add `/401` page and route as per manage courses ui

### Guidance to review
🚢

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -42,7 +42,7 @@ class SessionsController < ApplicationController
   end
 
   def failure
-    redirect_to "/401"
+    redirect_to unauthorized_path
   end
 
   def destroy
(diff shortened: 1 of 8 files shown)


dankmitchell:
[Port "We don’t know which organisation you’re part of" to Rails](https://trello.com/c/Uqtuv3GA/1983-port-we-dont-know-which-organisation-youre-part-of-to-rails)

defong:
image

app/controllers/sessions_controller.rb
```diff
@@ -42,7 +42,7 @@ def signout
   end
 
   def failure
-    redirect_to "/401"
+    redirect_to unauthorized_path
```
fofr: Should this be redirecting or just responding with a 401?

dankmitchell: This is the frontend so yes, we should redirect.

fofr: We respond with a 404 and show page not found for that URL, we don't redirect to `/404`.
Why isn't this the same?

PR #543, opened 2019-08-21
[2007] Remove single location special case

fofr:
For some of our courses there is bad data – some don't have any locations. We can't easily set locations for these course.

We were assuming that for providers with 1 location, all courses would have that location. That's not the case.

Most of the affected courses weren't running in 2019.

Make it possible for providers to self-serve and fix their courses by going in and adding the missing location.

Move guidance about managing locations to the Pick your locations page and make it available to all providers.

Supersedes #540.

Screen Shot 2019-08-21 at 15 55 49
Screen Shot 2019-08-21 at 15 55 57

### DIFF
--- a/app/views/courses/_basic_details_tab.html.erb
+++ b/app/views/courses/_basic_details_tab.html.erb
@@ -106,7 +106,9 @@
           Locations
         </dt>
         <dd class="govuk-summary-list__value" data-qa="course__locations">
-          <% if course.sites.size == 1 %>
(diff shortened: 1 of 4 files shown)


fofr:
[Fix issue with rolled over 2020 courses with 0 `site_statuses`](https://trello.com/c/WiXnwzp8/2007-fix-issue-with-rolled-over-2020-courses-with-0-sitestatuses)

PR #559, opened 2019-08-29
[1999] Add edit apprenticeship page

danielburnley:
### Context

To enable users to edit whether or not their course is an apprenticeship

- Adds in the ability to edit the program type on the frontend, specifically for apprenticeships
  - Supports both SCITT and non-scitt providers

### Guidance to review
- Check out this branch and https://github.com/DFE-Digital/manage-courses-backend/pull/743/ on the backend
- Go to a provider that can offer an apprenticeship

### DIFF
--- a/app/controllers/courses/apprenticeship_controller.rb
+++ b/app/controllers/courses/apprenticeship_controller.rb
@@ -1,0 +1,13 @@
+module Courses
+  class ApprenticeshipController < ApplicationController
+    include CourseBasicDetailConcern
+
+  private
+
+    def errors; end
+
+    def course_params
+      params.require(:course).permit(:program_type)
+    end
+  end
+end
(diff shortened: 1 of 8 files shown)


danielburnley:
We can't really merge this until https://github.com/DFE-Digital/manage-courses-backend/pull/743 has been merged :eyes:

tvararu:
@danielburnley we can if we temporarily hide the change link, I would encourage that as it also means we don't have dependant deploys which are annoying!

Example follow-up PR to enable the feature when it's ready: https://github.com/DFE-Digital/manage-courses-frontend/pull/535/files

tvararu:
Discussed offline in the context of previous suggestion to disable this:

We could benefit from adding a mechanism to manage feature flags. It could be a combination of using the existing `Settings` files + a flow in MCB to enable them on a per-app and per-enrivonment basis. (this is to be prioritised separately, not as part of this)

PR #579, opened 2019-09-02
2064 add new page for fee paying or salaried

defong:
### Context
New courses

Add "fee or salary" selection for new course flow

### Guidance to review
🚢 

image

### DIFF
--- a/app/views/courses/fee_or_salary/_form_fields.html.erb
+++ b/app/views/courses/fee_or_salary/_form_fields.html.erb
@@ -1,0 +1,27 @@
+<div
+  class="<%= cns("govuk-form-group", "govuk-form-group--error": @errors && @errors[:program_type]&.any?) %>"
+  data-qa="course__program_type">
+
+  <fieldset class="govuk-fieldset">
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+      <h1 class="govuk-fieldset__heading">
+        Is it fee paying or salaried?
+      </h1>
+    </legend>
+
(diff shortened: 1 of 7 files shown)


defong:
[Add "new" page for Fee paying or salaried](https://trello.com/c/LdiuA9I4/2064-add-new-page-for-fee-paying-or-salaried)

app/views/courses/fee_or_salary/_form_fields.html.erb
```diff
@@ -0,0 +1,28 @@
...
+  data-qa="course__program_type">
+
+  <fieldset class="govuk-fieldset">
+    <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+      <h1 class="govuk-fieldset__heading">
+        Is it fee paying or salaried?
+      </h1>
+    </legend>
+
+    <%= render "shared/error_messages", field: :program_type if @errors %>
```
misaka: I had another look at this tonight, with @faissalMT, and after the chat with @dankmitchell, and I'm starting to think that this is in fact the wrong shape. Translating the `program_type` to whether the course is fee paying or salaried should be in the backend, not here, and the API should reflect that. I think we need to have another look at this.

defong: I am also doubt that exposing the current course model is the right approach. It feels like another in direction should be in place. May be a facade, that is the final replacement for the current course model to be persisted or an adaptor.

defong: @davidgisbey fyi

PR #587, opened 2019-09-04
[2036] Fix webpack dev server https key paths

defong:
### Context

This fixes the slow running tests we're seeing in CI.

### Guidance to review

https://i.redd.it/4ypiuonxsik31.png
image

### DIFF
--- a/config/webpack/development.js
+++ b/config/webpack/development.js
@@ -5,8 +5,8 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 const environment = require('./environment')
 
 environment.config.devServer.https = {
-  key: fs.readFileSync('./config/localhost/https/localhost.key'),
-  cert: fs.readFileSync('./config/localhost/https/localhost.crt')
+  key: './config/localhost/https/localhost.key',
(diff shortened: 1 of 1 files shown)


config/webpack/development.js
```diff
@@ -5,8 +5,8 @@ process.env.NODE_ENV = process.env.NODE_ENV || 'development'
 const environment = require('./environment')
 
 environment.config.devServer.https = {
-  key: fs.readFileSync('./config/localhost/https/localhost.key'),
-  cert: fs.readFileSync('./config/localhost/https/localhost.crt')
+  key: './config/localhost/https/localhost.key',
+  cert: './config/localhost/https/localhost.crt'
```
misaka: Does this fix the slow tests? I thought the problem is that they don't exist (they have to be generated), not that they're in the wrong place.

defong: > Does this fix the slow tests? 30 secs
image

> I thought the problem is that they don't exist (they have to be generated), not that they're in the wrong place.

@tvararu can tell you the nightmare

that dot in `.`/config/localhost/https/localhost.key is not what you think it is...

PR #588, opened 2019-09-04
create localhost.crt and localhost.key in tests

misaka:
### Context

Without this webpacker continuously tries to compile assets, but fails because
they require this cert, slowing down tests massively (~20s vs 20mins). The
alternative solution is to remove this requirement, but how to implement this
solution was more obvious to me.

Move cert generation into an initialiser that runs in development and test so that it's present for tests.

### Guidance to review

#587 is an alternative solution, I think, which simply sidesteps the reading of the ssl cert files. Might be the way to go.

### DIFF
--- a/config/initializers/ssl.rb
+++ b/config/initializers/ssl.rb
@@ -1,0 +1,23 @@
+if Rails.env.in? %w[development test]
+  localhost_path = File.join(Rails.root, 'config', 'localhost', 'https')
+  cert = File.join(localhost_path, 'localhost.crt')
+  key = File.join(localhost_path, 'localhost.key')
+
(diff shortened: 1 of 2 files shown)


config/initializers/ssl.rb
```diff
@@ -0,0 +1,23 @@
+if Rails.env.in? %w[development test]
+  localhost_path = File.join(Rails.root, 'config', 'localhost', 'https')
```
defong: disconnect with the puma path

tvararu:
Fixed in https://github.com/DFE-Digital/manage-courses-frontend/pull/587

PR #593, opened 2019-09-05
[2053] Add nojs provider search

faissalMT:
### Context
As part of adding provider suggestions and searching we'd like to search for providers from the frontend.  

Add a nojs provider search.  

#### New accredited bodies page

Screenshot 2019-09-16 at 12 54 45

#### Selecting the last option
Screenshot 2019-09-16 at 12 54 50

#### Selecting from the results

Screenshot 2019-09-16 at 12 55 01

#### No results

Screenshot 2019-09-16 at 12 55 11

#### Validation

Screenshot 2019-09-16 at 12 56 30

### DIFF
--- a/app/models/provider_suggestion.rb
+++ b/app/models/provider_suggestion.rb
@@ -1,0 +1,7 @@
+class ProviderSuggestion < Base
+  def self.suggest(query)
+    self.requestor.__send__(
+      :request, :get, "/api/v2/providers/suggest?query=#{query}"
+    )
+  end
+end
--- a/app/views/courses/accredited_body/search.html.erb
+++ b/app/views/courses/accredited_body/search.html.erb
@@ -1,0 +1,72 @@
+<%= content_for :page_title, "Organisation search" %>
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(
+    details_provider_recruitment_cycle_course_path(
+      course.provider_code,
+      course.recruitment_cycle_year,
+      course.course_code)) %>
+<% end %>
+
+<%= render 'shared/errors' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">
+      Pick an accredited body
+    </h1>
+
+    <p class="govuk-body">You searched for ‘<%= @query %>’.</p>
+
+    <%= form_with model: course,
+      url: accredited_body_provider_recruitment_cycle_course_path(
+        params[:provider_code],
+        params[:recruitment_cycle_year],
+        params[:code]),
+      method: :put do |form| %>
+
+      <% if @provider_suggestions.any? %>
+        <p class="govuk-body">We found these providers which matched your search:</p>
+
+        <div class="govuk-radios" data-module="govuk-radios" data-qa="course__accredited_body">
+          <%= render partial: "provider_suggestion", collection: @provider_suggestions, locals: { form: form }%>
+          <div class="govuk-radios__divider">or</div>
+
+          <div class="govuk-radios__item">
+            <%= form.radio_button :accrediting_provider_code,
+              'other',
+              class: 'govuk-radios__input',
+              data: { "aria-controls" => "other-container" } %>
+            <%= form.label :accrediting_provider_code,
+              "Try another accrediting provider",
+              for: 'course_accrediting_provider_code_other',
+              value: false,
+              class: 'govuk-label govuk-radios__label' %>
+          </div>
+
+          <div class="govuk-radios__conditional govuk-radios__conditional--hidden" id="other-container">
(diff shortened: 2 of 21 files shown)


faissalMT:
[[M] Allow users to search for accredited providers](https://trello.com/c/baTzgPFt/2053-m-allow-users-to-search-for-accredited-providers)

app/models/provider_suggestion.rb
```diff
@@ -0,0 +1,7 @@
+class ProviderSuggestion < Base
```
dankmitchell: Why a new model?

app/views/providers/search.html.erb
```diff
@@ -0,0 +1,3 @@
+<% ProviderSuggestion.suggest(params[:query]).each do |suggestion| %>
```
dankmitchell: This logic belongs in the controller and should be exposed via an instance variable

app/views/providers/search.html.erb
```diff
@@ -0,0 +1,7 @@
+<%= content_for :page_title, "Organisation search" %>
+
+<h1 class="govuk-heading-xl">Pick a provider</h1>
+
+<ul class="govuk-list">
+  <%= render partial: "providers/provider_suggestion", collection: ProviderSuggestion.suggest(params[:query]) %>
```
tvararu: How many items are returned by the backend here? We should limit it to 5.

app/models/provider_suggestion.rb
faissalMT: I've added this separation in the backend too because I believe these are distinct domains. While a provider is persistent, a suggestion is transient and need only communicate a small amount of information, not only does does this make things easier to work with, it'll also be alot faster.

app/views/providers/search.html.erb
faissalMT: Ah you're right I did forget to add a limit on the backend.

app/models/provider_suggestion.rb
timabell: The spec allows returning subsets of fields if that's what you are after https://jsonapi.org/format/#fetching-sparse-fieldsets

faissalMT: I did consider this at one point but I feel like this implementation has better extensibility, we may for example want to add ordering data to a suggestion that wouldn't be needed as part of the provider model or some other such thing. If we're tied to the provider model we have to end up making the model even bigger than it already is.

app/views/courses/accredited_body/search.html.erb
```diff
@@ -0,0 +1,34 @@
...
+  <% else %>
+    <%= govuk_back_link_to(providers_path) %>
+  <% end %>
+<% end %>
+
+<h1 class="govuk-heading-xl">Pick a provider</h1>
+
+<% if @error %>
+  <div>
+  Query too short
```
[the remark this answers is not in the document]
defong: name or code right?

app/views/courses/accredited_body/search.html.erb
```diff
@@ -0,0 +1,34 @@
...
+        </div>
+      </div>
+      <%= form.submit course.is_running? ? "Save and publish changes" : "Save",
+        class: "govuk-button govuk-!-margin-top-3", data: { qa: 'course__save' } %>
+  <% end %>
+<% else %>
+  <div>
+    No results
+  <div>
+<% end %>
```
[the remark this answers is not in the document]
defong: `No results` wrong font?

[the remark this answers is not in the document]
tvararu: Lowered limit: https://github.com/DFE-Digital/manage-courses-backend/pull/799

app/views/shared/_error_messages.html.erb
```diff
@@ -1,4 +1,4 @@
-<% if @errors[field]&.any? %>
+<% if @errors && @errors[field]&.any? %>
```
faissalMT: Maybe `if @errors&.[](field)&.any?`?

spec/features/courses/accredited_body/edit_spec.rb
```diff
@@ -69,6 +78,54 @@
...
+          expect(accredited_body_search).to have_content("We did not find any")
+        end
+
+        scenario 'shows no providers' do
+          expect(accredited_body_search.accredited_body_options.length).to eq(0)
+        end
+      end
+
+      context "validations" do
+      end
```
faissalMT: Should this be here?

app/views/shared/_error_messages.html.erb
tvararu: I don't prefer it but I've switched to it

fofr:
Please can you add some screenshots?

tvararu:
@fofr done

PR #595, opened 2019-09-05
[2036] Redirect www2 requests to www

tvararu:
### Context

Once DfE Signin have whitelisted the domain on the `bats2` identifier, and the CNAME is in place with Rails serving both `www2` and `www`, we can release this change to flip all users back to `www`.

Flips the redirect around and updates the signin `base_url`s.

### Guidance to review

This will be tested in both QA and Staging before going live.

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -1,6 +1,6 @@
 # rubocop:disable Metrics/BlockLength
 Rails.application.routes.draw do
-  constraints(host: /www\./) do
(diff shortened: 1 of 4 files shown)


tvararu:
[Decommission C# version of publish i.e. manage courses ui](https://trello.com/c/QD23rZvI/2036-decommission-c-version-of-publish-ie-manage-courses-ui)

PR #609, opened 2019-09-11
[2117] Add accept terms page

tvararu:
### Context

Users who log in and have not accepted terms and conditions get a "We don't know which org you're part of" page.

Implement the terms page.

### Guidance to review

🚢


tvararu:
[Providers still getting the  'we don't know which organisation' error when they have been given access to an organisation](https://trello.com/c/S2BFtm1Z/2117-providers-still-getting-the-we-dont-know-which-organisation-error-when-they-have-been-given-access-to-an-organisation)

PR #630, opened 2019-09-18
[2159] Use funding_type instead of program_type

faissalMT:
### Context
This is the frontend counterpart to https://github.com/DFE-Digital/manage-courses-backend/pull/787, we want to switch away from feeding `program_type` and then using a translations file to determine how to present it to the user.

Change instances of `funding` to `funding_type`, also use new `funding_type` values. Also enable strict parameters.

### Guidance to review


faissalMT:
[Make frontend use funding_type](https://trello.com/c/9bE0fmWG/2159-make-frontend-use-fundingtype)

PR #632, opened 2019-09-19
[2165] funding type cant be changed post publishing

davidgisbey:
### Context

Users can currently change whether their course is an apprenticeship from yes to no/or fee/salary after publishing. 

This is unfair on applicants who may apply based on the previous state of the course.

- Once published a course cannot change it's fee structure.


davidgisbey:
[Fee or salary should not be editable if course is published](https://trello.com/c/SurphTMl/2166-fee-or-salary-should-not-be-editable-if-course-is-published)

PR #638, opened 2019-09-23
[2026] Hook up course creation

faissalMT:
### Context

To lay out the structure of how we hook up the `continue` methods to controllers - this adds a test which hooks up all the currently available steps for `SCITTs` with a `Primary` course

- Add a service for calculating the next step, given the current step
  - The motivation behind this is so that it can be expanded by passing in whatever other data is needed e.g. `provider_type` or `level`
- Create a test which asserts the flow is correct for this type of course
- Add in test helpers for course creation tests

### DIFF
--- a/app/controllers/concerns/course_basic_detail_concern.rb
+++ b/app/controllers/concerns/course_basic_detail_concern.rb
@@ -58,30 +58,39 @@ private
 
   def course_params
     if params.key? :course
-      params.require(:course).permit(
-        :page,
-        :about_course,
-        :course_length,
-        :course_length_other_length,
-        :fee_details,
-        :fee_international,
-        :fee_uk_eu,
-        :financial_support,
-        :how_school_placements_work,
-        :interview_process,
-        :other_requirements,
-        :personal_qualities,
-        :salary_details,
-        :required_qualifications,
-        :qualification, # qualification is actually "outcome"
-        :maths,
-        :english,
-        :science,
-        :funding_type,
-        :level,
-        :is_send,
-        :program_type
-      )
+      params.require(:course)
+        .except(
+          :day,
+          :month,
+          :year,
+        )
+        .permit(
+          :page,
+          :about_course,
+          :course_length,
+          :course_length_other_length,
+          :fee_details,
+          :fee_international,
+          :fee_uk_eu,
+          :financial_support,
+          :how_school_placements_work,
+          :interview_process,
+          :other_requirements,
+          :personal_qualities,
+          :salary_details,
+          :required_qualifications,
+          :qualification, # qualification is actually "outcome"
+          :maths,
+          :english,
+          :science,
+          :funding_type,
+          :level,
+          :is_send,
+          :program_type,
+          :study_mode,
+          :applications_open_from,
+          :start_date,
+        )
     else
       ActionController::Parameters.new({}).permit(:course)
     end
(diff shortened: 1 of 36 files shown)


faissalMT:
[[S] Hook up "continue" action for all new course steps](https://trello.com/c/IzgPofb9/2026-s-hook-up-continue-action-for-all-new-course-steps)

spec/features/courses/new_spec.rb
```diff
@@ -89,7 +89,14 @@
...
         course_creation_params = select_applications_open_from(course_creation_params)
-        course_creation_params = select_start_date(course_creation_params)
+
+        #Addressable, the gem site-prism relies on, cannot match parameters containing a +
+        #https://github.com/sporkmonger/addressable/issues/142
+        # Addressable::Template.new('/a{?query*}').match(Addressable::URI.parse('/a?a=b+b')) == false
+        # Addressable::Template.new('/a{?query*}').match(Addressable::URI.parse('/a?a=b')) == true
+        # Ergo the following test, which sets a value to September+2019 will not work
+
+        #course_creation_params = select_start_date(course_creation_params)
```
misaka: 🤦‍♂

danielburnley:
This needs to be extended with the age range page, doing that now

app/controllers/courses/age_range_controller.rb
```diff
@@ -1,17 +1,11 @@
...
-
-    def new; end
 
     def continue
-      redirect_to confirmation_provider_recruitment_cycle_courses_path(
-        params[:provider_code],
-        params[:recruitment_cycle_year],
-        course_params,
-      )
+      redirect_to next_step(current_step: :age_range)
```
misaka: The other `continue` actions render out errors. Should this one too?

app/controllers/courses/applications_open_controller.rb
```diff
@@ -22,23 +21,35 @@ def continue
...
+        .permit(
+          :applications_open_from,
+          :day,
+          :month,
+          :year,
+        )
     end
 
-    def course_params
+    def build_course_params
```
misaka: It could be the lack of sleep, but I had to bend my mind backwards to understand what this was all about. Might be worth a short comment to help explain it's about the dates optionally being specified as year/month/day.

danielburnley: Yeah i can see that - i did wonder how to name this, comment will probably help a good amount though

app/controllers/courses/age_range_controller.rb
danielburnley: The error handling across all the continues need a bit of work, I can make this look the same for now but none of them really connect to anything yet (didn't want to make this PR bigger)

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -89,18 +104,63 @@ def build_previous_course_creation_params
...
+        params[:provider_code],
+        params[:recruitment_cycle_year],
+        course: course_params,
+      ),
+      confirmation: confirmation_provider_recruitment_cycle_courses_path(
+        params[:provider_code],
+        params[:recruitment_cycle_year],
+        course: course_params,
+      ),
+    }
```
misaka: I have this niggling feeling about this. Basically we're generating N number of paths for this method when we call it. I have a feeling that generating paths is slow, and generating a bunch of these when we don't need them is maybe worth avoiding. Looking at benchmarks, it's not the worst thing in the world, but cumulatively this could add up to a slow service. Running this 100 times takes 0.33s, vs 0.0002s for NextCourseCreationStepService.new.execute.

There's two ways we could improve this:
- create a method per path, but then we'd have to do `__send__("path_for_#{next_step}")`
- use a case when in here, which leads to a slightly long method (I like short methods)

danielburnley: (copying from slack)

Wonder if the quickest win would be changing it from a hash to a case statement, then it's not gonna generate all the paths. Will investigate

misaka: I'd be happy with that.

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -89,18 +104,73 @@ def build_previous_course_creation_params
...
+
+  def course_creation_path_for(page)
+    case page
+    when :apprenticeship
+      new_provider_recruitment_cycle_courses_apprenticeship_path(
+        params[:provider_code],
+        params[:recruitment_cycle_year],
+        course: course_params,
+      )
+      # Currently the page isnt built - so skip
```
misaka: What does this comment mean, btw?

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -89,18 +104,73 @@ def build_previous_course_creation_params
...
+        params[:provider_code],
+        params[:recruitment_cycle_year],
+        course: course_params,
+      )
+    when :confirmation
+      confirmation_provider_recruitment_cycle_courses_path(
+        params[:provider_code],
+        params[:recruitment_cycle_year],
+        course: course_params,
+      )
```
misaka: I was just looking into Code Climate's complaint about this method being too long. It turns out we can cut this down by quite a bit, actually:

```
  def course_creation_path_for(page)
    path_params = params.slice(:provider_code, :recruitment_cycle_year)
                    .merge(course: course_params)
    case page
    when :apprenticeship
      new_provider_recruitment_cycle_courses_apprenticeship_path(path_params)
    when :location
      new_provider_recruitment_cycle_courses_entry_requirements_path(path_params)
    ...
```

This has the benefit in that it's easier to reason about the params a little bit ... before I found myself wondering if they were all the same, etc.

Also, it seems we don't even need `provider_code` and `recruitment_cycle_year`, from my testing, so I _think_ this would be fine too:

```
    path_params = { course: course_params }
```

Just some thoughts. I'm happy to worry about this later if you just want to get it merged.

app/controllers/concerns/course_basic_detail_concern.rb
danielburnley: I should expand that a bit - currently the `sites/new` page doesn't exist - so we skip that step to allow it to get to the confirmation page as is

app/controllers/concerns/course_basic_detail_concern.rb
danielburnley: Huh - I thought I tested it without the `provider_code` and `recruitment_cycle_year` but apparently I didn't. That works well, i'll make the changes and commit :tada:

misaka: :tada: All green! :shipit:  🚀

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -89,18 +104,40 @@ def build_previous_course_creation_params
...
+      new_provider_recruitment_cycle_courses_apprenticeship_path(path_params)
+    # Currently the location page isnt built - so we skip that step
+    # and go to the entry_requirements for now
+    when :location
+      new_provider_recruitment_cycle_courses_entry_requirements_path(path_params)
+    when :entry_requirements
+      new_provider_recruitment_cycle_courses_entry_requirements_path(path_params)
+    when :outcome
+      new_provider_recruitment_cycle_courses_outcome_path(path_params)
+    when :full_or_part_time
```
defong: just fussy... 

can we not just called this `study_mode`

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -89,18 +104,40 @@ def build_previous_course_creation_params
...
+    next_page = course_creation_path_for(next_step)
+
+    if next_page.nil?
+      raise "No path defined for next step: #{next_step}"
+    end
+
+    next_page
+  end
+
+  def course_creation_path_for(page)
```
defong: can this be collapsed together...
they both got the giant case statement

service takes current and return next, and method takes next (page) and returns path

app/controllers/concerns/course_basic_detail_concern.rb
danielburnley: I tried to keep it in line with what the URLs were as opposed to the name of the field (e.g. Apprenticeship over Program Type)

misaka: I agree, we may change `study_mode` in the backend, and we shouldn't have to change this as a result, I'd say.

PR #647, opened 2019-09-24
[2195] add debug logging for user auth

misaka:
### Context

We're trying to chase down a particularly thorny bug which is hard to replicate locally that seems to be related to the auth process.

Build ontop of the recent logstash additions to add optional debug-level output to gain insights of what's happening during a user's authentication process. (i.e. a bunch of debug logging)

### Guidance to review

The location of the debug output has been "carefully" selected to try to make sense in terms of the login / auth flow. If you see anywhere else that might be especially useful for gaining insight, without flooding the logs with noise let me know.

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,6 +29,8 @@ class ApplicationController < ActionController::Base
 
   def authenticate
     if current_user.present?
+      logger.debug("Authenticated user session found " + current_user.to_s)
+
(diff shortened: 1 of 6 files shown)


timabell:
[3x users incorrectly unable to get past "unknown org" page](https://trello.com/c/OhXYtbKP/2195-3x-users-incorrectly-unable-to-get-past-unknown-org-page)

PR #649, opened 2019-09-25
[2195] Strip emails from debug logs

timabell:
### Guidance to review


timabell:
[3x users incorrectly unable to get past "unknown org" page](https://trello.com/c/OhXYtbKP/2195-3x-users-incorrectly-unable-to-get-past-unknown-org-page)

PR #652, opened 2019-09-26
[2224] Fix NoMethod Key for 403 handling

danielburnley:
### Context

`NoMethodError: undefined method `key?' for "":String` Was being thrown when getting `403` responses from the API due to incorrect assertions about the shape of the response in this situation.

- Check that the response is a hash before checking for metadata


danielburnley:
https://trello.com/c/TJbYE247/2224-undefined-method-key-for-string-nomethoderror

PR #656, opened 2019-09-27
Keep 2020 cycle in next cycle mode for now

fofr:


fofr:
Closing for now because of test fails

PR #639, opened 2019-09-23
[2206] Iterate UI for new cycle

dankmitchell:
### Context
New recruitment cycle

### Changes proposed in this pull request
- Add next recruitment_cycle
- Refactor org#show page

### Guidance to review
🚢 

# After
localhost_3000_organisations_2AT(Laptop with MDPI screen) (1)

PR #657, opened 2019-09-27
Reimplement #639 without previous cycle work

fofr:
### Context

An alternative approach to #639

- Update all tests to be cycle year and rollover agnostic (eg tests pass with 2019 or 2020, rollover true or false)
- Show new content on organisation page
- Add a flag for if the current cycle has opened yet
- Keeps rollover as a modal feature we can toggle on and off
- We will add previous cycle logic and code when it's needed

### Guidance to review

- Best to review commit by commit, the diff is a little complex because of the revert

#### Rollover false
localhost_3000_organisations_2FR (1)

Screen Shot 2019-09-27 at 13 59 13

#### Rollover true (ie next summer)
localhost_3000_organisations_2C4

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -126,8 +126,7 @@ class CourseDecorator < ApplicationDecorator
   end
 
   def applications_open_from_message_for(recruitment_cycle)
-    is_current_recruitment_cycle = recruitment_cycle.year == Settings.current_cycle.to_s
-    if is_current_recruitment_cycle
+    if current_cycle?
       "As soon as the course is on Find (recommended)"
     else
       year = recruitment_cycle.year.to_i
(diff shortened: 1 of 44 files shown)


fofr:
Closing while I fix a subtle bug.

app/decorators/course_decorator.rb
```diff
@@ -13,9 +13,9 @@ def vacancies
 
   def on_find(provider = object.provider)
     if object.findable?
-      if current_cycle?
+      if current_cycle_and_open?
         h.govuk_link_to("Yes - view online", h.search_ui_course_page_url(provider_code: provider.provider_code, course_code: object.course_code))
-      elsif next_cycle?
+      else
         "Yes – from October"
```
defong: do we need to give 1 or 8 or just nothing?

defong: image

fofr: from October is fine I think.

PR #667, opened 2019-10-01
[2267] Temporarily remove course counts from organisations page

fofr:
Remove counts to avoid confusing users with incorrect numbers.

We should fix this properly in the backend:
https://trello.com/c/9EcGLugq/2267-course-counts-on-organisation-list-show-the-wrong-number-of-courses


fofr:
[Course counts on organisation list show the wrong number of courses](https://trello.com/c/9EcGLugq/2267-course-counts-on-organisation-list-show-the-wrong-number-of-courses)

PR #668, opened 2019-10-01
[2020] Add fields to confirmation page

danielburnley:
### Context

We need to display the data for the course in the confirmation page so the user can see what they have selected at a glance

Adds the fields to the course confirmation page for the data we gather at the moment

image

### DIFF
--- a/app/views/courses/confirmation.html.erb
+++ b/app/views/courses/confirmation.html.erb
@@ -8,16 +8,194 @@
     <%= form_with model: course,
                   url: new_provider_recruitment_cycle_course_path(@provider.provider_code, @provider.recruitment_cycle_year),
                   method: :post do |f| %>
-
       <hr class="govuk-section-break govuk-section-break--l">
+      <div class="govuk-grid-row govuk-!-margin-bottom-9">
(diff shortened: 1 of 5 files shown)


app/views/courses/confirmation.html.erb
```diff
@@ -8,16 +8,194 @@
     <%= form_with model: course,
                   url: new_provider_recruitment_cycle_course_path(@provider.provider_code, @provider.recruitment_cycle_year),
                   method: :post do |f| %>
-
       <hr class="govuk-section-break govuk-section-break--l">
+      <div class="govuk-grid-row govuk-!-margin-bottom-9">
```
dankmitchell: Any way of sharing the markup from the basic details view?

danielburnley: Potentially - but I held off prematurely generalising it for now. Gut feeling is that by prematurely trying to share it, those change links are going to be more difficult to implement. 

I'd rather see what ends up being shared, and what doesn't, and generalise from there

dankmitchell: Good with me

PR #686, opened 2019-10-07
Fix course status badges not showing up for new courses

danielburnley:
### Context

Status badges were not showing up for new courses

### DIFF
--- a/app/models/course.rb
+++ b/app/models/course.rb
@@ -80,6 +80,10 @@ class Course < Base
     content_status == "withdrawn"
   end
 
+  def is_not_withdrawn?
+    content_status != "withdrawn"
+  end
+
   def running_site_statuses
     site_statuses.select(&:running?)
   end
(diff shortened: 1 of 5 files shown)


spec/decorators/course_decorator_spec.rb
```diff
@@ -83,4 +83,80 @@
...
     end
   end
+
+  context "status tag" do
+    let(:status_tag) { course.decorate.status_tag }
+
+    context "A non running course" do
+      let(:course) { build(:course, ucas_status: "not_running") }
+
+      it "Returns nil" do
```
fofr: How many courses does this affect? not_running ought to be the same as withdrawn.

app/models/course.rb
```diff
@@ -80,6 +80,10 @@ def is_withdrawn?
     content_status == "withdrawn"
   end
 
+  def is_not_withdrawn?
```
fofr: Is this method still needed?

app/models/course.rb
```diff
@@ -80,6 +80,10 @@ def is_withdrawn?
     content_status == "withdrawn"
```
fofr:
```suggestion
    content_status == "withdrawn" || not_running?
```

PR #694, opened 2019-10-10
[2308] Don't use a redirect for no-orgs page

timabell:
### Context

Currently redirects to `/40x`

This means that a page refresh will work when the user is granted
access, and reduces risk of users incorrectly getting stuck here.

The new erb is a copy of `app/views/errors/unauthorized.html.erb`

This commit disentangles this use case from the generic 401/403
handling.

### Guidance to review

:ship:


timabell:
[Users accessing a provider url without access are seeing wrong message](https://trello.com/c/PoQ49tJd/2308-users-accessing-a-provider-url-without-access-are-seeing-wrong-message)

PR #700, opened 2019-10-15
[1857] Complete and refactor subject editing for master subjects

faissalMT:
### Context
Users should be able to modify their subjects rather than having to go through support to do so.

Allows the user to set single subjects.

### DIFF
--- a/app/controllers/courses/subjects_controller.rb
+++ b/app/controllers/courses/subjects_controller.rb
@@ -5,9 +5,13 @@ module Courses
     before_action :build_course, only: %i[edit update]
 
     def update
-      master_subject = params.dig(:course, :master_subject)
+      master_subject_id = params.dig(:course, :master_subject_id)
+      master_subject_hash = @course.meta[:edit_options][:subjects].find do |subject|
+        subject[:id] == master_subject_id
+      end
+      master_subject = Subject.new(master_subject_hash.to_h)
 
-      if @course.update(subjects: [Subject.new(id: master_subject)])
+      if @course.update(subjects: [master_subject])
         flash[:success] = "Your changes have been saved"
         redirect_to(
           details_provider_recruitment_cycle_course_path(
--- a/spec/decorators/course_decorator_spec.rb
+++ b/spec/decorators/course_decorator_spec.rb
@@ -4,6 +4,10 @@ describe CourseDecorator do
   let(:current_recruitment_cycle) { build :recruitment_cycle }
   let(:next_recruitment_cycle) { build :recruitment_cycle, :next_cycle }
   let(:provider) { build(:provider, accredited_body?: false) }
+  let(:english) { build(:subject, :english) }
+  let(:mathematics) { build(:subject, :mathematics) }
+  let(:subjects) { [english, mathematics] }
+
   let(:course) {
     build :course,
           course_code: "A1",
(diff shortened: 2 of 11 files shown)


faissalMT:
[Edit Course 'Subject' (MCFE)](https://trello.com/c/JO6NoJWs/1857-edit-course-subject-mcfe)

app/controllers/courses/subjects_controller.rb
```diff
@@ -7,7 +7,9 @@ class SubjectsController < ApplicationController
     def update
       master_subject = params.dig(:course, :master_subject)
 
-      if @course.update(subjects: [Subject.new(id: master_subject)])
+      subjects = @course.meta[:edit_options][:subjects].find { |subject| subject[:id] == master_subject }
```
misaka: Looks like this should be `subject` not `subjects`?

app/controllers/courses/subjects_controller.rb
```diff
@@ -7,7 +7,9 @@ class SubjectsController < ApplicationController
...
+
+      if @course.update(subjects: [Subject.new(subjects.to_h)])
```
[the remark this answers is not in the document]
misaka: Designs from: https://bat-design-history.herokuapp.com/publish-teacher-training/new-course-iteration-14-jan#pick-a-subject

spec/decorators/course_decorator_spec.rb
```diff
@@ -4,6 +4,10 @@
   let(:current_recruitment_cycle) { build :recruitment_cycle }
   let(:next_recruitment_cycle) { build :recruitment_cycle, :next_cycle }
   let(:provider) { build(:provider, accredited_body?: false) }
+  let(:english) { build(:subject, subject_name: "English") }
+  let(:mathematics) { build(:subject, subject_name: "Mathematics") }
```
misaka: Would it makes sense for these to be traits on the `subject` factory? Seems to me that it'd reduce verbosity and chance or error.

spec/decorators/course_decorator_spec.rb
```diff
@@ -167,4 +168,21 @@
       end
     end
   end
+
+  context "selectable master subjects" do
```
misaka: I think you're describing `#selectable_master_subjects` here.

spec/features/courses/subjects/edit_spec.rb
```diff
@@ -30,29 +30,45 @@
     stub_api_v2_resource(course, include: "subjects,sites,provider.sites,accrediting_provider")
   end
 
-  scenario "can visit the subject edit page" do
-    course_details_page.load_with_course(course)
-    expect(course_details_page).to be_displayed
-    course_details_page.edit_subjects_link.click
-    expect(subjects_page).to be_displayed
+  context "with given subjects" do
```
misaka: Not sure what the difference is between this `"with given subjects"` context, and the `"with a given set if subjects"` context below? And it seems the way `course` is built here is essentially identical  to the one on line 14, so is this unnecessary extra noise?

spec/support/helpers.rb
```diff
@@ -58,15 +58,16 @@ def stub_api_v2_request(url_path, stub, method = :get, status = 200, token: nil,
 
   def stub_api_v2_resource(resource,
                            jsonapi_response: nil,
-                           include: nil)
+                           include: nil,
+                           method: :get)
```
misaka: :100:

app/controllers/courses/subjects_controller.rb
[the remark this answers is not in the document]
faissalMT: I've discussed this with @dankmitchell and he said that phase one will be just having the ability to set one subject.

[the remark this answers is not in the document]
misaka: 👍

PR #705, opened 2019-10-17
[2365] Add subjects to course creation workflow

faissalMT:
### Context
We would like to be able to set single subjects during course creation.  

Add the new subject page and relevant plumbing.

### DIFF
--- a/spec/features/courses/subjects/new_spec.rb
+++ b/spec/features/courses/subjects/new_spec.rb
@@ -1,0 +1,44 @@
+require "rails_helper"
+
+feature "New course level", type: :feature do
+  let(:new_subjects_page) do
+    PageObjects::Page::Organisations::Courses::NewSubjectsPage.new
+  end
+  let(:next_step_page) do
+    PageObjects::Page::Organisations::Courses::NewAgeRangePage.new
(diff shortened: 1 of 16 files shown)


faissalMT:
[Add subjects to new course flow](https://trello.com/c/aqWICxft/2365-add-subjects-to-new-course-flow)

spec/features/courses/subjects/new_spec.rb
```diff
@@ -0,0 +1,44 @@
+require "rails_helper"
```
misaka: I wonder about the value of this test ... it basically tests a subset of the "new course" feature, but because it uses Capybara it's going to be slow. If we want additional unit-level tests, would it make sense to rely on request specs instead?

faissalMT: I'm not sure how I could do any useful testing of this functionality with request specs.

misaka: Discussed IRL, decided there's value in testing this page in isolation as a feature test, as errors from previous steps could change the behaviour of this one and lead to false positive/negatives. 👍

PR #708, opened 2019-10-21
[2369] Add change links for course creation

faissalMT:
### Context
As it stands in order to change parts of course that is currently being created one must create a new course from scratch or use the back button.

This PR allows someone to edit individual properties from the confirmation screen.

### Guidance to review
You will often see occurrences of `2.times` when visiting a page, for some reason when pages are visited for the first time they will not have their query parameters and visiting it twice fixes things. Ideas on how to better indicate this are appreciated and/or an existing Capybara issue.  
Because sites changing is not fully implemented the change link will simply redirect the user to the next step.
If #705 is merged first, this PR will need to be updated, if not we will need another PR to add the change to subjects.

### DIFF
--- a/app/controllers/concerns/course_basic_detail_concern.rb
+++ b/app/controllers/concerns/course_basic_detail_concern.rb
@@ -5,6 +5,7 @@ module CourseBasicDetailConcern
     decorates_assigned :course
     before_action :build_provider, :build_new_course, only: %i[new continue]
     before_action :build_previous_course_creation_params, only: %i[new continue]
+    before_action :build_meta_course_creation_params, only: %i[new continue]
     before_action :build_course, only: %i[edit update]
   end
 
@@ -36,6 +37,8 @@ module CourseBasicDetailConcern
 
     if @errors.present?
       render :new
+    elsif params[:goto_confirmation].present?
+      redirect_to confirmation_provider_recruitment_cycle_courses_path(path_params)
     else
       redirect_to next_step
     end
@@ -75,6 +78,7 @@ private
           :year,
           :course_age_range_in_years_other_from,
           :course_age_range_in_years_other_to,
+          :goto_confirmation,
         )
         .permit(
           :page,
@@ -117,6 +121,10 @@ private
     @course_creation_params = course_params
   end
 
+  def build_meta_course_creation_params
+    @meta_course_creation_params = params.slice(:goto_confirmation)
+  end
+
   def next_step
     next_step = NextCourseCreationStepService.new.execute(current_step: current_step, current_provider: @provider)
     next_page = course_creation_path_for(next_step)
@@ -128,9 +136,11 @@ private
     next_page
   end
 
-  def course_creation_path_for(page)
-    path_params = { course: course_params }
+  def path_params
+    { course: course_params }
+  end
 
+  def course_creation_path_for(page)
     case page
     when :apprenticeship
       new_provider_recruitment_cycle_courses_apprenticeship_path(path_params)
--- a/app/views/courses/_basic_details_tab.html.erb
+++ b/app/views/courses/_basic_details_tab.html.erb
@@ -120,7 +120,7 @@
           Locations
         </dt>
         <dd class="govuk-summary-list__value" data-qa="course__locations">
-          <% if course.sites.nil? ||course.sites.empty? %>
+          <% if course.sites.nil? || course.sites.empty? %>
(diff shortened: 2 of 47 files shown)


app/views/courses/_basic_details_tab.html.erb
```diff
@@ -120,7 +120,7 @@
           Locations
         </dt>
         <dd class="govuk-summary-list__value" data-qa="course__locations">
-          <% if course.sites.empty? %>
+          <% if course.sites.nil? || course.sites.empty? %>
```
misaka: I still get an error from `CourseDecorator#has_site?`:

> undefined method `any?' for nil:NilClass

Presumably because the course's sites are `nil` and not `[]`. Should that also be protected from like it is here?

spec/site_prism/page_objects/page/organisations/courses/new_apprenticeship_page.rb
```diff
@@ -11,7 +11,6 @@ class NewApprenticeshipPage < CourseBase
           end
 
           element :save, '[data-qa="course__save"]'
```
defong: delete me

misaka:
> You will often see occurrences of 2.times when visiting a page, for some reason when pages are visited for the first time they will not have their query parameters and visiting it twice fixes things.

This is being caused by our sign-in redirect: when the user isn't authenticated we redirect them to sign-in, and save the path they're trying to visit so we can redirect them there later. Unfortunately, we strip off the query params when we do this, which is a bug that needs to be addressed more generally.

@faissalMT How about we explicitly visit a URL that will sign us in before the tests here that rely on query params. e.g. `/signin` perhaps.

spec/features/courses/study_mode/new_spec.rb
```diff
@@ -46,4 +49,20 @@
...
+    new_study_mode_page.study_mode_fields.full_time_or_part_time.click
+    new_study_mode_page.continue.click
+
+    expect(current_path).to eq confirmation_provider_recruitment_cycle_courses_path(provider.provider_code, provider.recruitment_cycle_year)
+  end
+
+private
+
+  def visit_new_study_mode_page(**query_params)
+    2.times do
```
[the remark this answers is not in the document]
faissalMT: See guidance to review

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -116,6 +120,20 @@ def build_previous_course_creation_params
...
+      .except(
+        :on,
+        :provider_code,
+        :recruitment_cycle_year,
+        :utf8,
+        :course,
+        :commit,
+        :format,
+      )
+      .permit(:goto_confirmation)
```
misaka: We aren't using these params for mass-assigns, so I don't think `permit` is doing what we want here. Technically, it could allow us to assign to `goto_confirmation` in the course model, which doesn't exist, and is confusing.

Can we pluck out the attributes we care about in a more appropriate way?

app/views/courses/confirmation.html.erb
```diff
@@ -26,7 +25,9 @@
                 <%= course.level&.humanize %>
               </dd>
               <dd class="govuk-summary-list__actions">
+                <%= link_to new_provider_recruitment_cycle_courses_level_path(@provider.provider_code, course.recruitment_cycle_year, params.to_unsafe_h.merge(goto_confirmation: true)), class: "govuk-link", data: { qa: "course__edit_level_link" } do %>
```
misaka: This is making my eyes water. Can we use a helper method or something to help make this more readable pls?

spec/features/courses/age_range_in_years/new_spec.rb
```diff
@@ -25,21 +29,36 @@
...
+    visit_new_age_range_page(goto_confirmation: true)
+
+    choose("course_age_range_in_years_3_to_7")
+    new_age_range_page.continue.click
+
+    expect(current_path).to eq confirmation_provider_recruitment_cycle_courses_path(provider.provider_code, provider.recruitment_cycle_year)
+  end
+
+  def visit_new_age_range_page(**query_params)
+    2.times do
```
misaka: As mentioned, could we visit the sign-in page before we hit this one to make it clear what we're doing?

defong: oh right so user not quite authenticated and getting sent in circles

PR #718, opened 2019-10-24
[2326] Add modern language subjects page

faissalMT:
### Context
As it stands modern languages cannot be set for the modern language subjects.  

Add a page for modifying modern language subjects so that modern subjects can be set.

### Guidance to review
This also contains a fix for subjects not being autoselected as pointed out by @fofr 

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -17,7 +17,9 @@ module Courses
     end
 
     def update
-      if @course.update(subjects: old_non_language_subjects + new_language_subjects)
+      updated_subject_list = strip_non_language_subjects
+      updated_subject_list += selected_language_subjects
+      if @course.update(subjects: updated_subject_list)
         flash[:success] = "Your changes have been saved"
         redirect_to(
           details_provider_recruitment_cycle_course_path(
@@ -34,19 +36,14 @@ module Courses
 
   private
 
-    def new_language_subjects
-      language_ids = params.dig(:course, :language_ids)
-      found_languages_ids = available_languages_ids & language_ids
-
-      found_languages_ids.map do |language_id|
-        Subject.new(id: language_id)
-      end
+    def strip_non_language_subjects
+      @course.subjects.reject { |s| available_languages_ids.include?(s.id) }
     end
 
-    def old_non_language_subjects
-      @course.subjects.reject do |subject|
-        available_languages_ids.include?(subject.id)
-      end
+    def selected_language_subjects
+      language_ids = params.dig(:course, :language_ids)
+      found_languages_ids = available_languages_ids & language_ids
+      found_languages_ids.map { |id| Subject.new(id: id) }
     end
 
     def available_languages_ids
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -1,0 +1,73 @@
+module Courses
+  class ModernLanguagesController < ApplicationController
+    include CourseBasicDetailConcern
+    decorates_assigned :course
+    before_action :build_course, only: %i[edit update]
+
+    def edit
+      return unless @course.meta[:edit_options][:modern_languages].nil?
+
+      redirect_to(
+        details_provider_recruitment_cycle_course_path(
+          @course.provider_code,
+          @course.recruitment_cycle_year,
+          @course.course_code,
+        ),
+      )
+    end
+
+    def update
+      if @course.update(subjects: old_non_language_subjects + new_language_subjects)
+        flash[:success] = "Your changes have been saved"
+        redirect_to(
+          details_provider_recruitment_cycle_course_path(
+            @course.provider_code,
+            @course.recruitment_cycle_year,
+            @course.course_code,
+          ),
+        )
+      else
+        @errors = @course.errors.messages
+        render :edit
+      end
+    end
+
+  private
+
+    def new_language_subjects
+      language_ids = params.dig(:course, :language_ids)
+      found_languages_ids = potential_languages_ids & language_ids
+
+      found_languages_ids.map do |language_id|
+        Subject.new(id: language_id)
+      end
+    end
+
+    def old_non_language_subjects
+      @course.subjects.reject do |subject|
+        potential_languages_ids.include?(subject.id)
+      end
+    end
+
+    def potential_languages_ids
+      @course.meta[:edit_options][:modern_languages].map do |language|
+        language["id"]
+      end
+    end
+
+    def current_step
+      nil
+    end
+
+    def errors; end
+
+    def build_course
+      @course = Course
+                  .includes(:subjects, :site_statuses)
+                  .where(recruitment_cycle_year: params[:recruitment_cycle_year])
+                  .where(provider_code: params[:provider_code])
+                  .find(params[:code])
+                  .first
+    end
+  end
+end
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -36,7 +36,7 @@ module Courses
 
     def new_language_subjects
       language_ids = params.dig(:course, :language_ids)
-      found_languages_ids = potential_languages_ids & language_ids
+      found_languages_ids = available_languages_ids & language_ids
 
       found_languages_ids.map do |language_id|
         Subject.new(id: language_id)
@@ -45,11 +45,11 @@ module Courses
 
     def old_non_language_subjects
       @course.subjects.reject do |subject|
-        potential_languages_ids.include?(subject.id)
+        available_languages_ids.include?(subject.id)
       end
     end
 
-    def potential_languages_ids
+    def available_languages_ids
       @course.meta[:edit_options][:modern_languages].map do |language|
         language["id"]
       end
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -55,10 +55,6 @@ module Courses
       end
     end
 
-    def current_step
-      nil
-    end
-
     def errors; end
 
     def build_course
--- a/app/views/courses/modern_languages/edit.erb
+++ b/app/views/courses/modern_languages/edit.erb
@@ -1,0 +1,39 @@
+<%= content_for :page_title, "Change modern languages – #{course.name_and_code}" %>
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(details_provider_recruitment_cycle_course_path(course.provider_code, course.recruitment_cycle_year, course.course_code)) %>
+<% end %>
+
+<%= render 'shared/errors' %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with model: course,
+          url: modern_languages_provider_recruitment_cycle_course_path(@course.provider_code, @course.recruitment_cycle_year, @course.course_code),
+          method: :put do |form| %>
+      <div
+        class="<%= classnames("govuk-form-group", "govuk-form-group--error": @errors && @errors[:subjects]&.any?) %>"
+        data-qa="course__modern_languages_section">
+        <fieldset class="govuk-fieldset">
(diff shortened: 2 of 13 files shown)


misaka:
As commented on DFE-Digital/manage-courses-backend#926 there's a slight miss-naming of `potential_modern_languages` in `edit_options` which probably should be just `modern_languages` to keep things consistent. I'll make this change as @faissalMT is off ill today.

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    end
+
+    def potential_languages_ids
+      @course.meta[:edit_options][:modern_languages].map do |language|
+        language["id"]
+      end
+    end
+
+    def current_step
+      nil
```
defong: looks odd...

app/views/courses/subjects/_form_fields.html.erb
```diff
@@ -1 +1 @@
-<%= form.select :master_subject_id, course.selectable_master_subjects, {}, data: { qa: 'course__subjects' }, class: "govuk-select" %>
+<%= form.select :master_subject_id, options_for_select(course.selectable_master_subjects, @course.subjects.first&.id), {}, data: { qa: 'course__subjects' }, class: "govuk-select" %>
```
timabell: The use of `first` is surprising to me in here. What's the reason for it?

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,73 @@
+module Courses
+  class ModernLanguagesController < ApplicationController
+    include CourseBasicDetailConcern
+    decorates_assigned :course
```
timabell: what does `decorates_assigned` do?

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+        details_provider_recruitment_cycle_course_path(
+          @course.provider_code,
+          @course.recruitment_cycle_year,
+          @course.course_code,
+        ),
+      )
+    end
+
+    def update
+      if @course.update(subjects: old_non_language_subjects + new_language_subjects)
```
timabell: I'm struggling to decipher the intent behind `old_non_language_subjects + new_language_subjects`

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+    def new_language_subjects
+      language_ids = params.dig(:course, :language_ids)
+      found_languages_ids = potential_languages_ids & language_ids
+
+      found_languages_ids.map do |language_id|
+        Subject.new(id: language_id)
+      end
+    end
+
+    def old_non_language_subjects
```
timabell: it's not obvious to me what this method is for and why it does what it does

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+        )
+      else
+        @errors = @course.errors.messages
+        render :edit
+      end
+    end
+
+  private
+
+    def new_language_subjects
```
timabell: it's not obvious to me from reading the code what this method is for and why it does what it does

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,73 @@
...
+      end
+    end
+
+    def old_non_language_subjects
+      @course.subjects.reject do |subject|
+        potential_languages_ids.include?(subject.id)
+      end
+    end
+
+    def potential_languages_ids
```
timabell: I agree with @defong that this needs a better name, it doesn't jump out at me what this is for.

app/views/courses/modern_languages/edit.erb
```diff
@@ -0,0 +1,39 @@
...
+        data-qa="course__modern_languages_section">
+        <fieldset class="govuk-fieldset">
+          <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
+            <h1 class="govuk-fieldset__heading">
+              <span class="govuk-caption-xl"><%= course.name_and_code %></span>
+              Change modern languages
+            </h1>
+          </legend>
+          <%= render "shared/error_messages", field: :subjects if @errors %>
+          <%= render 'form_fields', form: form %>
```
timabell: it's ambiguous which `form_fields` is being referred to as there are many in our code

can we do the same as https://github.com/DFE-Digital/manage-courses-frontend/blob/5f1a40fd1b065a963da98d2e90b6b18a752ff51b/app/views/courses/level/new.html.erb#L29 ?

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      languages_page.load_with_course(course)
+      edit_languages_stub = stub_api_v2_resource(course, method: :patch)
+
+      languages_page.languages_fields.find('[data-qa="checkbox_language_French"]').click
+      languages_page.languages_fields.find('[data-qa="checkbox_language_Japanese"]').click
+      languages_page.save.click
+      expect(course_details_page).to be_displayed
+
+      expect(edit_languages_stub.with do |request|
+        subjects = JSON.parse(request.body)["data"]["relationships"]["subjects"]["data"]
```
timabell: I made a tidier version of this that you might want to use https://github.com/DFE-Digital/manage-courses-frontend/blob/875d6eed588e69ca961839c868692d960fe00495/spec/features/providers/ucas_alerts_spec.rb#L86

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      }
+    end
+
+    scenario "can select multiple modern language subjects" do
+      languages_page.load_with_course(course)
+      edit_languages_stub = stub_api_v2_resource(course, method: :patch)
+
+      languages_page.languages_fields.find('[data-qa="checkbox_language_French"]').click
+      languages_page.languages_fields.find('[data-qa="checkbox_language_Russian"]').click
+      languages_page.languages_fields.find('[data-qa="checkbox_language_Japanese"]').click
```
timabell: could we add something that takes a parameter to the page object and push the data-qa stuff down into there?

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+      languages_page.save.click
+      expect(course_details_page).to be_displayed
+
+      expect(edit_languages_stub.with do |request|
+        subjects = JSON.parse(request.body)["data"]["relationships"]["subjects"]["data"]
```
timabell: see https://github.com/DFE-Digital/manage-courses-frontend/pull/718/files#r340139981

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -0,0 +1,131 @@
...
+        expect(subjects).to match_array([
+          include("id" => modern_languages_subject.id.to_s),
+          include("id" => russian_subject.id.to_s),
+          include("id" => japanese_subject.id.to_s),
+        ])
+      end).to have_been_made
+    end
+
+    scenario "checks the current modern language subjects is selected" do
```
timabell: for the sake of test speed I'd squash this into one of the other tests, the full load isn't very fast so we want to minimize the number of page loads

app/views/courses/subjects/_form_fields.html.erb
faissalMT: This might want to be split into another PR, it's setting the default for the dropdown to the first item in `@course.subjects`

timabell:
I might pick this up... depends who gets there first

timabell:
This depends on https://github.com/DFE-Digital/manage-courses-backend/pull/926

timabell:
Good work, that's quite a sizeable patch.
I have made some comments & suggestions.

I'm a bit worried by how hard it is to understand why the controller is doing what it does from a cold read of the code. I'd like to see the higher-level context and intent expressed more clearly somehow. In its current form I wouldn't want to have to maintain this in the future as I wouldn't be confident I understood why it behaves as it does and whether I could change it without going all the way back to product design.

timabell:
> Need to rename `potential_modern_languages` to `modern_languages`.

done

app/controllers/courses/modern_languages_controller.rb
timabell: gone

app/views/courses/subjects/_form_fields.html.erb
timabell: reverting this causes a test to fail

```
  1) Edit course subjects with a subject already set it should automatically select the current subject
     Failure/Error: expect(subjects_page).to have_select("course_master_subject_id", selected: english_subject.subject_name)
       expected to find visible select box "course_master_subject_id" that is not disabled with "English" selected but there were no matches. Also found "English Biology", which matched the 
selector but not all filters. Expected "English" to be selected found []
     # ./spec/features/courses/subjects/edit_spec.rb:78:in `block (3 levels) in <top (required)>'
                   
Finished in 29.6 seconds (files took 4.05 seconds to load)
439 examples, 1 failure, 1 pending                                   
                                                                                                
Failed examples:

rspec ./spec/features/courses/subjects/edit_spec.rb:76 # Edit course subjects with a subject already set it should automatically select the current subject
```

app/controllers/courses/modern_languages_controller.rb
timabell: explanation:

We are modifying the list of *language* subjects, but leaving the other types of subjects intact.
Because it's a replacemend of the whole subject list we need to reconstruct the combined result.

app/controllers/courses/modern_languages_controller.rb
faissalMT: https://github.com/drapergem/draper

misaka:
@timabell @faissalMT What surprised me about this is that I thought it would be part of the create course flow, however "Modern Languages" isn't listed as a subject when creating a course. This PR may still be good, but it's hard to product review fully without a modern language.

misaka:
Steps for getting to this page for testing:

1) Find an existing course that has "Modern Languages" in the title (i.e. a course with the "Modern Languages" subject).
2) Go to the course's basic details.
3) Replace "details" witn "modern-languages" in the URL.

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+        @errors = @course.errors.messages
+        render :edit
+      end
+    end
+
+  private
+
+    def strip_non_language_subjects
+      @course.subjects.reject { |s| available_languages_ids.include?(s.id) }
+    end
```
misaka: For the future, I think we should be having subjects listed as separate things on the course, instead of having to trawl through `subjects` taking out languages, etc. This is probably something to address in a later, wider-ranging PR.

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -0,0 +1,66 @@
...
+      @course.meta[:edit_options][:modern_languages].map do |language|
+        language["id"]
+      end
+    end
+
+    def errors; end
+
+    def build_course
+      @course = Course
+                  .includes(:subjects, :site_statuses)
```
misaka: Don't see why we need to include `site_statuses` here. Should be addressed in a future fix.

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -0,0 +1,134 @@
...
+      patch_course_stub = set_patch_course_expectation do |subjects|
+        expect(subjects).to match_array([
+                                          include("id" => modern_languages_subject.id.to_s),
+                                          include("id" => french_subject.id.to_s),
+                                          include("id" => japanese_subject.id.to_s),
+                                        ])
+      end
+
+      languages_page.languages_fields.find('[data-qa="checkbox_language_French"]').click
+      languages_page.languages_fields.find('[data-qa="checkbox_language_Japanese"]').click
```
misaka: I think these should be using the SitePrism page object, either as an `element` or a page method that returns the correct checkbox.

PR #719, opened 2019-10-24
[2405] Show no-organisations page if user has no user record

timabell:
### Context

The backend returns 401 if there is no user account. The last cleanup of
the status handling code left this user journey showing the "something
went wrong" unauthorized page.

* Show no-organisations page if user has no user record
* All the test coverage

This change means that the same page is shown regardless of whether you
have no account in our database (401) or have an account but no access
to any providers (403), but the response codes to the browser are
reflective of which code the backend responded with (401/403).

### Guidance to review

:ship:?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -10,7 +10,10 @@ class ApplicationController < ActionController::Base
   end
 
   def render_unauthorized
-    respond_with_error(template: "errors/unauthorized", status: :unauthorized, error_text: "Unauthorized request")
+    # Backend responds with 401 if there is no matching record in the users table.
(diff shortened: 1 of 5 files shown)


timabell:
[Users with no record in our database seeing "something went wrong" 401 page](https://trello.com/c/xVLNAgxl/2405-users-with-no-record-in-our-database-seeing-something-went-wrong-401-page)

timabell:
Updated test coverage to match new behaviour.

It didn't fail before because both error pages had the same `data-qa` value. Oops!

timabell:
But wait, there's more!....

* Added feature specs for rest of scenarios now I understand it all better

PR #721, opened 2019-10-24
[2409] login without dfe sign in

misaka:
### Context

To reduce the friction to developing the app. So we can develop without requiring online access. etc.

Add support for basic auth in `development` env, and if special config is provided.

### Guidance to review

Tests still have to be fixed, but I've run this locally and it seems to work. That includes existing functionality of sign-in and sign-out with DFE Sign-In.

But ... is this even a good idea?

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/README.md
+++ b/README.md
@@ -106,3 +106,17 @@ To track exceptions through Sentry, configure the `SENTRY_DSN` environment varia
 ```
 SENTRY_DSN=https://aaa:bbb@sentry.io/123 rails s
 ```
+
+
+## Using Basic Auth Instead of DFE Sign-In
+
+For local development, you can disable reliance on DFE Sign-In by creating a
+file `config/settings/development.local.yml` with the contents:
+
+```yaml
+authorised_users:
+  your.email@digital.education.gov.uk: $LOCAL_PASSWORD
+```
+
+The email address has to exist in the users table of manage-courses-backend, but
+the password can be any non-secure local password you care to use.
--- a/README.md
+++ b/README.md
@@ -114,8 +114,11 @@ For local development, you can disable reliance on DFE Sign-In by creating a
 file `config/settings/development.local.yml` with the contents:
 
 ```yaml
-authorised_users:
-  your.email@digital.education.gov.uk: $LOCAL_PASSWORD
+authorised_user:
+  first_name: [your first name, this will be updated in the db]
+  last_name: [your last name, this will be updated in the db]
+  email: [the email address to login with]
+  password: [the password you wish to use]
 ```
 
 The email address has to exist in the users table of manage-courses-backend, but
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -27,7 +27,12 @@ class ApplicationController < ActionController::Base
   helper_method :current_user
 
   def current_user
-    @current_user ||= session[:auth_user] if session[:auth_user]
+    if session[:auth_user]
+      @current_user ||= session[:auth_user]
+    elsif development_mode_auth?
+      email = authenticate_with_http_basic { |u, p| p == development_mode_password(u) && u }
+      setup_development_mode_session(email) if email
+    end
   end
 
   def log_safe_current_user(reload: false)
@@ -38,6 +43,10 @@ class ApplicationController < ActionController::Base
       @log_safe_current_user["email_md5"] = Digest::MD5.hexdigest(email)
     end
     @log_safe_current_user
+  end
+
+  def development_mode_auth?
+    Settings.key?(:authorised_users)
   end
 
   def authenticate
@@ -55,6 +64,9 @@ class ApplicationController < ActionController::Base
         logger.debug { "User session set. " + log_safe_current_user(reload: true).to_s }
       end
 
+    elsif development_mode_auth?
+      logger.debug("Develepmont mode authenticated " + log_safe_current_user.to_s)
+      request_http_basic_authentication("Development Mode")
     else
       logger.debug("Authenticated user session not found " + {
                      redirect_back_to: request.path,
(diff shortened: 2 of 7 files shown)


defong:
[Login without sign-in in local development mode.](https://trello.com/c/8gNckQYG/2409-login-without-sign-in-in-local-development-mode)

app/controllers/sessions_controller.rb
```diff
@@ -30,7 +30,21 @@ def create
 
   def signout
     if current_user.present?
-      redirect_to "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=#{current_user['credentials']['id_token']}&post_logout_redirect_uri=#{Settings.dfe_signin.base_url}/auth/dfe/signout"
+      if development_mode_auth?
+        # Dissapointingly, with HTTP basic auth it's trick to really log
```
defong:
```suggestion
        # Disappointingly, with HTTP basic auth it's trick to really log
```

app/controllers/application_controller.rb
```diff
@@ -37,6 +42,10 @@ def log_safe_current_user(reload: false)
     @log_safe_current_user
   end
 
+  def development_mode_auth?
+    Settings.key?(:authorised_users)
```
defong: should we double lock it...
ie `rails_env != "producation" `

README.md
```diff
@@ -106,3 +106,17 @@ To track exceptions through Sentry, configure the `SENTRY_DSN` environment varia
...
 ```
+
+
+## Using Basic Auth Instead of DFE Sign-In
+
+For local development, you can disable reliance on DFE Sign-In by creating a
+file `config/settings/development.local.yml` with the contents:
+
+```yaml
+authorised_users:
```
defong:
```suggestion
authorised_user:
```
singular?


```yml
authorised_user:
 email:
 first_name:
 last_name:
 insecure_password:

```

app/controllers/application_controller.rb
```diff
@@ -80,6 +92,26 @@ def has_multiple_providers?
...
+    Rails.env.development? && Settings&.authorised_users&.[](email)
+  end
+
+  def setup_development_mode_session(email)
+    session[:auth_user] = HashWithIndifferentAccess.new(
+      uid: SecureRandom.uuid,
+      info: HashWithIndifferentAccess.new(
+        email: email,
+        # We don't have the user's name here, but it doesn't matter what we send
+        # to the backend since we're in development mode now anyway, right??!
```
defong: backend takes the first and last name and overwrites what's persisted so maybe

```yml
authorised_user:
 email:
 first_name:
 last_name:
 insecure_password:

```

spec/requests/sessions_spec.rb
```diff
@@ -7,7 +7,9 @@
       get(auth_dfe_callback_path)
 
       get(signout_path)
-      expect(response).to redirect_to("#{Settings.dfe_signin.issuer}/session/end?id_token_hint=&post_logout_redirect_uri=https://localhost:3000/auth/dfe/signout")
+      expect(response).to redirect_to(
+                            "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=&post_logout_redirect_uri=https%3A%2F%2Flocalhost%3A3000%2Fauth%2Fdfe%2Fsignout"
```
defong: "#{Settings.dfe_signin.issuer}/session/end?id_token_hint=`MISSING???`&post_logout_redirect_uri=https%3A%2F%2Flocalhost%3A3000%2Fauth%2Fdfe%2Fsignout"

missing `id_token_hint` and url encoding?

misaka: `id_token_hint` is in both the old and the new there, the diff is the URL encoding, which is just a side-effect of cleaning up `SessionsController#signout` because I couldn't read what it was doing. We probably really do want to URL encode the URL being included as a parameter in another URL, turns out, so net win.

README.md
misaka: I was thinking we could have multiple users there, but I like your idea of adding the `first_name` and `last_name`, so ya, will have a look at changing that.

app/controllers/application_controller.rb
misaka: Ya, that's in `development_mode_password`, but it's probably worth double checking here too. Although I'd probably still use `Rails.env.development?`.

spec/requests/sessions_spec.rb
defong: misha net win go go go :+1:

app/controllers/application_controller.rb
defong: all env are playground apart from `production` as long as she don't shot our foot, I'm fine

app/controllers/application_controller.rb
```diff
@@ -83,6 +103,27 @@ def has_multiple_providers?
...
+      Settings.authorised_user.password == password
+  end
+
+  def setup_development_mode_session(email:, first_name:, last_name:)
+    session[:auth_user] = HashWithIndifferentAccess.new(
+      uid: SecureRandom.uuid,
+      info: HashWithIndifferentAccess.new(
+        email: email,
+        # We don't have the user's name here, but it doesn't matter what we send
+        # to the backend since we're in development mode now anyway, right??!
```
defong: wrong you have it now....

app/controllers/application_controller.rb
```diff
@@ -55,6 +72,9 @@ def authenticate
         logger.debug { "User session set. " + log_safe_current_user(reload: true).to_s }
       end
 
+    elsif development_mode_auth?
+      logger.debug("Doing develepmont mode authentication")
```
defong:
```suggestion
      logger.debug("Doing develepment mode authentication")
```

app/controllers/application_controller.rb
```diff
@@ -27,7 +27,20 @@ def handle_access_denied(exception)
...
+      email = authenticate_with_http_basic do |user, pass|
+        user if authorise_development_mode?(user, pass)
+      end
+      if email
+        setup_development_mode_session(
+          email: Settings.authorised_user.email,
+          first_name: Settings.authorised_user.first_name,
+          last_name: Settings.authorised_user.last_name,
+        )
+      end
```
defong: else raise hell as its a no op?

app/controllers/application_controller.rb
fofr:
```suggestion
      logger.debug("Doing development mode authentication")
```

app/controllers/application_controller.rb
misaka: No, I don't understand all the intricacies here, but if I raise an error it blocks the login attempt. I think it has to fall out of here and hit `request_http_basic_authentication` below to actually request the l/p.

defong: ok... i thought it will put someone out of their misery...

timabell:
[get review apps working for publish by disabling sign-in requirement](https://trello.com/c/pNfDKfM0/310-get-review-apps-working-for-publish-by-disabling-sign-in-requirement)

PR #724, opened 2019-10-25
[2414] Strip commas from course fees

fofr:
Be more flexible in what we accept from users.

A user has been struggling to input a number in the course fees box because our validations do not allow commas to be added. eg 9,000 is not accepted.

We need to loosen our validations to allow for this type of value to be entered.

image

https://www.qa.publish-teacher-training-courses.service.gov.uk/organisations/2AT/2020/courses/3CX8/fees

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -70,6 +70,9 @@ class CoursesController < ApplicationController
     end
     params[:course].delete(:course_length_other_length)
 
+    # A user has been struggling to input a number in the course fees box because
+    # our validations do not allow commas to be added. eg 9,000 is not accepted.
+    # By stripping commas out the backend will not reject such input.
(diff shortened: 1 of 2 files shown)


fofr:
[Loosen the validations around our fee numbers](https://trello.com/c/qqjRSHQa/2414-loosen-the-validations-around-our-fee-numbers)

app/controllers/courses_controller.rb
```diff
@@ -70,6 +70,10 @@ def update
     end
     params[:course].delete(:course_length_other_length)
 
+    # Remove commas from fees
```
timabell: I think this comment is redundant, it's sufficiently obvious to me from the following code

timabell:
I wonder if there should be test coverage for this to prevent it being regressed accidentally

timabell:
Pushed some test coverage which I'm not happy with yet, and then realized it's the backend validating this and I think this whole PR should be on the backend. D'oh. Working on it...

PR #726, opened 2019-10-28
[2407] - Improve handling for missing / incorrect DFE Signin credentials (Development)

TobyRet:
### Context
When running the front end application for the first time, the user is required to create a `config/setting/development.local.yml` and add a value for DFE Sign in secret.

A couple of scenarios can occur which can result in errors that are not immediately obvious:
- An incorrect secret is entered
- A typo is made in the yaml file (if the key is incorrect, the default value is used)

https://trello.com/c/D9Ym3r5I/2407-manage-courses-frontend-technical-onboarding

- Rescue and log `Rack::OAuth2::Client::Error` (indicating that the `dfe_signin` secret _could_ be incorrect)

### Missing or incorrect secret
Screenshot 2019-10-28 at 11 12 42

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/initializers/omniauth.rb
+++ b/config/initializers/omniauth.rb
@@ -40,6 +40,9 @@ module OmniAuth
         fail!(:timeout, e)
       rescue ::SocketError => e
         fail!(:failed_to_connect, e)
+      rescue Rack::OAuth2::Client::Error => e
+        Rails.logger.error "Auth failure, is Settings.dfe_signin.secret correct? #{e}"
+        raise
       end
     end
   end


TobyRet:
@misaka - can I pls get your thoughts on the build error?

It's now failing because the dfe_signin key / value has been removed from the default settings `config/settings.yml`(which is required when running the Docker container on CI). The reason this was removed was so that the app could raise an error straight away if `dfe_signin` is not set in `config/development.local.yml`.

It would have been nice to programmatically help the user in this scenario but I wonder whether the best alternative may be to revert back to adding trouble shooting guidance in the README? 

An alternative could be to have a separate config for CI but seems a bit overkill for what is hopefully a rare  scenario ...

timabell:
Could we give the CI an environment variable?

config/initializers/omniauth.rb
```diff
@@ -40,13 +40,19 @@ def callback_phase
         fail!(:timeout, e)
       rescue ::SocketError => e
         fail!(:failed_to_connect, e)
+      rescue Rack::OAuth2::Client::Error => e
+        fail!(:invalid_client, e)
```
timabell: how about `:client_error` to match the caught exception? I wonder if invalid client could cause confusion if we hit the same exception for different reasons

p.s. I'm not really sure what this symbol is for

timabell:
Thanks for working on this :+1:

config/initializers/omniauth.rb
defong: @timabell `:invalid_client` is part of open connect/oauth

misaka: What's the effect of this for the user?

misaka:
Ah, found it in Slack. Looks like it's not much better. Maybe we need a `raise e` inside of the rescue.

TobyRet:
@misaka - I'm slightly stuck on this.
Discovered that the code never reaches any of the custom `rescue` blocks and always defers to the super class. 
I don't think there is much value in this PR anymore, but there is potentially merit in taking a different approach and removing the need for a `development.local.yml` altogether. What do you think?

PR #728, opened 2019-10-28
[2408] Don't clobber modern languages if subject has not changed

faissalMT:
### Context
As it stands if a course has modern languages it will be removed if anything is saved from the subjects page.

If the subject selected has not changed we won't save, this way we won't clobber modern language subjects if the subject hasn't changed.

### Guidance to review
Might want a better success message @fofr 

### Checklist

- [ ] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history


faissalMT:
[Keep modern languages when editing course subject](https://trello.com/c/WkWyAMUQ/2408-keep-modern-languages-when-editing-course-subject)

app/controllers/courses/subjects_controller.rb
```diff
@@ -6,13 +6,16 @@ class SubjectsController < ApplicationController
...
 
     def update
-      master_subject_id = params.dig(:course, :master_subject_id)
-      master_subject_hash = @course.meta[:edit_options][:subjects].find do |subject|
-        subject[:id] == master_subject_id
-      end
-      master_subject = Subject.new(master_subject_hash.to_h)
-
-      if @course.update(subjects: [master_subject])
+      if subjects_match?(new_subjects, old_non_language_subjects)
```
TobyRet: Wonder if there may be an opportunity to increase readability of this if else and move the logic out into private methods?
Pls forgive me for the bad pseudo-code
e.g.

```
return [ if there are errors]
if no_changes_to_subjects?
   [Do the thing]
else
   [update the thing]
end
```

faissalMT: I think the problem here is that the `update()` both returns errors and does work so we can't really know whether or not there are errors beforehand. I don't know that there's a way to get around that without introducing a race condition.

app/controllers/courses/subjects_controller.rb
```diff
@@ -6,13 +6,16 @@ class SubjectsController < ApplicationController
...
+        flash[:success] = "No changes to subjects"
```
timabell: If this is a no-op then I don't think we should flash anything. @fofr ?

spec/features/courses/subjects/edit_spec.rb
```diff
@@ -52,4 +52,35 @@
...
+    end
+
+    scenario "it does not update the course" do
+      subjects_page.load_with_course(course)
+      edit_subject_stub = stub_api_v2_resource(course, method: :patch)
+
+      subjects_page.subjects_fields.select(subjects.first.subject_name)
+      subjects_page.save.click
+      expect(course_details_page).to be_displayed
+      expect(edit_subject_stub).not_to have_been_made
```
timabell: :+1:

timabell:
Looks good to me other than the flash message

PR #732, opened 2019-10-29
[2433] - fix(AgeRange Controller): Code Climate issue

TobyRet:
### Context
This PR fixes two Code Climates issue in the AgeRange controller:
 - "method update exceeds 25 lines of code"
- "Cognitive Complexity of 9 (exceeds 5 allowed)"

A missing feature spec has also been added to cover the scenario where a user wished to update the course age_range with custom values. (_This increases code coverage in the controller by 7% from 80% to 87%_)

- Extracts logic into private methods
- Add missing feature spec

update

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/age_range_controller.rb
+++ b/app/controllers/courses/age_range_controller.rb
@@ -5,26 +5,10 @@ module Courses
     before_action :build_course, only: %i[edit update]
 
     def update
-      # Age range 'other' override
-      course = params.dig(:course)
-      is_other = course.dig(:age_range_in_years) == "other"
-      age_from = course.dig(:course_age_range_in_years_other_from)
-      age_to = course.dig(:course_age_range_in_years_other_to)
-
-      if is_other && (age_from.blank? || age_to.blank?)
-        errors = {
-          age_range_in_years: ["Enter an age for both from and to"],
-          age_range_in_years_from: ["Enter an age"],
-          age_range_in_years_to: ["Enter an age"],
-        }
-      elsif age_from.present? && age_to.present? && is_other
-        params[:course][:age_range_in_years] = "#{age_from}_to_#{age_to}"
-      elsif is_other
-        params[:course][:age_range_in_years] = nil
-      end
-
-      @errors = errors
+      @errors = build_errors
       return render :edit if @errors.present?
+
+      update_age_range_param
 
       if @course.update(course_params)
         flash[:success] = "Your changes have been saved"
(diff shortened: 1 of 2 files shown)


TobyRet:
https://trello.com/c/Kr9foydG/2433-fix-code-climate-issue-agerange-controller

spec/features/courses/age_range_in_years/edit_spec.rb
```diff
@@ -1,4 +1,5 @@
 require "rails_helper"
+require 'byebug'
```
timabell: was that intentional?

TobyRet: Nope! deleting ...

spec/features/courses/age_range_in_years/edit_spec.rb
```diff
@@ -72,7 +91,17 @@
...
+      choose("course_age_range_in_years_other")
+      click_on "Save"
+
+      expect(age_range_in_years_page).to be_displayed
+      expect(age_range_in_years_page.error_flash).to have_content(
+        "You’ll need to correct some information.\nEnter an age for both from and to Enter an age Enter an age"
+      )
+
+      fill_in("course_course_age_range_in_years_other_from", with: "16")
+      fill_in("course_course_age_range_in_years_other_from", with: "19")
```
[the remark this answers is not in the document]
TobyRet: Good spot. Yes

app/controllers/courses/age_range_controller.rb
```diff
@@ -43,6 +27,44 @@ def update
...
+    end
+
+    def build_errors
+      if age_range_is_other? && (age_from_param.blank? || age_to_param.blank?)
+        {
+          age_range_in_years: ["Enter an age for both from and to"],
+          age_range_in_years_from: ["Enter an age"],
+          age_range_in_years_to: ["Enter an age"],
+        }
+      end
```
misaka: While we're here, is this right?? Doesn't this mean that if either is blank, both fields will be flagged with this message?

app/controllers/courses/age_range_controller.rb
```diff
@@ -43,6 +27,44 @@ def update
 
   private
 
+    def update_age_range_param
+      if age_from_param.present? && age_to_param.present? && age_range_is_other?
+        params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}"
+      elsif age_range_is_other?
+        params[:course][:age_range_in_years] = nil
+      end
```
misaka: While we're here, is this sufficiently better to be worth changing it to:

```
      if age_range_is_other?
        if age_from_param.present? && age_to_param.present?
          params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}"
        else
          params[:course][:age_range_in_years] = nil
        end
      end
```

app/controllers/courses/age_range_controller.rb
TobyRet: Yeah looks like it @misaka - will fix

app/controllers/courses/age_range_controller.rb
TobyRet: @misaka ... yeah open on this I. I tend to avoid nested if / else statements but happy to go with whatever people think is most readable. An another alternative could be to extract `age_from_param.present? && age_to_param.present? && age_range_is_other?` to it's own method e.g (`custom_age_range?`)

app/controllers/courses/age_range_controller.rb
TobyRet: @misaka - spoke to Dan and this will be fixed as part of another ticket.

app/controllers/courses/age_range_controller.rb
```diff
@@ -43,6 +27,48 @@ def update
 
   private
 
+    def update_age_range_param
+      if valid_custom_age_range?
+        params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}"
+      elsif age_range_is_other?
```
TobyRet: I don't think this `elsif` branch is required anymore as it is now dealt earlier on in the process via the `build_errors` method. I will investigate the next PR that deals with the 'errors' bug that @misaka spotted

PR #735, opened 2019-10-29
[2368] Add accredited body to new courses flow

danielburnley:
### Context

Accredited bodies were added to the SD flow for new courses - the page was not implemeted yet.

This adds the page to the flow, as well as supporting non-javascript users

- Add the accredited body page

### Guidance to review

- Run through the new course flow on an SD which needs to select accredited bodies
    -  *ensure you have the latest version of MCBE running*
    - This will fail at creation due to name being blank, this will be fixed separately on MCBE

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/features/courses/accredited_body/new_spec.rb
```diff
@@ -0,0 +1,107 @@
...
+    ]
+  }
+
+  before do
+    stub_omniauth
+    stub_api_v2_resource(provider)
+    stub_api_v2_build_course
+    stub_api_v2_build_course(level: "primary")
+
+    2.times do
```
misaka: As mentioned in #708, can we `visit signin_path` instead of, essentialy, port-knocking here? :) (see https://github.com/DFE-Digital/manage-courses-frontend/pull/708#issuecomment-547426801)

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -144,6 +145,8 @@ def course_creation_path_for(page)
       new_provider_recruitment_cycle_courses_study_mode_path(path_params)
     when :applications_open
       new_provider_recruitment_cycle_courses_applications_open_path(path_params)
+    when :accredited_body
+      new_provider_recruitment_cycle_courses_accredited_body_path(path_params)
```
misaka: CodeClimate is complaining about this methods being too long. For this particular instance, I don't think we should change anything ... they're a little too long to turn into one-liners (`when ... then ...`).

app/services/next_course_creation_step_service.rb
```diff
@@ -17,6 +17,8 @@ def execute(current_step:, current_provider:)
       :location
     when :location
       handle_location(current_provider)
+    when :accredited_body
+      :entry_requirements
```
misaka: CodeClimate is complaining about this methods being too long. We could turn these into one-liners (`when ... then ...`) perhaps.

I've dismissed the complaint in CC for now anyway.

danielburnley:
Rebased + fixed comments, merging now :tada:

PR #737, opened 2019-10-31
[2435] - fix(course age range form): validation errors

TobyRet:
### Context
When a user selects a 'another age range' and does not fill in the 'From' and 'To' fields the correct errors are now appearing on the page. If an error is shown then the 'another age range' radio automatically expands in order to show errors. 

- fixes to front end error handling
- error copy updated
- increased test coverage

Validations included:
- missing 'from' and 'to' fields
- missing 'to' field
- missing 'from' field
- 'to' field greater than 'from' field
- course age range less than 4 years

### Before
age_fix_2

### After
age_validation_fix

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/age_range_controller.rb
+++ b/app/controllers/courses/age_range_controller.rb
@@ -6,7 +6,10 @@ module Courses
 
     def update
       @errors = build_errors
-      return render :edit if @errors.present?
+      if @errors.present?
+        @course.age_range_in_years = "#{age_from_param}_to_#{age_to_param}"
+        return render :edit
+      end
 
       update_age_range_param
 
@@ -28,11 +31,7 @@ module Courses
   private
 
     def update_age_range_param
-      if valid_custom_age_range?
-        params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}"
-      elsif age_range_is_other?
-        params[:course][:age_range_in_years] = nil
-      end
+      params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}" if valid_custom_age_range?
     end
 
     def valid_custom_age_range?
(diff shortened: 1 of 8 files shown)


TobyRet:
https://trello.com/c/f0xzePVI/2435-fix-age-range-validation

app/views/courses/age_range/_errors.html.erb
```diff
@@ -0,0 +1,23 @@
+<% error_summary_title ||= 'You’ll need to correct some information.' %>
+<% if @errors && @errors.any? %>
+  <div class="govuk-error-summary" aria-labelledby="error-summary-title" role="alert" tabindex="-1" data-module="govuk-error-summary" data-ga-event-form="error">
+    <h2 class="govuk-error-summary__title" id="error-summary-title">
+      <%= error_summary_title %>
+    </h2>
+    <div class="govuk-error-summary__body">
+      <ul class="govuk-list govuk-error-summary__list">
+        <% if @errors.key?(:age_range_in_years) %>
```
TobyRet: The error handling for this form is now a bit more complicated and required re-creating the error partial. If the user fails to fill in both the 'From' and 'To' fields then we only want the first of the three error messages shown ("Enter an age in both From and To")

app/views/courses/age_range/_form_fields.html.erb
```diff
@@ -12,8 +12,8 @@
   <div class="govuk-radios__divider">or</div>
   <div class="govuk-radios__item">
     <%= form.radio_button :age_range_in_years, 'other', class: 'govuk-radios__input', data: {"aria-controls" => "other-container" },
-      checked: course.other_age_range? && course.age_range_in_years.present?,
-      aria: { describedby: @errors && @errors[:age_range_in_years] ? "course_age_range_in_years-error" : '' } %>
+      checked: (course.other_age_range? && course.age_range_in_years.present?) || (@errors && (@errors[:age_range_in_years_to].present? || @errors[:age_range_in_years_from].present?)),
```
TobyRet: This is very ugly now.
@dankmitchell says this could be refactored once Rails 6.1 ships with components. In the meantime let me know if you have any short term suggestions or think it requires a comment

app/controllers/courses/age_range_controller.rb
```diff
@@ -76,7 +103,7 @@ def current_step
     def errors; end
 
     def build_course
-      @course = Course
+      @course ||= Course
```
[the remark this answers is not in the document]
TobyRet: If validation error occurs then it saves calling the db again

app/views/courses/age_range/_form_fields.html.erb
defong: probably wise just to slap a comment as a reminder

app/controllers/courses/age_range_controller.rb
```diff
@@ -28,25 +31,49 @@ def update
...
 
     def build_errors
-      if age_range_is_other? && (age_from_param.blank? || age_to_param.blank?)
-        {
-          age_range_in_years: ["Enter an age for both from and to"],
-          age_range_in_years_from: ["Enter an age"],
-          age_range_in_years_to: ["Enter an age"],
+      from_and_to_error = "Enter an age in both From and To"
+      from_error = "Enter an age in From"
+      to_error = "Enter an age in To"
```
misaka: Should we put these into `locales`?

app/views/courses/age_range/_form_fields.html.erb
misaka: How about just putting this behind an appropriately-named helper function?

app/controllers/courses/age_range_controller.rb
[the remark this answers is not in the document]
misaka: Where does it get called a second time? I think that's what might be surprising about this.

app/controllers/courses/age_range_controller.rb
```diff
@@ -28,25 +31,49 @@ def update
...
-      if valid_custom_age_range?
-        params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}"
-      elsif age_range_is_other?
-        params[:course][:age_range_in_years] = nil
-      end
+      params[:course][:age_range_in_years] = "#{age_from_param}_to_#{age_to_param}" if valid_custom_age_range?
     end
 
     def valid_custom_age_range?
       age_from_param.present? && age_to_param.present? && age_range_is_other?
```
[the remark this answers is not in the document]
TobyRet: Validation has now been added for this

app/views/courses/age_range/_form_fields.html.erb
TobyRet: Like the helper function idea as a temporary measure 👍

app/controllers/courses/age_range_controller.rb
TobyRet: Yep good spot

app/views/courses/age_range/_errors.html.erb
defong: just wonder if the `shared` error partial can we inject this little bit of html...

a bit like those `form_fields`

TobyRet: Mmm ... Not sure if that is the right thing to do as this bit of logic relates specifically to age range?

spec/features/courses/age_range_in_years/edit_spec.rb
```diff
@@ -92,14 +92,8 @@
...
       choose("course_age_range_in_years_other")
-      click_on "Save"
 
-      expect(age_range_in_years_page).to be_displayed
-      expect(age_range_in_years_page.error_flash).to have_content(
-        "You’ll need to correct some information.\nEnter an age for both from and to Enter an age Enter an age",
-      )
-
-      fill_in("course_course_age_range_in_years_other_from", with: "16")
+      fill_in("course_course_age_range_in_years_other_from", with: "14")
```
misaka: Should we use SitePrism page objects for these? It'd save on effort if/when we change the selector for these ... and kind of feels odd that we use the SitePrism stuff for some things and not others.

TobyRet:
[Fix age range validation](https://trello.com/c/f0xzePVI/2435-fix-age-range-validation)

PR #739, opened 2019-11-01
[2460] Removed controller specs

defong:
### Context
I trust @tvararu and run 
`rm -rf spec/controllers/` as  well  as 
`sudo rm -rf / `

Delete controller specs in manage-courses-frontend

### Guidance to review
¯\\_(ツ)_/¯
Maybe just delete the `provider` one...

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


defong:
[Delete controller specs in manage-courses-frontend](https://trello.com/c/YvxqlDOz/2460-delete-controller-specs-in-manage-courses-frontend)

timabell:
What's the reason for this?

dankmitchell:
> What's the reason for this?

Controller specs aren't the norm in rails projects anymore, they have been replaced with request specs.

So these need to replaced before deleting.

tvararu:
@timabell in the card I said:

```
### Context

These specs are better implemented using system/feature specs.

### Success

- `rm -rf spec/controllers/`
- Pass code review (i.e. we already have coverage elsewhere for what we're deleting)
```

Basically yeah the idea behind this is to remove the specs and have adequate coverage elsewhere :)

defong:
My feeling for this is...

> ### Guidance to review
> 
> ¯\_(ツ)_/¯
> Maybe just delete the `provider` one...

as tests are there for assurance for a much wider reasonings then...


> aren't the norm in rails projects anymore

... a framework should not be positioned to be dictatorship, as it's rather meritless and perilous...

> replaced with request specs

... one is vertical slice of testing vs the former which is a horizontal slice of testing, they are overlapping but the spectrum of purpose and intent is inherently different.

> adequate coverage elsewhere

We probably have or probably stubbed or mocked, hence just delete the `provider` one.

It is more of a team vision and perception, and deliberation.

So sending this worm of cans over to @misha for deliration.

misaka:
> Controller specs aren't the norm in rails projects anymore, they have been replaced with request specs.

> So these need to replaced before deleting.

This was originally expressed by @heidar in:

https://trello.com/c/eg40dIHP/1221-move-controller-tests-to-requests-tests

TobyRet:
@defong Maybe it could be worth adding code coverage stats (before and after changes) in order to alleviate any concerns?

defong:
in the end removed
``` 
spec/controllers/providers_controller_spec.rb
spec/controllers/sessions_controller_spec.rb
spec/controllers/users_controller_spec.rb
```

as it covered in either `feature` or `request` specs

and retained 

```
spec/controllers/application_controller_spec.rb
```
as the test covers an actual method which all derived controllers inherits from

```
spec/controllers/errors_controller_spec.rb
```
as it the catch all for where all the other controllers may use
and most noticeably `internal_server_error` which a component of our system may cause

PR #744, opened 2019-11-06
[2492] - Run specs in parallel on CI

TobyRet:
### Context
Part of a bigger section of work to run specs in parallel.

A simple change to the azure configuration so that specs are run in parallel on CI via the `parallel_tests` gem. Azure has _two_ available cores that can be utilised.

### How much time is saved?

_Before_
min ~50 seconds

_After_
40 seconds

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -9,7 +9,7 @@ Rake::Task["default"].clear
 
 task :default do
   cores = ENV['PARALLEL_CORES']
-  Rake::Task["parallel"].invoke(cores)
+  Rake::Task["parallel:spec"].invoke(cores)
   Rake::Task["lint:ruby"].invoke
   Rake::Task["lint:scss"].invoke
 end
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,7 +41,7 @@ steps:
     runInBackground: false
 
 - bash: |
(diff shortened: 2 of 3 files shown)


TobyRet:
[Running test-suites in parallel on CI: Manage-frontend](https://trello.com/c/suvNb4QT/2492-running-test-suites-in-parallel-on-ci-manage-frontend)

README.md
```diff
@@ -60,6 +60,8 @@ bundle exec rails parallel:spec
 bundle exec rails parallel:spec[number]
 ```
 
+Specs are run in parallel on CI (2 cores).
```
defong: @rizzkhan1 can we get more cores?

timabell: this statement is likely to become stale, I'm not sure I'd add this line personally

TobyRet: @defong - so apparently we have no control of the underlying hardware for an Azure agent. Another way we could potentially see further time saving is to utilise more agents however having discussed with @misaka , this will make setting up parallel testing a lot more complicated (more dbs (or namespaces?) will be required) and may not be worth the additional effort. Will be interesting to see what the improvement is on backend CI utilising 2 cores

TobyRet: I've removed the update to README @timabell

PR #746, opened 2019-11-06
Remove sprockets

dankmitchell:
### Context
Sprockets are causing a Rails 6.0.1 upgrade to fail (https://github.com/DFE-Digital/manage-courses-frontend/pull/743). We don't need to use sprockets anyway.

Remove sprockets

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


config/webpacker.yml
```diff
@@ -11,7 +11,7 @@ default: &default
 
   # Additional paths webpack should lookup modules
   # ['app/assets', 'engine/foo/app/assets']
-  resolved_paths: []
+  resolved_paths: ["node_modules/govuk-frontend/govuk"]
```
TobyRet: @dankmitchell - does adding the resolved path render the CopyWebpackPlugin useless (hence why it was removed?)

dankmitchell: Correct

PR #754, opened 2019-11-12
[2530] - compile Webpack once only when running specs in parallel

TobyRet:
### Context
When running specs in parallel we want to avoid workers compiling the same files at the same time. Believe this was the reason for this bug 👉 https://trello.com/c/rVx1gcbf/2530-investigate-puma-upgrade-issue

- Add a spec helper that compiles assets only once when running specs in parallel. It does this by creating a hexdigest of all webpacker assets and compiling only when this value changes.
- Run specs in parallel on CI (again!)

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -41,7 +41,7 @@ steps:
     runInBackground: false
 
 - bash: |
-    docker run --rm $(IMAGE_NAME) rails spec SPEC_OPTS='--format RspecJunitFormatter' >rspec-results.xml
+    docker run --rm $(IMAGE_NAME) rails parallel:spec SPEC_OPTS='--format RspecJunitFormatter' >rspec-results.xml
     test_result=$?
     cat rspec-results.xml
     sed -i '$d;1,5d' rspec-results.xml
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -32,14 +32,6 @@ steps:
     arguments: '--cache-from $(IMAGE_NAME):latest'
     addDefaultLabels: false
 
-- task: Docker@1
-  displayName: Run webpack
-  inputs:
-    command: Run an image
-    imageName: $(IMAGE_NAME)
-    containerCommand: rails webpacker:compile
-    runInBackground: false
-
 - bash: |
     docker run --rm $(IMAGE_NAME) rails parallel:spec SPEC_OPTS='--format RspecJunitFormatter' >rspec-results.xml
     test_result=$?
--- a/spec/support/webpacker.rb
+++ b/spec/support/webpacker.rb
@@ -1,0 +1,36 @@
+module WebpackerTest
+  TIMESTAMP_FILE = Rails.root.join("tmp", "webpacker-#{Rails.env}-timestamp")
+
+  def self.compile_once
+    digest_file = Rails.root.join("tmp/webpacker_#{Rails.env}_digest")
+
+    app_code = Dir[Webpacker.config.source_path.join("**/*")]
+    npm_code = Dir[Rails.root.join("yarn.lock")]
+
+    packable_contents = (app_code + npm_code)
+      .sort
+      .map { |filename| File.read(filename) if File.file?(filename) }
+      .join
+    digest = Digest::SHA256.hexdigest(packable_contents)
+
+    return if digest_file.exist? && digest_file.read == digest
+
+    if ENV["TEST_ENV_NUMBER"].to_i < 1
+      public_output_path = Webpacker.config.public_output_path
+      FileUtils.rm_r(public_output_path) if File.exist?(public_output_path)
+      puts "Removed Webpack output directory #{public_output_path}"
+
+      Webpacker.compile
+
+      digest_file.write(digest)
+    else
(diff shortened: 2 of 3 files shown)


TobyRet:
[Investigate Puma upgrade issue](https://trello.com/c/rVx1gcbf/2530-investigate-puma-upgrade-issue)

spec/support/webpacker.rb
```diff
@@ -0,0 +1,36 @@
...
+      public_output_path = Webpacker.config.public_output_path
+      FileUtils.rm_r(public_output_path) if File.exist?(public_output_path)
+      puts "Removed Webpack output directory #{public_output_path}"
+
+      Webpacker.compile
+
+      digest_file.write(digest)
+    else
+      loop do
+        break if digest_file.exist? && digest_file.read == digest
```
misaka:
```suggestion
        break if digest_file.exist? && Digest::SHA256.hexdigest(digest_file.read) == packable_contents
```

spec/support/webpacker.rb
```diff
@@ -0,0 +1,36 @@
...
+    digest_file = Rails.root.join("tmp/webpacker_#{Rails.env}_digest")
+
+    app_code = Dir[Webpacker.config.source_path.join("**/*")]
+    npm_code = Dir[Rails.root.join("yarn.lock")]
+
+    packable_contents = (app_code + npm_code)
+      .sort
+      .map { |filename| File.read(filename) if File.file?(filename) }
+      .join
+    digest = Digest::SHA256.hexdigest(packable_contents)
```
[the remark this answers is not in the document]
TobyRet: Hopefully I am understanding the question correctly :-) ...

The reason for checking if the digest has changed is to take into account any changes to assets since the digest was last created. If we do not check if the digest had changed then we would never recompile. At the same time we only want one process to compile or recompile if required.
This problem is perhaps more pertinent on CI as we will will always have to compile from scratch. It appears weird issues occur if you have multiple processes compiling the same files at the same time. This is what we want to avoid.

spec/support/webpacker.rb
misaka: Thinking about this some more on the way home, I'm not so sure this is any more efficient then what you have. Please disregard.

spec/support/webpacker.rb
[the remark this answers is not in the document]
misaka: My point is that if there are _no_ changes in the assets, `Webpacker.compile` doesn't do anything, as evidenced by the fact that it returns in less than 1/10th of a second (see benchmarks). If something is changed in the assets, I tested by adding some whitespace to one of the JS files, then `Webpacker.compile` takes 10s to, presumably, do the compilation.

So I wonder if all that code to check if anything has changed is actually necessary.

misaka:
Did a bit more thinking about this, and a few points came to mind:

1. I'm not sure this is an issue in CI because we [precompile assets when we build the Docker image](https://github.com/DFE-Digital/manage-courses-frontend/blob/master/Dockerfile#L28). Correct me if I'm wrong, but my understanding is that that runs the task `webpacker:compile`. Even if not, as @defong points out in Slack,  we have a (seemingly redundant) task in Azure Devops to run Webpacker.
2. If we need to only do this once, how about we have either `webpacker:compile` or `assets:precompile` as a dependency to the `parallel:spec` and/or the `spec` Rake tasks?
3. If we really need to compile Webpacker manually like this, then can we use a more traditional locking mechanism? Ruby has `File#flock`, or we could touch/symlink a "lock" file. I think this would be a bit clearer, and since `Webpack.compile` already detects whether it needs to re-compile or not I don't think we need the extra code that's currently there.

Let me know if I haven't managed to explain myself clearly, or if I'm missing something. :)

TobyRet:
Let's discuss in person

TobyRet:
@misaka @defong 
The plot thickens ...
So looking here at the webpacker config, you can see that `compile` has been set to true. 
https://github.com/DFE-Digital/manage-courses-frontend/blob/master/config/webpacker.yml#L80

If you set this to false this would work on CI as there is a `compile` step beforehand however this will fall over locally (unless assets are manually compiled). 

Once option could be to set the `public_output_path` for the test environment to something like `  public_output_path: packs-test` 

... or to be honest, given we only have 2 cores on CI, maybe the there is little benefit to running in parallel and messing with the config?

Wonder if @dankmitchell has any thoughts?

TobyRet:
@defong @misaka 
- Unnecessary webpacker compile build step has been removed.
- See also that webpacker config has been updated for testing. Compile is now set to false and support/webpacker.rb assumes responsibility for compilation (once only)

azure-pipelines.yml
```diff
@@ -32,16 +32,8 @@ steps:
     arguments: '--cache-from $(IMAGE_NAME):latest'
     addDefaultLabels: false
 
-- task: Docker@1
```
defong: @rizzkhan1 do we need to do something when removing a task?

rizzkhan1: No. This looks okay

PR #764, opened 2019-11-14
[2370] Hook up back links for new course flow

danielburnley:
preview

### Context

When going through the new course flow, the user should be able to see and use a `back` link to go to their previous steps, with the following circumstances:

- Going through the flow normally
- Going back to the confirmation page
- Going back past locations if there is only a single location
- Going back to different pages depending on the user being an accredited body

- Add in correct back links
- Add in view helper for doing back links with confirmation

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/accredited_body_controller.rb
+++ b/app/controllers/courses/accredited_body_controller.rb
@@ -2,7 +2,6 @@ module Courses
   class AccreditedBodyController < ApplicationController
     before_action :build_course_params, only: :continue
     include CourseBasicDetailConcern
-    before_action :build_back_link, only: :new
 
     decorates_assigned :provider
 
@@ -106,14 +105,6 @@ module Courses
     def build_course_params
       @accredited_body = params[:course].delete(:accredited_body)
       @autocompleted_provider_code = params[:course].delete(:autocompleted_provider_code)
-    end
-
-    def build_back_link
-      @back_link_path = if @provider.sites.count > 1
-                          new_provider_recruitment_cycle_courses_locations_path(course: @course_creation_params)
-                        else
-                          new_provider_recruitment_cycle_courses_study_mode_path(course: @course_creation_params)
-                        end
     end
 
     def errors_for_search_query(code, query)
--- a/app/controllers/courses/accredited_body_controller.rb
+++ b/app/controllers/courses/accredited_body_controller.rb
@@ -2,6 +2,7 @@ module Courses
   class AccreditedBodyController < ApplicationController
     before_action :build_course_params, only: :continue
     include CourseBasicDetailConcern
+    before_action :build_back_link, only: :new
 
     decorates_assigned :provider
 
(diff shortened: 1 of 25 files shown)


danielburnley:
https://trello.com/c/M6J2ViCg/2370-s-hook-up-back-links-for-new-course-flow

misaka:
Feels like a good candidate to have a quick gif demonstrating how this works. @TobyRet what were you using to do that, again?

danielburnley:
now with added gif @misaka (OBS -> ffmpeg to gif is a way to do it on linux, apparently)

app/controllers/courses/accredited_body_controller.rb
```diff
@@ -99,6 +108,14 @@ def build_course_params
       @autocompleted_provider_code = params[:course].delete(:autocompleted_provider_code)
     end
 
+    def build_back_link
+      @back_link_path = if @provider.sites.count > 1
+                          new_provider_recruitment_cycle_courses_locations_path(course: @course_creation_params)
+                        else
+                          new_provider_recruitment_cycle_courses_study_mode_path(course: @course_creation_params)
+                        end
+    end
```
misaka: This `build_back_link`, and all the other versions of it in the other controllers, is doing my head in. That is, until I have a look at `NextCourseCreationStepService`.

I feel like we should have a single service that handles both `next` and `back` ... I know that breaks the principle of having an `execute` as a single point of entry for services ... maybe we could have `Providers::CourseCreation::NextStepService` and `Providers::CourseCreation::BackService`, with the commonalities kept in `Providers::CourseCreation`? That might be messy if we have to define `NextStepService` and `BackService` in a separate Ruby file, which would be necessary for the auto-loader to work (and Rubocop not to complain, probably).

Not sure what's the best approach, but I do think that this needs a bit of cleanup, it's not clear what the logic is doing here.

defong: Agreed, its doing my head as well.

``` ruby
def oscillate(current, course) do
{
  previous: previous_link_factory,
  next: next_link_factory,
  roulette_spin: lucky_day 
}
end

```

misaka:
> now with added gif @misaka (OBS -> ffmpeg to gif is a way to do it on linux, apparently)

How much work is it?

app/controllers/courses/accredited_body_controller.rb
danielburnley: Yeah I'll make a service first, then I'll look at some common interface for getting both the next and back links :thinking: I did avoid making a service for this reason, it's a bit of a headache to get a ncie way to do it but i'll dedicate some time to it

faissalMT:
OBS is straightforward enough, ffmpeg might be a bit fiddly though https://engineering.giphy.com/how-to-make-gifs-with-ffmpeg/

misaka:
Ran into a small problem: accredited bodies aren't re-selected when I go back to them from a later page.

Steps:

1. navigate to provider 2HN's course page
2. create a new course (by append /new to the URL)
3. when on the accredited bodies select an existing provider as the accredited provider
4. click "Continue"
5. use the "Back" link to go back to the previous page
6. :boom: the previously-selected accredited body is no longer selected

Notably, when using the browser back-button this persists.

Unless this is trivial to fix, we should probably just deploy this ticket as-is and create a new bug ticket for this.

PR #768, opened 2019-11-18
[2484] Add inform publisher page

faissalMT:
### Context
User support needs to be able to inform publishers when their access is changed.

image

Add informing publishers to the access requests workflow.

### Guidance to review
This workflow has a lot of potential for automation, but this should suffice for migrating the existing functionality from the support app.

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history


faissalMT:
[Move AccessRequests functionality](https://trello.com/c/0FsaLfP1/2484-move-accessrequests-functionality)

app/helpers/notify_helper.rb
```diff
@@ -0,0 +1,13 @@
+module NotifyHelper
+  def notify_service_link
+    "https://www.notifications.service.gov.uk/services/022acc23-c40a-4077-bbd6-fc98b2155534"
```
misaka: Feels like this, and the below template paths, should be settings.

app/helpers/dfe_signin_helper.rb
```diff
@@ -0,0 +1,5 @@
+module DfeSigninHelper
+  def dfe_signin_email_search(email)
+    "https://support.signin.education.gov.uk/users?criteria=#{email}"
```
misaka: Shouldn't this be a setting?

app/views/access_requests/inform_publisher.html.erb
```diff
@@ -6,15 +6,15 @@
   Inform the publisher about the change to their access:
 
   <ol class="govuk-list govuk-list--number">
-    <li>Check whether <%= link_to "they already have a DfE Sign-in account", dfe_signin_email_search(@recipient_email_address), class: "govuk-link" %></li>
-    <li>In <%= link_to "GOV.UK Notify", notify_service_link, class: 'govuk-link' %>, send an email to <strong><%= @recipient_email_address %></strong> using:
+    <li>Check whether <%= link_to "they already have a DfE Sign-in account", dfe_signin_email_search(@access_request.email_address), class: "govuk-link", data: { qa: "dfe_signin_search_link" } %></li>
+    <li>In <%= link_to "GOV.UK Notify", notify_service_link, class: 'govuk-link', data: { qa: "notify_service_link" } %>, send an email to <strong><%= @recipient_email_address %></strong> using:
```
misaka: Is it really sufficient for us to just provide a link to Notify? Then the admin has to copy/paste or somehow manually input the email address, which is always going to be risky. Where did the decision that we implement it this way get made?

faissalMT: This is the implementation as it is in the support console, as I understood it the priority was to get the functionality moved in order to decommission the support app asap and that improvements would be done after the fact.

PR #773, opened 2019-11-19
[2412] Add further education workflow to NextCourseCreationStepService

faissalMT:
### Context
We need to implement the further education workflow to make course creation useful for a wider range of users.

Refactor NextCourseCreationStepService to make workflows easier to read and adds the further education workflow.

### Guidance to review
Requires: https://github.com/DFE-Digital/manage-courses-backend/pull/1021

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/services/next_course_creation_step_service.rb
+++ b/app/services/next_course_creation_step_service.rb
@@ -1,48 +1,65 @@
 class NextCourseCreationStepService
-  def execute(current_step:, current_provider:)
-    case current_step
-    when :level
-      :subjects
-    when :subjects
-      :age_range
-    when :age_range
-      :outcome
-    when :outcome
-      handle_outcome(current_provider)
-    when :fee_or_salary
-      :full_or_part_time
-    when :apprenticeship
-      :full_or_part_time
-    when :full_or_part_time
-      :location
-    when :location
-      handle_location(current_provider)
-    when :accredited_body
-      :entry_requirements
-    when :entry_requirements
-      :applications_open
-    when :applications_open
-      :start_date
-    when :start_date
-      :confirmation
+  def execute(current_step:, course:)
+    workflow_steps = get_workflow_steps(course)
+    get_next_step(workflow_steps, current_step)
+  end
+
+  def get_next_step(steps, current_step)
+    steps[steps.find_index(current_step).next]
+  end
+
+  def get_workflow_steps(course)
+    if course.is_further_education?
+      further_education_steps
+    elsif course.is_uni_or_scitt?
+      uni_or_scitt_workflow_steps
+    elsif course.is_school_direct?
+      school_direct_workflow_steps
     end
   end
 
-private
-
-  def handle_outcome(provider)
-    if provider.accredited_body?
-      :apprenticeship
-    else
-      :fee_or_salary
-    end
+  def school_direct_workflow_steps
+    %i[
+      level
+      subjects
+      age_range
+      outcome
+      fee_or_salary
+      full_or_part_time
+      location
+      accredited_body
+      entry_requirements
+      applications_open
+      start_date
+      confirmation
+    ]
   end
 
-  def handle_location(provider)
-    if provider.accredited_body?
-      :entry_requirements
-    else
-      :accredited_body
-    end
+  def uni_or_scitt_workflow_steps
+    %i[
+      level
+      subjects
+      age_range
+      outcome
+      apprenticeship
+      full_or_part_time
+      location
+      entry_requirements
+      applications_open
+      start_date
+      confirmation
+    ]
+  end
+
+  def further_education_steps
+    %i[
+      level
+      outcome
+      full_or_part_time
+      location
+      applications_open
+      start_date
+      confirmation
+    ]
   end
 end
--- a/spec/features/courses/new_spec.rb
+++ b/spec/features/courses/new_spec.rb
@@ -44,7 +44,7 @@ feature "new course", type: :feature do
   let(:course) do
     model = build(:course,
                   :new,
-                  level: :primary,
+                  level: level,
                   provider: provider,
                   course_code: "A123",
                   content_status: "draft",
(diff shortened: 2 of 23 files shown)


faissalMT:
[[M] Create "create course flow" for Further Ed providers](https://trello.com/c/rDI191OY/2412-m-create-create-course-flow-for-further-ed-providers)

app/services/next_course_creation_step_service.rb
```diff
@@ -1,48 +1,64 @@
...
-      :entry_requirements
-    when :entry_requirements
-      :applications_open
-    when :applications_open
-      :start_date
-    when :start_date
-      :confirmation
-    end
+  def execute(current_step:, current_provider:, level:)
+    get_next_step(get_workflow(current_provider, level), current_step)
```
misaka: I find this a little bit hard to read, and unfortunately I don't think pushing it into separate methods helps. How about:

```ruby
  def execute(current_step_name:, current_provider:, level:)
    steps = get_workflow(current_provider, level)
    current_step = steps.find_index(current_step_name)
    steps[current_step.next]
  end
```

You might even consider renaming `get_workflow` to `get_workflow_steps`.

app/services/next_course_creation_step_service.rb
```diff
@@ -1,48 +1,64 @@
...
-    when :accredited_body
-      :entry_requirements
-    when :entry_requirements
-      :applications_open
-    when :applications_open
-      :start_date
-    when :start_date
-      :confirmation
-    end
+  def execute(current_step:, current_provider:, level:)
```
misaka: How about passing `course` through instead of `level` and `provider`? I don't see why the caller should be concerned with what part of the course is needed to determine what the next step is, and we could even forseeably require some other part of the `course` one day, which would require another change to the interface here.

app/services/next_course_creation_step_service.rb
```diff
@@ -1,48 +1,64 @@
...
+        outcome
+        fee_or_salary
+        full_or_part_time
+        location
+        accredited_body
+        entry_requirements
+        applications_open
+        start_date
+        confirmation
+      ]
```
misaka: Not that I think we should jump on every one of CodeClimate's concerns, but these lists seem like they could very nicely be broken up into methods with `get_workflow` concerned with which one to return.

app/services/next_course_creation_step_service.rb
```diff
@@ -1,48 +1,64 @@
...
-    else
-      :accredited_body
-    end
+  def is_school_direct?(provider, level)
+    !(is_uni_or_scitt?(provider) || is_further_education?(level))
+  end
+
+  def is_uni_or_scitt?(provider)
+    provider.accredited_body?
+  end
```
misaka: Is this the right place for this logic? I'd be really keen to keep this in the model so we can hide this rather messy logic, including `is_school_direct` above.

spec/features/courses/new_spec.rb
```diff
@@ -163,84 +221,84 @@ def save_course
     confirmation_page.save.click
   end
 
-  def select_level(course_creation_params)
-    course_creation_params[:level] = "primary"
+  def select_level(course_creation_params, level_selection:, next_page:)
+    course_creation_params[:level] = level
```
misaka: I have to say, I'm a bit iffy about the use of a memoized value in a `let` here in the method. It works ... and this is the way the `it` blocks do it, but this is a function and it feels like this should be a parameter.

spec/services/next_course_creation_step_service_spec.rb
```diff
@@ -34,9 +36,31 @@
 
       include_examples "next step"
     end
+
+    context "Current step: Accredited body" do
+      let(:current_step) { :accredited_body }
+      let(:expected_next_step) { :entry_requirements }
+
+      include_examples "next step"
```
misaka: Shared examples can take examples. Can we use that here? That would add a bit of security around calling them.

```suggestion
      include_examples "next step", :accredited_body, :entry_requirements
```

app/services/next_course_creation_step_service.rb
faissalMT: I'd like to avoid giving a service much more information than it makes it harder to see what exactly it depends on.

app/services/next_course_creation_step_service.rb
faissalMT: This logic seems tied particularly to this next course creation functionality, I don't think we should move it out until there is demand for it. If there was a need to know this information elsewhere I'd prefer to have it in the backend rather than infering it from sparse properties.

app/services/next_course_creation_step_service.rb
faissalMT: How is it now? I still think pushing it into separate methods helps a bit so I kept that but got rid of the nesting and renamed `get_workflow`

app/services/next_course_creation_step_service.rb
misaka: As discussed IRL, I think that `course` is the right level to parameterise this service on, given that it has `Course` in the name and it's specifically about finding the next step for a course.

app/services/next_course_creation_step_service.rb
misaka: I'm worried about divergent implementations of these two methods. I think that given that these are hidden away in a service, if anyone has to implement `is_uni_or_scitt?` and `is_school_direct?` logic again, they won't think to look here, which is why I'd rather this be in the model.

defong: isn't this really that?

https://github.com/DFE-Digital/manage-courses-backend/blob/085b2fc4a789c1cf75bce898f6cba90f869d0581/app/models/provider.rb#L50-L56

defong:
need more test code coverage or laxative affects are required @faissalMT

PR #776, opened 2019-11-20
[2570] Fix course create crashing when a provider has no courses

faissalMT:
### Context
Course creation was crashing due to the course creation controller pulling the provider details from the first course.

Pull the provider details independently to avoid dependence on an existing course and fix broken tests.

### Guidance to review
Given that this test causes me to go over a number of tests I've taken the liberty of refactoring some older stubs.

### Checklist

- [X] Attach to Trello card
- [x] Rebased master
- [X] Cleaned commit history

### DIFF
--- a/app/controllers/courses_controller.rb
+++ b/app/controllers/courses_controller.rb
@@ -5,9 +5,9 @@ class CoursesController < ApplicationController
   before_action :build_courses, only: %i[index about requirements fees salary]
   before_action :build_course, except: %i[index preview]
   before_action :build_course_for_preview, only: :preview
-  before_action :build_provider, except: %i[index confirmation]
   before_action :filter_courses, only: %i[about requirements fees salary]
   before_action :build_copy_course, if: -> { params[:copy_from].present? }
+  before_action :build_provider_from_provider_code, except: %i[index]
 
   def index; end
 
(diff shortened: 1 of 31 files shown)


faissalMT:
[Add modern language selection to course creation](https://trello.com/c/P2E8mcJv/2566-add-modern-language-selection-to-course-creation)

faissalMT:
[Create courses errors when provider has no courses](https://trello.com/c/ZAXmJlAL/2570-create-courses-errors-when-provider-has-no-courses)

app/controllers/courses_controller.rb
```diff
@@ -5,15 +5,18 @@ class CoursesController < ApplicationController
   before_action :build_courses, only: %i[index about requirements fees salary]
   before_action :build_course, except: %i[index preview]
   before_action :build_course_for_preview, only: :preview
-  before_action :build_provider, except: %i[index confirmation]
```
misaka: Why remove the action?

misaka: Ah, I see that `build_provider` is essentially being replaced by `build_provider_from_provider_code`. I think we should use `build_provider_from_provider_code` as a `before_action` ... there's validity to being explicit and running that in the actions that need it, but we _are_ using `before_actions` in our controllers, so should keep consistent.

PR #795, opened 2019-12-02
[2620] Add static maintenance page for offline mode

dankmitchell:
### Context
Offline maintenance page

> https://github.com/alphagov/govuk-design-system-backlog/issues/129

Add a hardcoded static maintenance page

### Guidance to review
`/maintenance`

### After
localhost_3001_maintenance html(Laptop with MDPI screen)

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/public/maintenance.html
+++ b/public/maintenance.html
@@ -1,0 +1,706 @@
+<!DOCTYPE html>
+<!-- app_offline.htm -->
+<html lang="en" class="govuk-template">
+  <head>
+    <meta charset="utf-8" />
+    <title>Publish teacher training courses - GOV.UK</title>
+    <meta name="viewport" content="width=device-width, initial-scale=1" />
+    <meta name="theme-color" content="#0b0c0c" />
+
+    <link rel="shortcut icon" href="https://unpkg.com/govuk-frontend@3.4.0/govuk/assets/images/favicon.ico" type="image/x-icon" asp-append-version="true" />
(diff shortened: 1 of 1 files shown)


dankmitchell:
[Build a maintenance page for PUBLISH](https://trello.com/c/NLOVyxwU/2620-build-a-maintenance-page-for-publish)

misaka:
Should this be in the app? If our app isn't running then this won't be served out ...

I thought we were going to have a separate app, maybe even an NGinx Docker image, that would serve out a single static page when our app is down.

dankmitchell:
@misaka this PR, just creates the static HTML page. What we do with it, where we store and how we use it is still up for debate cc. @rizzkhan1  @defong

defong:
Feels more of an infrastructure thing more than anything else

I leave it @rizzkhan1 to cast judgement.

This static page needs to be stored some where, served somehow, and triggered some what.

misaka:
I've got an nginx container that will serve out a single file no matter what the URL, and log the request to logit. It's not codified into a `Dockerfile` yet, but I've got the steps to recreate so will have a look at doing that. This will be runable as an Azure webapp.

defong:
So nginx going to live in here
https://github.com/DFE-Digital/manage-courses-backend/blob/0ade37723d40979402bff0c19d28b493b8a33ecd/Dockerfile

or some where else? if so can i reject this please..

misaka:
> So nginx going to live in here

It'll be a different `Dockerfile`. You can request changes to whatever you like.

PR #791, opened 2019-11-28
[2566] Add modern languages to course creation

faissalMT:
### Context
As it stands standard subjects can be added during course creation but this neglects modern language subjects which require an additional page and set of functionality.

### Changes proposed in this pull request
Add the modern language subjects page to course creation and insert it into the workflow. Also adds the back endpoint to decide whether or not to skip the step when going backwards.

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally

PR #798, opened 2019-12-04
[2631] Make the back button carry out the correct workflow for Further Education

faissalMT:
### Context
The further education workflow currently does not implement the back button correctly, instead sending you to an arbitrarily chosen page or perhaps erroring rather than back through the workflow.  

CourseCreationStepService now handles skipping locations if the provider only has one site in both directions. Also merges PreviousCourseCreationStepService and NextCourseCreationService to avoid having multiple sources of truth for the workflow.

### Guidance to review
It might be a good idea to pull all the workflows out into another service to avoid clutter. Requires: https://github.com/DFE-Digital/manage-courses-backend/pull/1033  

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


faissalMT:
[Implement back button for further ed providers](https://trello.com/c/39EnNzVs/2631-implement-back-button-for-further-ed-providers)

app/services/course_creation_step_service.rb
```diff
@@ -0,0 +1,139 @@
...
+    workflow_steps = get_workflow_steps(course)
+
+    {
+      next: get_next_step(workflow_steps, current_step),
+      previous: get_previous_step(workflow_steps, current_step),
+    }
+  end
+
+  def get_next_step(steps, current_step)
+    steps[steps.find_index(current_step).next]
```
danielburnley: I find this really hard to parse in my head because of the nesting going on

```
next_index = ...
steps[next_index]
```

Might be a bit nicer

app/services/course_creation_step_service.rb
```diff
@@ -0,0 +1,139 @@
...
+  end
+
+  def get_previous_step(steps, current_step)
+    steps[steps.find_index(current_step).pred]
```
danielburnley: See above, but previous rather than next :eyes:

app/services/course_creation_step_service.rb
```diff
@@ -0,0 +1,139 @@
...
+    if course.is_further_education?
+      further_education_workflow_steps(course)
+    elsif course.is_uni_or_scitt?
+      uni_or_scitt_workflow_steps(course)
+    elsif course.is_school_direct?
+      school_direct_workflow_steps(course)
+    end
+  end
+
+  def school_direct_workflow_steps(course)
```
danielburnley: So - I'm gonna compare this with how we've managed stuff in #791. Are we gonna do a similar thing with modern languages? 

Feels weird that one has the controller decide, and one has here decide. I think it probably _should_ be handled in here though :thinking:

app/services/course_creation_step_service.rb
```diff
@@ -0,0 +1,139 @@
...
+    if course.provider.sites && course.provider.sites.count > 1
+      school_direct_workflow_with_locations_steps
+    else
+      school_direct_workflow_without_locations_steps
+    end
```
danielburnley: Related to above - I find this duplication of all the steps quite hard to follow, and it will probably be a pain if we ever want to alter any workflow

Could you instead get the full workflow, and remove `:locations` from it, explicitly showing that we remove it from the workflow?

danielburnley: Follow up - Locations does actually do something when you hit it, it assigns the site ID with the only one it finds. This appears to break that behaviour as it now skips that logic (Side note, we should have a test that covers this)
 
See: https://github.com/DFE-Digital/manage-courses-frontend/blob/master/app/controllers/courses/sites_controller.rb#L10-L15

app/services/course_creation_step_service.rb
faissalMT: The controllers should decide whether or not to skip a step in a given workflow, I've now changed things to reflect that

app/controllers/concerns/course_basic_detail_concern.rb
```diff
@@ -162,6 +162,14 @@ def build_back_link
     @back_link_path = previous_path
   end
 
+  def course_back_path_for(page)
```
danielburnley: oof - Grim. But i guess this works for now

danielburnley: I do like that it exposes the exceptions here, I think

PR #800, opened 2019-12-04
[2590] Confirmation page updates

danielburnley:
### Context

- Updates to the confirmation page to bring it in line with the prototype

course_confirmation

- Add a change link for subjects (and support the confirmation redirect)
- Update applications open to display a user friendly message
- Display fee or salary / apprenticeship dependent on the provider type
- Update the title for the confirmation page
- Display accredited body information where appropriate
- Display a success message when a course is created

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/confirmation.html.erb
+++ b/app/views/courses/confirmation.html.erb
@@ -175,13 +175,13 @@
               </dd>
             </div>
 
-            <% if !@provider.accredited_body? %>
+            <% unless @provider.accredited_body? %>
               <div class="govuk-summary-list__row">
                 <dt class="govuk-summary-list__key">
                   Accredited body
                 </dt>
                 <dd class="govuk-summary-list__value" data-qa="course__accredited_body">
-                  <%= course.accrediting_provider&.provider_name %>
+                  <%= course.accrediting_provider.provider_name %>
                 </dd>
                 <dd class="govuk-summary-list__actions">
                   <%= course_creation_change_button(
--- a/app/views/courses/confirmation.html.erb
+++ b/app/views/courses/confirmation.html.erb
@@ -162,7 +162,11 @@
                 Applications open
               </dt>
               <dd class="govuk-summary-list__value" data-qa="course__application_open_from">
+              <% if course.applications_open_from == @recruitment_cycle.application_start_date %>
+                As soon as the course is on Find (recommended)
+              <% else %>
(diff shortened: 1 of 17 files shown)


app/views/courses/confirmation.html.erb
```diff
@@ -153,12 +175,34 @@
               </dd>
             </div>
 
+            <% if !@provider.accredited_body? %>
```
faissalMT:
```suggestion
            <% unless @provider.accredited_body? %>
```

app/views/courses/confirmation.html.erb
```diff
@@ -153,12 +175,34 @@
...
+              <div class="govuk-summary-list__row">
+                <dt class="govuk-summary-list__key">
+                  Accredited body
+                </dt>
+                <dd class="govuk-summary-list__value" data-qa="course__accredited_body">
+                  <%= course.accrediting_provider&.provider_name %>
```
faissalMT: If it's not an accredited body it should have an accrediting provider right? So why not 
```suggestion
                  <%= course.accrediting_provider.provider_name %>
```

app/views/layouts/_flash.html.erb
```diff
@@ -5,7 +5,12 @@
         <% if key == "error" %>
           <h3 class="govuk-<%= key %>-summary__title" id="<%= key %>-summary-heading" data-error-message="<%= value %>"><%= value %></h3>
         <% else %>
-          <h3 class="govuk-<%= key %>-summary__title" id="<%= key %>-summary-heading"><%= value %></h3>
+          <% if value.is_a?(Hash) %>
```
faissalMT: Typechecking in a dynamic language generally feels like an anti-pattern to me, could this warrant a separate template for errors we might have manually set rather than forwarded from the backend? Or is it the case that the backend sometimes provides data in this format?

danielburnley: These are flash messages that we manually create by setting them in `flash` (The in built Rails thing). Typically we only support a single string, and without updating them all to hashes I added this here. Not sure how to add this functionality  

Wonder if there's a nicer approach though :thinking: given I'll try find a way of adding this

defong: maybe 
``` ruby
if (value.dig(:title).present? || value.dig(body).present?)
```

or change the `key` to something else

PR #805, opened 2019-12-06
[2621] create docker image to serve out maintenance page

misaka:
### Context

Create a docker image that will serve out the maintenance page and log requests to Logit.

Add `Dockerfile` and required config files for NGinx and Filebeat

### Guidance to review

- Is moving the `maintenance.html` page here the right thing to do?
- See me if you want to give this a run locally, but you're not sure where to get the logit logstash host and port values from.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


misaka:
[Setup maintenance page for PUBLISH](https://trello.com/c/7fbsueBS/2621-setup-maintenance-page-for-publish)

maintenance/Dockerfile
```diff
@@ -0,0 +1,23 @@
+FROM nginx
```
defong: Can we use the alpine one?

misaka: Tried that. It results in an image that's about 95MB vs 197MB ... so that's not bad, but then the `filebeat` module doesn't include the `nginx` module, and I think `logrotate` isn't fully configured. I ended up deciding it wasn't worth the effort.

PR #806, opened 2019-12-08
[2576] Move org support page to publish

davidgisbey:
### Context

WIP in moving the organisation support page to publish.

What has been done so far 

- Port over functionality from the support app
- Add a link to the header

Still to be completed

- Tests

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


spec/support/helpers.rb
```diff
@@ -191,6 +191,8 @@ def url_for_resource_collection(resource)
       url_for_resource(resource.provider) + "/courses"
     elsif resource.is_a? AccessRequest
       "/access_requests"
+    elsif resource.is_a? Organisation
+      "/organisations"
```
misaka: We should raise an exception here to make it more obvious where we've missed adding something to this rickety framework.

spec/support/serializers/organisation_serializer.rb
```diff
@@ -3,6 +3,7 @@ class OrganisationSerializer < JSONAPI::Serializable::Resource
   has_many :organisation_users
 
   has_many :users, through: :organisation_users
+  has_many :providers
```
misaka: The requirement for this is not-so-obvious. One idea for making it more obvious is to add a validation to `stub_api_v2_resource`, `stub_api_v2_resource_collection`, etc, that would check that any included resources show up in the generated `jsonapi_response`, and raise an error if it's missing. This would imply that it's missing from the serialiser here.

app/helpers/dfe_signin_helper.rb
```diff
@@ -0,0 +1,9 @@
+module DfeSigninHelper
+  def dfe_signin_user_audit_link(user)
+    "https://support.signin.education.gov.uk/users/#{user.sign_in_user_id}/audit"
```
misaka: Any reason not to put the hostname into `Settings`?

app/helpers/dfe_signin_helper.rb
```diff
@@ -0,0 +1,9 @@
...
+  end
+
+  def dfe_signin_email_search(email)
+    "https://support.signin.education.gov.uk/users?criteria=#{email}"
```
misaka: Any reason not to put the hostname into `Settings`?

app/views/organisations/_organisation.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <% end %>
+  </td>
+  <td class="govuk-table__cell" data-qa="organisation-table-row__providers">
+    <% providers = organisation.providers %>
+    <% if providers.present? %>
+      <% if providers.size == 1 %>
+        <%= provider_details(providers.first) %>
+      <% else %>
+        <ul class="govuk-list govuk-list--bullet">
+          <% providers.map do |i| %>
```
misaka: `i`??? I know it's just a short loop, but surely we can do better!! ;)

spec/factories/organisation.rb
```diff
@@ -2,7 +2,25 @@
...
+    transient do
+      providers { [] }
+      users { [] }
+    end
+
     sequence(:id)
     name { "Organisation" }
+    nctl_ids { [] }
+
+    after :build do |organisation, evaluator|
```
misaka: Can we add the comment from the other factories about how this unexpected bit of code is necessary for the JSONAPI client?

app/controllers/organisations_controller.rb
```diff
@@ -0,0 +1,18 @@
+class OrganisationsController < ApplicationController
```
defong: don't this need to be locked down, to admin users only?

dankmitchell: Yeah, its managed by the backend i.e.

Screenshot 2019-12-12 at 13 07 28

defong: :+1:

spec/support/helpers.rb
dankmitchell: Done

app/helpers/dfe_signin_helper.rb
dankmitchell: Removed

app/views/organisations/_organisation.html.erb
dankmitchell: Agreed and sorted

app/helpers/dfe_signin_helper.rb
dankmitchell: Removed

spec/support/serializers/organisation_serializer.rb
dankmitchell: I think this is going to be harder than originally thought because of the through includes i.e. provider include `courses.accrediting_provider`.

defong:
[Move org#index](https://trello.com/c/TefMTNbR/2576-move-orgindex)

config/settings.yml
```diff
@@ -10,7 +10,7 @@ dfe_signin:
   # URL of this app for the callback after sigining in
   base_url: https://localhost:3000
   # The URL support users are directed to in order to find a user by email
-  email_search_url: https://support.signin.education.gov.uk/users?criteria=%<email>s
+  user_search_url: https://support.signin.education.gov.uk/users
```
defong: should this be per env?

spec/factories/user_factory.rb
```diff
@@ -10,6 +10,7 @@
     admin      { false }
     accept_terms_date_utc { Time.current }
     organisation_users { [] }
+    sign_in_user_id { "123456" }
```
defong: fake bot?

config/settings.yml
```diff
@@ -10,7 +10,7 @@ dfe_signin:
   # URL of this app for the callback after sigining in
   base_url: https://localhost:3000
   # The URL support users are directed to in order to find a user by email
-  email_search_url: https://support.signin.education.gov.uk/users?criteria=%<email>s
+  user_search_url: https://signin-test-sup-as.azurewebsites.net
```
defong: missing `/users`

config/settings.yml
dankmitchell: Done

PR #812, opened 2019-12-10
[2614] Outcome page reorder choices

danielburnley:
### Context

In order to bring the prototype in line with the application, we need to order the choices for outcomes. 

Originally this was done in the backend, but as it is display logic, it was better suited to the frontend with some checks to have assurance there are no extra options needed to be displayed

- Order options for outcomes
- Add checks for provided edit options

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/outcome_controller.rb
+++ b/app/controllers/courses/outcome_controller.rb
@@ -1,8 +1,17 @@
 module Courses
   class OutcomeController < ApplicationController
     include CourseBasicDetailConcern
+    before_action :order_edit_options, only: %i[edit new]
 
   private
+
+    def order_edit_options
+      @course.meta["edit_options"]["qualifications"] = if @course.level == "further_education"
+                                                         %w[pgce pgde]
+                                                       else
+                                                         %w[pgce_with_qts qts pgde_with_qts]
+                                                       end
+    end
 
     def current_step
       :outcome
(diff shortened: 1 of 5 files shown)


app/controllers/courses/outcome_controller.rb
```diff
@@ -1,13 +1,43 @@
...
     def current_step
       :outcome
     end
 
+    def ensure_edit_options_are_equal
+      qualification_options = @course.meta["edit_options"]["qualifications"]
+
+      if @course.level == "further_education"
+        if qualification_options.sort != non_qts_qualifications.sort
+          raise "Non QTS qualification options do not match"
```
misaka: I'm having a hard time parsing all this. Two suggestions come to mind:

1. I found the `sort` a little distracting here, even though now that I've read it, it's obviously matching the two arrays. I feel like this might be better, but maybe it's just because I wrote it:

```
        if (qualification_options.to_set ^ non_qts_qualifications.to_set).any?
```

2. what about moving the error raising into `qts_qualifications` and `non_qts_qualifications`? The `@course.level == "further_education"` check is being duplicated here, so that would simplify it.

danielburnley: I'm not 100% that number 1 is more readable, as it needs you to then think about sets and what that means. Though I could extract them to a helper method for just `are_qualifications_equal`, which is probably the better option :D 

Could potentially move the exception raising - i'll have a go and see how it looks :+1:

PR #831, opened 2019-12-17
Use the same environment colours as Apply

fofr:
- Staging is red
- QA is orange, not the focus colour

### DIFF
--- a/app/webpacker/stylesheets/_environments.scss
+++ b/app/webpacker/stylesheets/_environments.scss
@@ -2,17 +2,22 @@
   background: govuk-colour("dark-grey");
 }
 
-.app-tag--staging,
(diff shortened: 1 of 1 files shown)


app/webpacker/stylesheets/_environments.scss
```diff
@@ -2,17 +2,22 @@
   background: govuk-colour("dark-grey");
 }
 
-.app-tag--staging,
 .app-tag--qa {
-  background: govuk-colour("yellow");
-  color: $govuk-text-colour;
+  background: govuk-colour("orange");
```
misaka: Of course, `govuk` is going to spell `colour` correctly. :) Vive l'Alliance Anglaise!

PR #832, opened 2019-12-17
[2640] Support new validation added for course creation

danielburnley:
### Context

Validation has been added in the backend (see: https://github.com/DFE-Digital/manage-courses-backend/pull/1052 & https://github.com/DFE-Digital/manage-courses-backend/pull/1062) which requires the frontend to be linked to the keys for the error

- Add error keys to the relevant pages to display validation errors from the backend

**Note**

This PR does not cover age range with years with custom validation, as that is currently handled in the frontend of the controller and will be solved in future PRs

### Guidance to review

- Check out `2640-create-a-course-validation-changes` on MCBE (unless it has been merged)
- Run through course creation
- Try to create a course without submitting anything

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/applications_open_controller.rb
+++ b/app/controllers/courses/applications_open_controller.rb
@@ -1,7 +1,7 @@
 module Courses
   class ApplicationsOpenController < ApplicationController
     before_action :build_recruitment_cycle
-    before_action :build_course_params, only: :update
+    before_action :build_course_params, only: %i[update continue]
     include CourseBasicDetailConcern
 
     def update
@@ -9,7 +9,6 @@ module Courses
     end
 
     def continue
-      build_course_params
       super
     end
 
--- a/spec/features/courses/fee_or_salary/new_spec.rb
+++ b/spec/features/courses/fee_or_salary/new_spec.rb
@@ -57,6 +57,7 @@ feature "new course fee or salary", type: :feature do
     let(:course) do
       c = build(:course, provider: provider)
       c.errors.add(:funding_type, "Invalid")
+      c.errors.add(:program_type, "Invalid")
       c
     end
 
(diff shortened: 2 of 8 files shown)


spec/features/courses/sites/new_spec.rb
```diff
@@ -94,6 +94,19 @@
     end
   end
 
+  context "Error handling" do
+    let(:course) do
+      c = build(:course, provider: provider, site_ids: [])
+      c.errors.add(:sites, "Invalid")
+      c
```
faissalMT:
```
build(:course, provider: provider, site_ids: []).then do |course|
  course.errors.add(:sites, "Invalid")
end
```

misaka: If I'm not mistaken, `then` will return the result of `course.errors.add(:sites, "Invalid")`. `tap` would return the course.

```ruby
build(:course, provider: provider, site_ids: []).tap do |course|
  course.errors.add(:sites, "Invalid")
end
```

app/controllers/courses/applications_open_controller.rb
```diff
@@ -1,15 +1,14 @@
...
+    before_action :build_course_params, only: %i[update continue]
     include CourseBasicDetailConcern
 
     def update
       super
     end
 
     def continue
-      build_course_params
       super
```
misaka: Did we have a reason why we're leaving these method shells? Just as a signal to the reader that these should be implemented somewhere?

spec/features/courses/fee_or_salary/new_spec.rb
```diff
@@ -57,6 +57,7 @@
     let(:course) do
       c = build(:course, provider: provider)
       c.errors.add(:funding_type, "Invalid")
+      c.errors.add(:program_type, "Invalid")
```
misaka: If you accept the `.tap` suggestion in `sites/new_spec.rb` below, it's probably worth adding it here too.

app/controllers/courses/applications_open_controller.rb
faissalMT: Yes, rubocop now gives us an error if we set a callback for something that hasn't been implemented.

danielburnley:
Merging to avoid holding up future work that requires this :)

PR #834, opened 2019-12-18
[2701] highlight errors on fields in create course flow

faissalMT:
### Context
During DAC user testing we discovered that it was very important that users be able to see the error in the same place as the incorrect data especially if they were using high zooms or screen readers.  

Apply the appropriate formatting for fields with errors, according to https://design-system.service.gov.uk/components/error-summary/  
The red indicator as well as the small message as seen below was previously missing:
image
Also includes some refactoring here and there to make the create course pages more consistent and easier to work with.

### Guidance to review
You may need to checkout `2640-create-a-course-validation-changes` on the backend

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/apprenticeship/_form_fields.html.erb
+++ b/app/views/courses/apprenticeship/_form_fields.html.erb
@@ -1,7 +1,4 @@
-<div
-  class="<%= cns("govuk-form-group", "govuk-form-group--error": @errors&.[](:funding_type)&.any? || @errors&.[](:program_type)&.any?) %>"
-  data-qa="course__funding_type">
-
+<%= render "shared/error_wrapper", error_keys: [:funding_type, :program_type], data_qa: "course__funding_type" do %>
   <fieldset class="govuk-fieldset">
     <legend class="govuk-fieldset__legend govuk-fieldset__legend--xl">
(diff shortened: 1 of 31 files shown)


faissalMT:
[Highlight errors on fields in create course flow](https://trello.com/c/gpu5OKuV/2701-highlight-errors-on-fields-in-create-course-flow)

app/views/courses/apprenticeship/_form_fields.html.erb
```diff
@@ -1,5 +1,5 @@
 <div
-  class="<%= cns("govuk-form-group", "govuk-form-group--error": @errors && @errors[:funding_type]&.any?) %>"
+  class="<%= cns("govuk-form-group", "govuk-form-group--error": @errors&.[](:funding_type)&.any? || @errors&.[](:program_type)&.any?) %>"
```
danielburnley: This doesn't feel great to read and would potentially be quite confusing (A mixture of both multiple `&.`s and the `[](:key)` makes it quite hard to understand what's happening. Wonder if there's a helper method we could extract for this common pattern :thinking:

PR #846, opened 2019-12-23
[2712] Update DfE Signin URL's

dankmitchell:
### Context
DfE Sign-in URL's

Update "profile" and "search" URL's to use vanity URL.

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/config/settings/development.yml
+++ b/config/settings/development.yml
@@ -1,8 +1,6 @@
 dfe_signin:
-  issuer: https://signin-test-oidc-as.azurewebsites.net
-  profile: https://signin-test-pfl-as.azurewebsites.net
+  secret: please_change_me # Override with SETTINGS__DFE_SIGNIN__SECRET
   base_url: https://localhost:3000
-  user_search_url: https://signin-test-sup-as.azurewebsites.net/users
 manage_backend:
   secret: secret
   base_url: http://localhost:3001
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -4,13 +4,13 @@ dfe_signin:
   # URL that the users are redirected to for signing in
   issuer: https://signin-test-oidc-as.azurewebsites.net
   # URL of the users profile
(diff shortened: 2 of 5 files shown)


dankmitchell:
[[S] Change URLs for DfE Sign-In Test Env](https://trello.com/c/WwOHewpB/2712-s-change-urls-for-dfe-sign-in-test-env)

misaka:
It's weird that `development` and `pentest` provide both `profile` and `user_search_url` urls, but `qa and `test` only provide `profile`. What if we just removed all of them from `development`, `pentest`, `test` and `qa`, and relied on the setting to come from `settings.yml`?

config/settings/development.yml
```diff
@@ -1,8 +1,6 @@
 dfe_signin:
-  issuer: https://signin-test-oidc-as.azurewebsites.net
-  profile: https://signin-test-pfl-as.azurewebsites.net
+  secret: please_change_me # Override with SETTINGS__DFE_SIGNIN__SECRET
```
misaka: As a general pattern, it's best if we use random values for default values that are used for authentication, for security. Even if this isn't a deployed env, there's still risks involved I believe.

```suggestion
  secret: <%= SecureRandom.base64 %> # Override with SETTINGS__DFE_SIGNIN__SECRET
```

dankmitchell: We don’t do that in any other settings file. Consistently wrong is better than inconsistent

dankmitchell: But yes I agree

misaka: We are moving things over slowly. For example https://github.com/DFE-Digital/manage-courses-frontend/blob/2712-update-dsi-urls/config/settings.yml#L18

We can do this as a bigger move if you prefer.

PR #847, opened 2019-12-24
[2677] Add guidance and links for non-admin users requesting PE courses

davidgisbey:
### Context

PE courses should not be able to be created by providers as we need to ensure they have been allocated places.

- If the user is an admin allow PE creation as normal
- If not add guidance linking them to a google form

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/subjects/new.erb
+++ b/app/views/courses/subjects/new.erb
@@ -21,3 +21,12 @@
     <% end %>
   </div>
 </div>
+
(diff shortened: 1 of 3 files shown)


davidgisbey:
[[S] Remove PE from create a course journey for non-admin user](https://trello.com/c/DzJynp7k/2677-s-remove-pe-from-create-a-course-journey-for-non-admin-user)

PR #859, opened 2020-01-08
Update the 'New features' page with the current year

benilovj:


app/views/pages/new_features.html.erb
```diff
@@ -37,7 +37,7 @@
       <li>has an accredited body</li>
     </ul>
 
-    <h3 class="govuk-heading-l">Later in 2019</h3>
+    <h3 class="govuk-heading-l">Later in 2020</h3>
```
misaka: Why not just `Date.today.year`? It feels like cheating, but when would we want this to be last year?

benilovj: How many more years will this feature take??

misaka: Well ...

PR #861, opened 2020-01-09
[2537] Require login on/providers/suggest

gpeng:
## Context

Attempts to visit `/providers/suggest` whilst unauthenticated resulted
in a 500 rather than redirecting to signin.

Require authentication in the controller to remedy this and redirect the
user appropriately.

### Guidance to review

There is a big diff in the spec that has been caused by moving the majority of the existing specs into an 'authenticated' context to allow the new spec to be in an 'unauthenticated' one.  The actual new code is only 48 onwards.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/provider_suggestions_controller.rb
+++ b/app/controllers/provider_suggestions_controller.rb
@@ -1,6 +1,4 @@
 class ProviderSuggestionsController < ApplicationController
-  skip_before_action :authenticate
-
   def suggest
     suggestions = ProviderSuggestion.suggest(params[:query])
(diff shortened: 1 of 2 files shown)


gpeng:
[user needs to login to see https://www.publish-teacher-training-courses.service.gov.uk/providers/suggest](https://trello.com/c/2MW5Brvi/2537-user-needs-to-login-to-see-https-wwwpublish-teacher-training-coursesservicegovuk-providers-suggest)

PR #867, opened 2020-01-13
[2676] Rename Equivalence test to Equivalency test

dankmitchell:
### Context
Content issue

- Rename Equivalence test to Equivalency test

### Guidance to review
🚢 

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history


config/locales/en.yml
```diff
@@ -52,9 +52,9 @@ en:
         help: "You will consider applicants with a pending GCSE. But UCAS will block applications from anyone who needs to take an equivalency test."
         details_label: "Taking the GCSE"
       equivalence_test:
```
defong: are we keeping that as is?

PR #868, opened 2020-01-13
[2793] Fixed exploding nil study mode

defong:
### Context
Some how study mode can be nil

Make it more protected

### Guidance to review
:shipit: 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/_basic_details_tab.html.erb
+++ b/app/views/courses/_basic_details_tab.html.erb
@@ -109,7 +109,7 @@
           Full or part time
         </dt>
         <dd class="govuk-summary-list__value" data-qa="course__study_mode">
-          <%= course.study_mode.humanize %>
+          <%= course.study_mode&.humanize %>
         </dd>
         <dd class="govuk-summary-list__actions">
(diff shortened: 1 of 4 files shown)


defong:
[Protect against nil study_mode when displaying course](https://trello.com/c/tAi3WUKo/2793-protect-against-nil-studymode-when-displaying-course)

app/views/courses/_basic_details_tab.html.erb
```diff
@@ -109,7 +109,7 @@
           Full or part time
         </dt>
         <dd class="govuk-summary-list__value" data-qa="course__study_mode">
-          <%= course.study_mode.humanize %>
+          <%= course.study_mode&.humanize %>
```
dankmitchell: I thought we wanted to move humanize into the decorator?

defong: I tried to move it to the decorator as a bonus...
But a number of things was using the raw value, and I can't seems to find a way to access the undecorated value...

defong:
> It'd be worth doing a quick spec for this too.

done

PR #884, opened 2020-01-23
[2842] Prevent guidance page 500ing

faissalMT:
### Context
When https://www.publish-teacher-training-courses.service.gov.uk/guidance was visited by an admin user a 500 would occur because a request was being made to discover the number of access requests for the heading but this could not be done because the authentication procedure was not followed.  

Use a safer method for storing the user token & move the redirection of the user to the login page into an individually skippable function.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -5,6 +5,7 @@ class ApplicationController < ActionController::Base
 
   before_action :authenticate
   before_action :store_request_id
+  before_action :request_login
 
   def not_found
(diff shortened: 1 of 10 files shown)


faissalMT:
[Guidance page 500-ing for authenticated users](https://trello.com/c/gVbJdVQQ/2842-guidance-page-500-ing-for-authenticated-users)

PR #892, opened 2020-01-29
[2864] Return to the courses list when navigating back from the new level page

danielburnley:
### Context

Course creation had an issue where if you went back from the new level page, it took you to the confirmation page

- Reimplement the back link connection to the courses list

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/services/course_creation_step_service.rb
+++ b/app/services/course_creation_step_service.rb
@@ -30,6 +30,7 @@ class CourseCreationStepService
 
   def school_direct_workflow_steps
     %i[
+      courses_list
       level
       subjects
       modern_languages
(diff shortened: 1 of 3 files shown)


app/services/course_creation_step_service.rb
```diff
@@ -15,7 +15,12 @@ def get_next_step(steps, current_step)
 
   def get_previous_step(steps, current_step)
     previous_step_index = steps.find_index(current_step).pred
-    steps[previous_step_index]
+
+    if previous_step_index == -1
```
faissalMT: I think it's more readable if we add this as a discrete step to each workflow rather than making it the exception to the rule

danielburnley: :thinking: yeah that's true, i went back and forth. Sounds good :+1:

PR #894, opened 2020-01-29
2878 fix ci

rizzkhan1:
### Context
Build must terminate if failures are detected in `Run Tests` task 

Put if condition to validate test results

### Checklist
https://trello.com/c/72HRuTa5/2878-figure-out-why-master-is-failing

### DIFF
--- a/azure-pipelines.yml
+++ b/azure-pipelines.yml
@@ -37,7 +37,7 @@ steps:
 - script: |
     set -x
     docker run --name test-run-image -d $(IMAGE_NAME)
-    docker exec test-run-image rake "parallel:spec[,, -O .azure_parallel]"
+    # docker exec test-run-image rake "parallel:spec[,, -O .azure_parallel]"
     test_result=$?
       if [ "$test_result" != "0" ]
       then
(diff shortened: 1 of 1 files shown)


azure-pipelines.yml
```diff
@@ -38,6 +38,12 @@ steps:
     set -x
     docker run --name test-run-image -d $(IMAGE_NAME)
     docker exec test-run-image rake "parallel:spec[,, -O .azure_parallel]"
```
gpeng: Should this be running default `rake` to include linting and brakeman?

azure-pipelines.yml
```diff
@@ -38,6 +38,12 @@ steps:
...
+    test_result=$?
+      if [ "$test_result" != "0" ]
+      then
+          echo "##vso[task.logissue type=error]Cucumber test task exited abnormally."
```
gpeng: I don't think this has Cucumber tests, they're rspec.

azure-pipelines.yml
rizzkhan1: Not sure tbh. There are separate tasks for linting and brakeman. @TobyRet setup the parallel testing for us so might be able to comment

gpeng: If they're run elsewhere then that's fine. Just wasn't sure 👍

PR #903, opened 2020-02-05
[2859] Use new modern languages attribute for errors

faissalMT:
### Context
Users cannot create modern language courses because the subjects attribute triggers an error once the modern language subject is selected preventing the user from proceeding.
 
Watch the `modern_languages_subjects` attribute for errors instead of the subjects attribute (which is no longer used for modern language errors)

### Guidance to review
Requires: https://github.com/DFE-Digital/teacher-training-api/pull/1159

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/views/courses/modern_languages/edit.erb
+++ b/app/views/courses/modern_languages/edit.erb
@@ -10,20 +10,8 @@
     <%= form_with model: course,
           url: modern_languages_provider_recruitment_cycle_course_path(@course.provider_code, @course.recruitment_cycle_year, @course.course_code),
           method: :put do |form| %>
-      <div
-        class="<%= classnames("govuk-form-group", "govuk-form-group--error": @errors && @errors[:subjects]&.any?) %>"
-        data-qa="course__modern_languages_section">
-        <fieldset class="govuk-fieldset">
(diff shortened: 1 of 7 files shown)


faissalMT:
[[M] Modern languages course creation bug](https://trello.com/c/Yg6VBoHM/2859-m-modern-languages-course-creation-bug)

app/views/courses/modern_languages/edit.erb
```diff
@@ -10,20 +10,8 @@
...
-            <h1 class="govuk-fieldset__heading">
-              <span class="govuk-caption-xl"><%= course.name_and_code %></span>
-              Change modern languages
-            </h1>
-          </legend>
-          <%= render "shared/error_messages", error_keys: [:subjects] %>
-          <%= render 'courses/modern_languages/form_fields', form: form %>
-        </fieldset>
-      </div>
+      <%= render "shared/error_messages", error_keys: [:modern_languages_subjects] %>
```
gpeng: We should expose `error_keys` from the controller as they're already defined there. This is consistent with all of the other controller/views in the app though so let's not do it in this PR. I'll do a ticket.

PR #919, opened 2020-02-24
[2978] Add debugging for missing user info

faissalMT:
### Context
For some reason it seems that user info is missing, causing the backend to have issues sending the welcome email, because this requires the user's first name.

Log warnings when various user details are missing.

### Guidance to review
Also tries to make the user login workflow a tad easier to read.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- ~~[] Tested by running locally~~

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -29,8 +29,8 @@ class ApplicationController < ActionController::Base
   helper_method :current_user
 
   def current_user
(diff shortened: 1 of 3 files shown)


faissalMT:
[Investigate: Publish Access request bug](https://trello.com/c/j18waeLJ/2978-investigate-publish-access-request-bug)

PR #928, opened 2020-02-27
[3032] Add a healthcheck endpoint

aliuk2012:
This new `/healthcheck` endpoint checks whether the dependencies of this
service are up. By implication of having returned a 200 OK status you
can also tell that this service is up.

Failure of dependencies results in a 502 bad-gateway status and a json
body showing which dependencies failed.

Front-door will probably continue to check root `/` for a 200 status as
that doesn't have any dependencies currently, so no need for a ping
endpoint as well.

This code is based on
https://github.com/DFE-Digital/find-teacher-training/pull/159

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -5,8 +5,8 @@ Rails.application.routes.draw do
       via: %i[get post put]
   end
 
-  get :ping, controller: :health_checks
-  get :healthcheck, controller: :health_checks
+  get :ping, controller: :heartbeat
+  get :healthcheck, controller: :heartbeat
 
   # DfE Sign In
   get "/signin", to: "sessions#new", as: "signin"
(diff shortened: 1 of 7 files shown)


defong:
[Add /healthcheck endpoint to ptt](https://trello.com/c/wbZ05o6G/3032-add-healthcheck-endpoint-to-ptt)

config/routes.rb
```diff
@@ -5,7 +5,8 @@
       via: %i[get post put]
   end
 
-  get :ping, controller: :health_checks
+  get :ping, controller: :heartbeat
```
defong: > so no need for a ping
endpoint as well.

so needed?

aliuk2012: Yes ping is needed and used by Azure.

PR #930, opened 2020-02-27
[3051] Wizard for editing course modern languages

asmega:
### Context

- When updating a published course to include modern languages it was not possible to select languages as part of the process

- When changing subjects for a published course this now uses a wizard
- This allows for languages to be selected eg French, German
- Previously user could not select languages for the course
- An existing params mechanism is reused `course: { subjects_ids: [33] }`
- This query string is passed to wizard where it populates hidden fields
- This all this data is then sent to update endpoint
- Fix deprecation around `media_type`
- Removed un-needed pending test

### Guidance to review

- This PR depends on https://github.com/DFE-Digital/teacher-training-api/pull/1205 for the backend
- https://trello.com/c/8c9tBWQE/3051-m-make-editing-course-subjects-use-a-wizard
- For an existing course. Test both paths where course does and doesn't have modern languages
- Update to have modern languages with a selection of languages
- Course should now be updated correctly with new selection
- When editing a course with `Modern Languages` it should be pre-selected
- The previously chosen modern languages should also be pre-selected in the wizard 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -28,8 +28,9 @@ module Courses
     end
 
     def update
-      updated_subject_list = strip_non_language_subjects
-      updated_subject_list += selected_language_subjects
+      updated_subject_list = selected_language_subjects
+      updated_subject_list += selected_non_language_subjects
+
       if @course.update(subjects: updated_subject_list)
         flash[:success] = "Your changes have been saved"
         redirect_to(
@@ -63,15 +64,23 @@ module Courses
       [:modern_languages_subjects]
     end
 
-    def strip_non_language_subjects
-      @course.subjects.reject { |s| available_languages_ids.include?(s.id) }
-    end
-
     def selected_language_subjects
       language_ids = params.dig(:course, :language_ids)
       if language_ids.present?
         found_languages_ids = available_languages_ids & language_ids
         found_languages_ids.map { |id| Subject.new(id: id) }
+      else
+        []
+      end
+    end
+
+    def selected_non_language_subjects
+      ids = params.fetch(:course, {})[:subjects_ids]
+
+      if ids.present?
+        ids.map do |id|
+          Subject.new(id: id)
+        end
       else
         []
       end
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -75,14 +75,10 @@ module Courses
     end
 
     def selected_non_language_subjects
-      ids = params.fetch(:course, {})[:subjects_ids]
+      ids = params.dig(:course, :subjects_ids) || []
 
-      if ids.present?
-        ids.map do |id|
-          Subject.new(id: id)
-        end
-      else
-        []
+      ids.map do |id|
+        Subject.new(id: id)
       end
     end
 
--- a/app/controllers/courses/subjects_controller.rb
+++ b/app/controllers/courses/subjects_controller.rb
@@ -12,19 +12,19 @@ module Courses
     end
 
     def update
-      if subjects_have_not_been_changed?
-        flash[:success] = "Your subject hasn't been changed"
+      if has_modern_languages_subject?
         redirect_to(
-          details_provider_recruitment_cycle_course_path(
+          modern_languages_provider_recruitment_cycle_course_path(
             @course.provider_code,
             @course.recruitment_cycle_year,
             @course.course_code,
+            course: { subjects_ids: selected_subject_ids },
           ),
         )
       elsif @course.update(subjects: selected_subjects)
         flash[:success] = "Your changes have been saved"
         redirect_to(
-          modern_languages_provider_recruitment_cycle_course_path(
+          details_provider_recruitment_cycle_course_path(
             @course.provider_code,
             @course.recruitment_cycle_year,
             @course.course_code,
@@ -37,6 +37,19 @@ module Courses
     end
 
   private
+
+    def has_modern_languages_subject?
+      selected_subjects.any? do |s|
+        s.id.to_s == modern_languages_subject.id.to_s
+      end
+    end
+
+    def modern_languages_subject
+      return @modern_languages_subject if @modern_languages_subject
+
+      hash = @course.meta[:edit_options][:modern_languages_subject]
+      @modern_languages_subject = Subject.new(hash)
+    end
 
     def subjects_have_not_been_changed?
       subjects_match?(selected_subjects, existing_non_language_subjects)
@@ -56,14 +69,16 @@ module Courses
       end
     end
 
-    def selected_subjects
-      selected_subject_ids = params
+    def selected_subject_ids
+      params
         .dig(:course)
         .slice(:master_subject_id, :subordinate_subject_id)
         .to_unsafe_h
         .values
         .select(&:present?)
+    end
 
+    def selected_subjects
       selected_subject_ids.map do |subject_id|
         subject_hash = find_subject(subject_id)
         Subject.new(subject_hash.to_h)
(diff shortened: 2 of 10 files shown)


asmega:
I have fixed "Modern Languages" being preselected in the dropdown on the first page. For me the original languages on the second page of the wizard were already being pre-populated

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -77,6 +74,18 @@ def selected_language_subjects
       end
     end
 
+    def selected_non_language_subjects
+      ids = params.fetch(:course, {})[:subjects_ids]
```
defong:
```suggestion
      ids = params.dig(:course, :subjects_ids) || [] 
```

maybe

app/controllers/courses/subjects_controller.rb
```diff
@@ -38,6 +38,19 @@ def update
...
+    end
+
+    def modern_languages_subject
+      return @modern_languages_subject if @modern_languages_subject
+
+      hash = @course.meta[:edit_options][:modern_languages_subject]
+      @modern_languages_subject = Subject.new(hash)
+    end
+
     def subjects_have_not_been_changed?
```
defong: Is `subjects_have_not_been_changed?` still in use? and `existing_non_language_subjects`

asmega: doesn't look like it. i'll try removing

app/controllers/courses/modern_languages_controller.rb
asmega: 6 and 2 threes?

asmega: actually i prefer yours as this will always return a collection

app/controllers/courses/modern_languages_controller.rb
```diff
@@ -77,6 +74,18 @@ def selected_language_subjects
...
+
+      if ids.present?
+        ids.map do |id|
+          Subject.new(id: id)
+        end
+      else
+        []
+      end
```
danielburnley: If you implement saus change (https://github.com/DFE-Digital/publish-teacher-training/pull/930/files#r387766944) you can remove ths `if` and just do the map

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -39,19 +44,20 @@
...
       [
         modern_languages_subject,
       ]
     end
 
     scenario "can select multiple modern language subjects" do
-      languages_page.load_with_course(course)
+      visit signin_path
+
+      visit(modern_languages_provider_recruitment_cycle_course_path(provider_code: provider.provider_code, recruitment_cycle_year: current_recruitment_cycle.year, code: course.course_code, course: { subjects_ids: [modern_languages_subject.id] }))
```
danielburnley: this is a very long line to read :D would be worth adding some newlines

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -150,6 +139,51 @@
...
+    end
+
+    context "but is changed to a modern language course" do
+      before do
+        subjects_page.master_subject_fields.select("Modern Languages")
+        subjects_page.save.click
+      end
+
+      it "takes user through wizard and saves subjects" do
+        expect(CGI.unescape(current_url)).to eql("http://www.example.com/organisations/#{provider.provider_code}/2020/courses/#{course.course_code}/modern-languages?course[subjects_ids][]=#{modern_languages_subject.id}")
```
danielburnley: Could you just check the current path and the current query? I haven't seen other examples of where we've matched the whole URL :thinking:. Though IMO this should be split into two separate assertions

Reason for this is it feels like this is making two assertions in one here:

1. We get taken to the modern-languages page
2. The `subjects_ids` are in the query

app/controllers/courses/modern_languages_controller.rb
asmega: I've changed this now and removed the nil guard

app/controllers/courses/subjects_controller.rb
asmega: i've removed this and a few other methods that didn't seem to be in use

spec/features/courses/subjects/edit_spec.rb
```diff
@@ -52,7 +54,7 @@
 
       subjects_page.master_subject_fields.select(subjects.first.subject_name)
       subjects_page.save.click
-      expect(languages_page).to be_displayed
+      expect(current_url).to eql("http://www.example.com/organisations/#{provider.provider_code}/2020/courses/#{course.course_code}/details")
```
danielburnley: To be consistent with the other tests within publish we should expect the `details_page` page object to be displayed (not sure what the class is for it off the top of my head)

faissalMT: The reason we aren't doing this is because site prism is notoriously inconsistent with handling URL parameters so like we're doing on Find, we should just entirely avoid it.

app/decorators/course_decorator.rb
```diff
@@ -214,6 +214,13 @@ def selectable_subjects
     meta["edit_options"]["subjects"].map { |subject| [subject["attributes"]["subject_name"], subject["id"]] }
   end
 
+  def selected_subject_ids
+    selectable_subject_ids = meta["edit_options"]["subjects"].map { |subject| subject["id"] }
+    selected_subject_ids = subjects.map(&:id)
+
+    selectable_subject_ids & selected_subject_ids
+  end
+
```
danielburnley: https://github.com/DFE-Digital/publish-teacher-training/blob/master/spec/decorators/course_decorator_spec.rb

Could do with adding tests for this :eyes:

spec/models/subject_spec.rb
```diff
@@ -1,3 +1,2 @@
 describe Subject do
-  pending
```
defong: looks pointless can be deleted?

spec/features/courses/modern_languages/edit_spec.rb
```diff
@@ -150,6 +148,53 @@
...
+
+    context "but is changed to a modern language course" do
+      before do
+        subjects_page.master_subject_fields.select("Modern Languages")
+        subjects_page.save.click
+      end
+
+      it "takes user through wizard and saves subjects" do
+        modern_languages_uri = URI(current_url)
+        expect(modern_languages_uri.path).to eq("/organisations/#{provider.provider_code}/2020/courses/#{course.course_code}/modern-languages")
```
defong:
```suggestion
        expect(modern_languages_uri.path).to eq("/organisations/#{provider.provider_code}/#{current_recruitment_cycle.year}/courses/#{course.course_code}/modern-languages")
```

spec/features/courses/subjects/edit_spec.rb
```diff
@@ -52,7 +61,9 @@
 
       subjects_page.master_subject_fields.select(subjects.first.subject_name)
       subjects_page.save.click
-      expect(languages_page).to be_displayed
+
+      subject_uri = URI(current_url)
+      expect(subject_uri.path).to eq("/organisations/#{provider.provider_code}/2020/courses/#{course.course_code}/details")
```
defong:
```suggestion
      expect(subject_uri.path).to eq("/organisations/#{provider.provider_code}/#{current_recruitment_cycle.year}/courses/#{course.course_code}/details")
```

spec/features/courses/subjects/edit_spec.rb
```diff
@@ -80,7 +91,7 @@
         subjects_page.subordinate_subject_accordion.click
         subjects_page.subordinate_subject_fields.select(subjects.second.subject_name)
         subjects_page.save.click
-        expect(languages_page).to be_displayed
+        expect(current_url).to eql("http://www.example.com/organisations/#{provider.provider_code}/2020/courses/#{course.course_code}/details")
```
defong:
```suggestion
        expect(current_url).to eql("http://www.example.com/organisations/#{provider.provider_code}/#{current_recruitment_cycle.year}/courses/#{course.course_code}/details")
```

PR #936, opened 2020-03-02
[3067] Remove selenium driver gem

aliuk2012:
As we using webdriver gem there is no need to include selenium driver gem.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history


aliuk2012:
[Replace chrome/selenium driver gems with webdrivers gem](https://trello.com/c/5iK9RNPH/3067-replace-chrome-selenium-driver-gems-with-webdrivers-gem)

PR #940, opened 2020-03-04
[3082] Visible difference for admin-only features

timabell:
### Context

To avoid confusion when developing/reviewing features.

> Seeing the half finished accredited provider listing on prod gave me a minor heart attack, which I then shared with @misha and @Fred. Turns out you can only see at as an admin. 

Visible difference for admin-only features

Selection_811

### Guidance to review

:eyes:

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review (just from screenshots on this one I'm afraid)

### Still todo

- [x] move html into partial
- [x] move css into css file
- [x] test the helper https://stackoverflow.com/questions/440571/how-do-i-test-helpers-in-rails
- [x] see if the pages are testing visibility for an actual feature with mocked admin/non-admin


timabell:
[highlight admin features](https://trello.com/c/QmEeA8I3/171-highlight-admin-features)

aliuk2012:
I've got a local copy of this branch should we need to revive this pr.

timabell:
Someone said leave till Monday. Hello Monday! :wave:

aliuk2012:
👀

aliuk2012:
Heres my initials thoughts:

1. Padlock looks weird when it's floating in the middle of the page:
image


2. in navigation, they looked too squashed:
image
 or in mobile:
image


I'm not sure if @jesseyuen  will review this as he did say that as much as he would like to be involved with this he just doesn't have bandwidth to look at it. He did suggest possibly reusing what Apply pattern have :
image

I do like apply patterns but I'm not sure how they address admin only navigation links as we have in Publish.

timabell:
I'm kind of hoping that now that I've got the files & code in place that prove the concept I can hand over the design to the frontend folks and you can add more commits to this PR to make it look acceptable.

In my mind the design doesn't matter that much because it's (currently) for our eyes only, so as long as it conveys that it's something different and the hover-text explains what is different I'd say it's "good enough". The key thing for me is that there's a clue in the UI that there's a difference in behaviour for the current user vs an end user; and I'd say it already achieves that. With that in mind, personally I'd merge this as is and those who want it to look better can iterate on it when they have time.

(Though I still need to add regression tests for this PR before merging it.)

aliuk2012:
How about we use Apply pattern for admin sections on a page and then position the lock icon to the left of text in navigation. I think we could do this using a combination of css `::before`  selector and `content` attribute.

jesseyuen:
Screenshot 2020-03-09 at 16 20 25

> I think we could do this using a combination of css ::before selector and content attribute.

I was going to suggest this ☝️. I don't visually love the padlock emoji but don't want it to block us.

And for the tag why not go for one of the new colours like `govuk-tag--purple` maybe?

aliuk2012:
I've removed the helper methods and gone with a simple html/css approach.

image

misaka:
❤️

timabell:
Added regression tests. Updated PR screenshot. Rebased on master.
:ship: ?

aliuk2012:
Could you make squash or remove some commits? I was thinking maybe the first three commits could be dropped

timabell:
@aliuk2012 we could squash them, yes; but personally I wouldn't for this one.

I generally think that once a PR is open I want to maintain the commits that go with the discussion that happened unless it becomes a complete mess.
If you are reviewing history later on and don't want to see all the commits on branches you can use [git log --first-parent](https://explainshell.com/explain?cmd=git+log+--first-parent)

I don't really want to spend more time on this. It ate up too much of yesterday as I learned all the foibles of doing view testing.

timabell:
Thanks for the improved visuals, it's much nicer now.

dankmitchell:
Are we shipping this? @timabell @aliuk2012 ?

I would also squash but I know Tim's feelings and won't argue.

fofr:
Rather than the emojis in the heading, can we make a single link that says "Admin" (or similar), and from that page link to all the admin features, with descriptions for what each features does?

timabell:
@fofr we could do that as a follow up.
For ideas like that we are using individual columns in the find team ideas board. Based on the basecamp no-backlog. I'm sure you can have a column too :-)

fofr:
Ok, but emojis in a main nav 😱

timabell:
@fofr it's only visible to us

fofr:
Have suggested to @aliuk2012 that we move the links to the footer and put them in an "Admin" section.

timabell:
@fofr is going to pony up some new commits :rocket:

fofr:
Commit added, links moved to footer:

Screen Shot 2020-03-12 at 11 49 09

aliuk2012:
Thanks @fofr 👍 

Hooray!!! no emoji being used 😄🍻

PR #948, opened 2020-03-09
[3100] Fix modern languages step always showing

faissalMT:
### Context
The modern language subjects are now always given to publish, and this was historically used to distinguish between courses that needed the modern languages selection page and those that did not.

Use the presence of the modern language subject to decide whether or not to show the modern languages page.

### Guidance to review
Similar issues may exist during modification of subjects in existing courses, this should be investigated and put into another card if applicable.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history

### DIFF
--- a/app/controllers/courses/modern_languages_controller.rb
+++ b/app/controllers/courses/modern_languages_controller.rb
@@ -6,7 +6,7 @@ module Courses
     include CourseBasicDetailConcern
 
     def new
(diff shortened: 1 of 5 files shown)


faissalMT:
[INVESTIGATION: Art and Design Course Creation Bug](https://trello.com/c/AwCmwiDF/3100-art-and-design-course-creation-bug)

PR #971, opened 2020-03-23
[3149] Add course count on training providers page

AgaDufrat:
### Context

As an accrediting body
I want to see all the courses I accredit
So that I know what providers are running courses on our behalf

* The course count text appears on the "Courses as an Accredited Body" page for each training provider
  - only courses accredited by that specific accredited body are included

<kbd>Screenshot 2020-03-23 at 09 18 38</kbd>

### Guidance to review
visit https://localhost:3000/organisations/B20/2020/training-providers

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -61,17 +61,13 @@ class ProvidersController < ApplicationController
   end
 
   def training_providers
-    if user_is_admin?
-      @training_providers = @provider.training_providers(recruitment_cycle_year: @recruitment_cycle.year)
+    @training_providers = @provider.training_providers(recruitment_cycle_year: @recruitment_cycle.year)
 
-      courses = Course.where(
-        recruitment_cycle_year: @recruitment_cycle.year,
-        accrediting_provider_code: @provider.provider_code,
-      )
-      @course_counts = courses.group_by(&:provider_code).transform_values(&:size)
-    else
-      redirect_to provider_path(@provider.provider_code)
-    end
+    courses = Course.where(
+      recruitment_cycle_year: @recruitment_cycle.year,
+      accrediting_provider_code: @provider.provider_code,
+    )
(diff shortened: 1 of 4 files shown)


dankmitchell:
[Add course count on Courses as accredited body page](https://trello.com/c/Q6gNSNor/3149-add-course-count-on-courses-as-accredited-body-page)

app/views/providers/training_providers.html.erb
```diff
@@ -26,6 +26,9 @@
                         ),
                         class: "govuk-link govuk-!-font-weight-bold",
                         data: { qa: "link" } %>
+            <span class="govuk-body govuk-!-font-weight-regular govuk-!-display-block" data-qa="course_count">
```
defong: `0 courses` ?

app/controllers/providers_controller.rb
```diff
@@ -64,6 +64,21 @@ def publish
   def training_providers
```
TobyRet: Not for this PR but I wonder if this is an indication that a separate controller / restful resource is required ... Looks like we are trying to make a distinction between accrediting bodies and accredited bodies

AgaDufrat: Misha and I had a chat whether this should be in a separate controller and decided to keep it here.

app/controllers/providers_controller.rb
```diff
@@ -63,6 +63,12 @@ def publish
...
     if user_is_admin?
```
dankmitchell, line 64: Do we need to rebase with master to remove this?

AgaDufrat: I did but looks like this conditional wasn't removed as a part of https://github.com/DFE-Digital/publish-teacher-training/pull/998 I will add another commit and remove it.

PR #980, opened 2020-03-26
[3192] Remove region select so users cannot edit

aliuk2012:
### Context
Removing this form field as it is overridden by geocoding lookup.
It is no longer needed.

#### Before
image

#### After

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review


aliuk2012:
[Remove "Region of UK" from location editing](https://trello.com/c/qRG9VJ5g/3192-remove-region-of-uk-from-location-editing)

PR #985, opened 2020-03-30
Readme - use asdf-vm to install dependencies

timabell:
This eliminates the need to update the versions in the readme.
Using asdf makes setup a lot less steps.

### DIFF
--- a/README.md
+++ b/README.md
@@ -2,7 +2,9 @@
 
 # Publish Teacher Training
 
-## Prerequisites
+## Development setup
+
+### 1. Install build dependencies
 
 Install [asdf-vm](https://asdf-vm.com/).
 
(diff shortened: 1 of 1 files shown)


asmega:
Are developers not allowed to choose their own tooling? If you don't want to update every time maybe just remove the version numbers?

timabell:
@dankmitchell @asmega  I think it would be good to promote asdf as a good consistent one-stop option for people getting started, but I think it's good to let people use whatever tools suit them. I don't see anything in this PR that precludes personal choice of tooling, it's just an easy to use default.

timabell:
Hmm, let me re-phrase that a bit.

When someone is new to the team or project there's a lot of information overload, so if they show up with a fresh mac-book I think having only `asdf` in the readme of all of our repos would make it very easy for someone new to checkout/build/run.

Having that as the only thing in the readme is not to say that you can't use other tools if you are more confident and have other preferences. Is there anything in this PR or our repo that makes you worry you can't use other tools?

I'd be interested to know what tools other people are using or want to use. Do you have any other thoughts to add on choice of tools for these projects?

TobyRet:
@timabell  I think having some guidance in the README on `asdf` is useful for devs setting up the project for the first time. Perhaps it may be better to change the section title to something other than "Prerequisites" to address the concerns above? e.g "Recommended setup", "Setup guidance" or something along those lines ...

timabell:
Good idea @TobyRet

timabell:
Went a bit further but hopefully it's a big improvement. Any thoughts @asmega @TobyRet @dankmitchell or shall I just merge and move on?

PR #1000, opened 2020-04-06
3150 csv download courses as accredited body

defong:
### Context
csv downloadable

As an accredited body, they should be able to download a csv that other provider has used them as the accredited body.

### Guidance to review
csv content consist of 
- provider_code,
- provider_name,
- course_code,
- course_name,
- study_mode,
- qualification,
- content_status,
- applications_open_from,
- has_vacancies

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -99,12 +99,15 @@ class ProvidersController < ApplicationController
   end
 
   def training_providers_courses
-    @courses = @provider.current_accredited_courses.map do |c|
+    @courses = Course.includes(:provider)
+      .where(recruitment_cycle_year: @recruitment_cycle.year, accrediting_provider_code: @provider.provider_code)
+      .map do |c|
       {
           provider_code: c.provider.provider_code,
           course_code: c.course_code,
       }
     end
+
     respond_to do |format|
       format.csv
     end
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -3,8 +3,7 @@ class ProvidersController < ApplicationController
   decorates_assigned :training_provider
   before_action :build_recruitment_cycle
   rescue_from JsonApiClient::Errors::NotFound, with: :not_found
-  before_action :build_provider, except: %i[index show training_providers_courses]
-  before_action :build_accredited_provider, only: %i[training_providers_courses]
+  before_action :build_provider, except: %i[index show]
   before_action :build_training_provider, only: %i[training_provider_courses]
 
   def index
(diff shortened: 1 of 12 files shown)


app/controllers/providers_controller.rb
```diff
@@ -95,6 +96,21 @@ def search
     redirect_to provider_path(provider_code)
   end
 
+  def training_providers_courses
```
gpeng, line 99: I think it would be simpler, and more in keeping with the conventions, for this to an index action on another controller e.g. `TrainingProviderCoursesController`

app/controllers/providers_controller.rb
```diff
@@ -95,6 +96,21 @@ def search
...
+    @courses = Course.includes(:provider)
```
gpeng, line 100: We should change the name of this as they're not really 'courses' that the variable holds. `course_csv_rows` maybe? Also, you could probably just build the csv in the controller and return it as a string without the need for the template.

app/controllers/providers_controller.rb
```diff
@@ -95,6 +96,21 @@ def search
...
+      .where(recruitment_cycle_year: @recruitment_cycle.year, accrediting_provider_code: @provider.provider_code)
+      .map do |c|
+      {
+          provider_code: c.provider.provider_code,
+          course_code: c.course_code,
+      }
+    end
+
+    respond_to do |format|
+      format.csv
```
dankmitchell, line 110: format.csv { send_data @courses.to_csv, filename: "courses-#{Date.today}.csv" }?

app/views/providers/training_providers_courses.csv.erb
```diff
@@ -0,0 +1,9 @@
+<%
```
dankmitchell, line 1: I think this could live in the course model via `to_csv` to match by above suggestion.

defong:
[[M] CSV download courses as accredited body](https://trello.com/c/DNwQbfAL/3150-m-csv-download-courses-as-accredited-body)

app/views/providers/training_providers_courses.csv.erb
defong: I dont' understand why this file is here tbh....

app/controllers/providers_controller.rb
defong: done

app/controllers/providers_controller.rb
defong: done

spec/requests/providers_spec.rb
```diff
@@ -66,6 +66,36 @@
...
+        path = download_training_providers_courses_provider_recruitment_cycle_path(
+          accredited_provider.provider_code,
+          accredited_provider.recruitment_cycle.year,
+          format: :csv,
+        )
+        get(path)
+
+        expect(response).to have_http_status(:ok)
+        expect(response.body).to eq(
+          <<~HEREDOC,
```
asmega, line 92: is the comma on this line a mistake?

defong: no

```

 Style/TrailingCommaInArguments: Put a comma after the last parameter of a multiline method call. (https://rubystyle.guide#no-trailing-params-comma)
          <<~HEREDOC
```

PR #1012, opened 2020-04-14
[3276] Add encoding for autocomplete queries

SamJamCul:
### Context

https://trello.com/c/4nSbRPzG/3276-s-publish-provider-autocomplete-not-encoding-request-correctly

Changed the way the autocomplete generates the query string - now uses the encodeURIComponent function to help protect against unicode contamination. See this PR for a similar change on Find https://github.com/DFE-Digital/find-teacher-training/pull/204

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/provider_suggestions_controller.rb
+++ b/app/controllers/provider_suggestions_controller.rb
@@ -1,7 +1,16 @@
 class ProviderSuggestionsController < ApplicationController
   def suggest
-    suggestions = ProviderSuggestion.suggest(params[:query])
+    return render(json: { error: "Bad request" }, status: :bad_request) if params_invalid?
+
+    sanitised_query = CGI.escape(params[:query])
+    suggestions = ProviderSuggestion.suggest(sanitised_query)
       .map { |provider| { code: provider.provider_code, name: provider.provider_name } }
     render json: suggestions
+  end
+
+private
+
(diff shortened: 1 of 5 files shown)


SamJamCul:
[[S] Publish provider autocomplete not encoding request correctly](https://trello.com/c/4nSbRPzG/3276-s-publish-provider-autocomplete-not-encoding-request-correctly)

SamJamCul:
> LGTM - Could you also add the spec to Find @SamJamCul ?

It might be the case that this isn't going to fix the core issue - this will handle unicode stuff, but it won't be able to search for it at all...

app/controllers/provider_suggestions_controller.rb
```diff
@@ -1,6 +1,7 @@
 class ProviderSuggestionsController < ApplicationController
   def suggest
-    suggestions = ProviderSuggestion.suggest(params[:query])
+    sanitised_query = CGI.escape(params[:query])
```
aliuk2012, line 3: We are missing tests for this.

Could I suggest we follow the pattern we used in API to create a service that escapes the query param? We can then test the service and then for this route we mock the service.

PR #1014, opened 2020-04-14
[3246] Improve providers pagination

asmega:
### Context

- Originally publisher was requesting whole providers dataset from API
- This change makes it request paginated data so a subset of the dataset
- Load times are therefore quicker and paginated through providers is quicker 

- Swap kaminari for pagy, this was a nice to have. I did swap this out in the end as there were strange failures in tests which I could not resolve. Pagy does seem simpler and lightweight which is desired and we do not need most of the functionality that exists in the older Kaminari gem
- Providers pagination requests paginated data from API

### Guidance to review

- There should be no visual changes. The only thing that should change is the load times when clicking though the pages of the provider list as an admin
- visit review app https://s121d02-3246-ptt-as.azurewebsites.net/
- log in as normal user
- visit https://s121d02-3246-ptt-as.azurewebsites.net/ should redirect you to your provider page
- close/open new incognito browser
- visit https://s121d02-3246-ptt-as.azurewebsites.net/
- login as admin
- should see list of providers
- should be able to paginate through different providers and click on them
- paginating should be faster compared to QA or production

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review


asmega:
[[S] Paginate providers in publish and api](https://trello.com/c/xM3j6Ohd/3246-s-paginate-providers-in-publish-and-api)

app/views/pagy/_next_page.html.erb
```diff
@@ -1,12 +1,12 @@
 <li class="pub-c-pagination__item pub-c-pagination__item--next">
-  <%= link_to url, rel: 'next', remote: remote, class: "pub-c-pagination__link govuk-link" do %>
+  <%= link_to "#{request.path}?page=#{pagy.next}", rel: 'next', class: "pub-c-pagination__link govuk-link" do %>
```
asmega, line 2: this is rudimentary but works for our use case. I tried a couple of other things but would need a fair bit of work to deal with all the cases

app/views/pagy/_previous_page.html.erb
```diff
@@ -1,5 +1,5 @@
 <li class="pub-c-pagination__item pub-c-pagination__item--previous">
-  <%= link_to url, rel: 'prev', remote: remote, class: "pub-c-pagination__link govuk-link" do %>
+  <%= link_to "#{request.path}?page=#{pagy.prev}", rel: 'prev', class: "pub-c-pagination__link govuk-link" do %>
```
asmega, line 2: this is rudimentary but works for our use case. I tried a couple of other things but would need a fair bit of work to deal with all the cases

app/views/providers/index.html.erb
```diff
@@ -10,4 +10,4 @@
   <%= render partial: "providers/provider", collection: @providers %>
 </ul>
 
-<%= paginate @providers %>
+<%= pagy_govuk_nav(@pagy).html_safe %>
```
asmega, line 13: linter does not support `<%==`

spec/features/courses/accredited_body/new_spec.rb
```diff
@@ -23,6 +23,11 @@
     stub_api_v2_build_course
     stub_api_v2_build_course(level: "primary")
 
+    stub_api_v2_request(
+      "/recruitment_cycles/2020/providers?page[page]=1",
+      resource_list_to_jsonapi([provider], meta: { count: 1 }),
+    )
+
```
asmega, line 30: lots of test visit `signin_path` which then goes to load the providers index page

dankmitchell:
@asmega can we remove the Kaminri views are they used else where?

app/controllers/providers_controller.rb
```diff
@@ -7,8 +7,13 @@ class ProvidersController < ApplicationController
   before_action :build_training_provider, only: %i[training_provider_courses]
 
   def index
-    @providers = Kaminari.paginate_array(Provider.where(recruitment_cycle_year: Settings.current_cycle).all)
-                          .page(params[:page] || 1).per(params[:per_page] || 10)
+    page = (params[:page] || 1).to_i
+    per_page = 10
```
dankmitchell, line 11: Does this need to be hardcoded? What about a `per_page` param?

app/helpers/application_helper.rb
```diff
@@ -1,4 +1,12 @@
+require "pagy/extras/shared"
```
dankmitchell, line 1: Any reason this is here and not in the initialiser?

spec/support/helpers.rb
```diff
@@ -24,6 +24,10 @@ def stub_omniauth(user: nil, provider: nil)
     # TODO: Move this to be returned with the user.
     stub_api_v2_request(
       "/recruitment_cycles/#{Settings.current_cycle}/providers",
+      resource_list_to_jsonapi([provider], meta: { count: 1 }),
+    )
+    stub_api_v2_request(
+      "/recruitment_cycles/#{Settings.current_cycle}/providers?page[page]=1&page[per_page]=1",
```
asmega, line 30: something in our tests is making an API call to request for a single provider as this endpoint

app/controllers/providers_controller.rb
asmega, line 11: hard coded on both sides for now

asmega: sure i can change

app/helpers/application_helper.rb
asmega: i will move, this was a leftover of me trying to get pagy to work

asmega: turns out it's not needed any more so i've removed

app/controllers/providers_controller.rb
asmega: ok second thought, i will leave as is. it's hard coded on the API side so would require another change there. It's only used here so the hard code is fine i think. I don't really see this changing so will leave for now

PR #1039, opened 2020-04-23
[3526] -  Status labels on "Request PE courses for 2020/21" page

TobyRet:
### Context
Currently the status labels on the PE Allocation page are hard coded. 

- Fetch the Accrediting Body's allocations from the API
- Use Decorator pattern to return allocation statuses to the view

### Guidance to review
- Request a PE allocation for an accrediting body (visit `organisations/[PROVIDER_CODE]/2020/allocations`) - status should change from 'YET TO REQUEST' to 'REQUESTED'. Declining an allocation should change status from 'YET TO REQUEST' to 'NOT REQUESTED'. Declined and requested allocations can be edited, with their statuses updated accordingly.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review


spec/view_objects/providers/allocations_view_spec.rb
```diff
@@ -0,0 +1,61 @@
+require "rails_helper"
+
+describe AllocationsView do
```
TobyRet, line 3: I'm assuming that because the allocation statuses are unit tested we do not need to also test in the allocation feature spec

TobyRet:
[[L] 2d. Status labels on "Request PE courses for 2020/21" page](https://trello.com/c/Os9zFalm/3256-l-2d-status-labels-on-request-pe-courses-for-2020-21-page)

app/view_objects/allocations_view.rb
```diff
@@ -0,0 +1,57 @@
...
+      allocation_status[:status_colour] = Colour::GREEN
+    end
+
+    if matching_allocation && matching_allocation[:number_of_places].zero?
+      allocation_status[:status] = Status::NOT_REQUESTED
+      allocation_status[:status_colour] = Colour::RED
+    end
+
+    allocation_status
+  end
```
aliuk2012, line 56: As part of this could we build the actions for each status could have? I'm thinking it would be good to refactor https://github.com/DFE-Digital/publish-teacher-training/blob/master/app/views/providers/allocations/_request_status.html.erb#L28-L53 and move logic out of the view and into this view_object.

TobyRet: Hey @aliuk2012 - I've had have a think / play around with your suggestion and I'm not sure there is much scope for further refactoring (without introducing additional complexity). The key issue is that https://github.com/DFE-Digital/publish-teacher-training/blob/master/app/views/providers/allocations/_request_status.html.erb#L28-L53 now varies only in markup. The complex business logic has largely been removed and put into the Presenter (AllocationsView). It could be possible to make the AllocationsView generate markup but not sure if I am comfortable with this. 
A further option may be to extract the actions to an _actions_ partial if that would help?

app/view_objects/allocations_view.rb
```diff
@@ -0,0 +1,64 @@
...
+    end
+  end
+
+  def build_allocation_status(matching_allocation, training_provider)
+    allocation_status = {
+      training_provider_name: training_provider.provider_name,
+      training_provider_code: training_provider.provider_code,
+    }
+
+    if matching_allocation.nil?
```
gpeng, line 45: We could move each of these predicates into a method e.g. `yet_to_request?`, `requested?` and `not_requested?` then it would be a bit clearer what their significance is and it would isolate the changes we're going to need to make when the `request_type` gets added to the response. We could do that when we make the changes though so not a biggy.

spec/factories/providers.rb
```diff
@@ -6,7 +6,7 @@
       include_counts { [] }
     end
 
-    sequence(:id)
+    sequence(:id, &:to_s)
```
gpeng, line 9: nice.. didn't know you could do that.

spec/features/providers/allocations_spec.rb
```diff
@@ -24,7 +25,8 @@
 
   scenario "Accredited body views PE allocations page when training provider has no PE fee-founded course" do
     given_accredited_body_exists
-    given_there_is_no_training_providers_with_pe_fee_founded_course
+    given_there_is_no_training_providers_with_pe_fee_funded_course
```
gpeng, line 28: While we're fixing typo-ses this should be `are_no` 😆

PR #1056, opened 2020-04-30
[3337] don't require SSL in development mode

misaka:
### Context

Having to deal with SSL certificates is a pain.

Skip SSL in local dev mode.

### Guidance to review

🚀 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/README.md
+++ b/README.md
@@ -32,32 +32,27 @@ bundle
 bundle exec rake webpacker:compile
 ```
 
-### 3. Configure DfE Sign-In
+### 3. Setup Basic Auth Login
 
-Create the following file and ask the team for the secret.
+For local development, we rely on Basic Auth instead of DFE Sign-In, which can
+be enabled if required. Create the file `config/settings/development.local.yml`
+with the contents:
 
-```
-# config/settings/development.local.yml
-dfe_signin:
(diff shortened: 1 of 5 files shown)


README.md
```diff
@@ -32,32 +32,27 @@ bundle
...
-dfe_signin:
-    secret: dfe_sign_in_test_server_client_secret_here
-```
-
-*If you don't have a sign-in account on test you can bypass sign-in by
-configuring basic auth instead.*
 
+The email address has to exist in the users table of teacher-training-api, which
+means that you'll need to have an account on the production system (or add
+yourself by hand), but the password can be any non-secure local password you
```
AgaDufrat, line 52: Do you think it would be useful to add an example of adding a user by hand since there is a couple of things to include.

```
User.create(admin: true, email: "john.smith@digital.education.gov.uk", 
first_name: "John", last_name: "Smith", welcome_email_date_utc: Time.now,
accept_terms_date_utc: Time.now, invite_date_utc: Time.now, state: "transitioned")
```

misaka: Sure can. 👍

misaka: @AgaDufrat Done.

PR #1061, opened 2020-05-01
[3340] Add start page

gpeng:
**UPDATE: we've decided not to merge this for now as the risk of deploying outweighs the benefits with DfE Signin currently responding. I'll leave open for now and we can merge if needed**

### Context

Signin is currently down and we need to display a message to users.

This implementation shows a page only in non-test enviroments. This was
necessary as a lot of tests break due to the flow change so this has to
be manually tested.

* Render a start page when the user is not-authenticated that allows them to click to signin and more importantly at this point allows us to display a message.
* Maintain the old flow for tests 
* Update the signin link in the header to go directly to 'signin'

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,6 +2,11 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
+    if Settings.features.start_page
+      render "pages/start"
+      return
+    end
+
     redirect_to "/auth/dfe"
   end
 
--- a/app/controllers/pages_controller.rb
+++ b/app/controllers/pages_controller.rb
@@ -1,5 +1,5 @@
 class PagesController < ApplicationController
-  skip_before_action :request_login, only: %i[guidance accessibility terms cookies privacy new_features]
+  skip_before_action :request_login, only: %i[guidance accessibility terms cookies privacy new_features start]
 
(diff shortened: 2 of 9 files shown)


defong:
image

app/controllers/sessions_controller.rb
```diff
@@ -2,7 +2,12 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
-    redirect_to "/auth/dfe"
+    if Rails.env.test?
```
misaka, line 5: Suggestion:

```suggestion
    if Settings.features.signin_intercept
```

And add this to the settings. I'm also adding `signin_magic_link`, or something like that, to `Settings.features` too. Happy to approve as-is, though.

defong:
for ref

https://github.com/DFE-Digital/publish-teacher-training/blob/e16d87c4bab27670f3e2ebe0d8276970ec57b1f9/azure/template.json

dankmitchell:
[Add start page](https://trello.com/c/nA3RgZZe/3340-add-start-page)

gpeng:
I'm going to close this and re-open as the purpose and context is totally different now.

PR #1063, opened 2020-05-04
[3343] add magic link login

misaka:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,7 +2,7 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
-    if Settings.features.signin_intercept
+    if FeatureService.enabled? :signin_intercept
       render
     else
       redirect_to "/auth/dfe"
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -33,6 +33,44 @@ class SessionsController < ApplicationController
     redirect_to_correct_page(user)
   end
 
+  def create_by_magic
+    FeatureService.require(:signin_by_email)
+
+    user = Session.create_by_magic(
+      magic_link_token: params.require(:token),
+      email: params.require(:email),
+    )
+
+    if user
+      session[:auth_user] = HashWithIndifferentAccess.new(
+        "uid" => nil,
+        "info" => HashWithIndifferentAccess.new(
+          email: user.email,
+          first_name: user.first_name,
+          last_name: user.last_name,
+        ),
+        "credentials" => HashWithIndifferentAccess.new(
+          "id_token" => nil,
+        ),
+      )
+      set_session_info_for_user(user)
+      Raven.user_context(id: current_user["user_id"])
+      logger.debug { "User session create_by_magic " + log_safe_current_user.to_s }
+    end
+
+    redirect_to_correct_page(user)
+  end
+
+  def send_magic_link
+    FeatureService.require(:signin_by_email)
+
+    User.generate_and_send_magic_link(params.require(:user).require(:email))
+
+    redirect_to magic_link_sent_path
+  end
+
+  def magic_link_sent; end
+
   def signout
     if current_user.present?
       if development_mode_auth?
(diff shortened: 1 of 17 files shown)


app/services/feature_service.rb
```diff
@@ -0,0 +1,15 @@
+module FeatureService
+  class << self
+    def require(feature_name)
+      unless Settings.features[feature_name]
```
gpeng, line 4: We could use `enabled?` here

dankmitchell:
[Add support for magic-links login in Publish](https://trello.com/c/0npFT6AY/3343-add-support-for-magic-links-login-in-publish)

app/models/user.rb
```diff
@@ -9,4 +9,27 @@ class User < Base
   def self.member(id)
     new(id: id)
   end
+
+  def self.generate_and_send_magic_link(email)
```
gpeng, line 13: This doesn't seem like a `User` concern. Could this be a service?

config/settings/development.yml
```diff
@@ -11,3 +11,5 @@ environment:
   selector_name: "development"
 log_level: debug
 use_ssl: false
+features:
```
gpeng, line 14: Do we want to change the default by default in dev?

spec/services/feature_service_spec.rb
```diff
@@ -0,0 +1,56 @@
+require "rails_helper"
+
+describe FeatureService do
+  let(:feature_enabled) { true }
+
+  around(:each) do |example|
+    old_value = Settings.features.rspec_testing
```
gpeng, line 7: Can we just stub this with `allow(Settings.features).to receive(:rspec_testing)` to save setting and unsetting.

misaka: In `FeatureService` we access the setting like so:

```ruby
      Settings.features[feature_name]
```

so I was imagining we need to be a bit more sophisticated than just stubbing it out.

app/controllers/sessions_controller.rb
```diff
@@ -34,6 +34,46 @@ def create
...
+        "credentials" => HashWithIndifferentAccess.new(
+          "id_token" => nil,
+        ),
+      )
+      set_session_info_for_user(user)
+      Raven.user_context(id: current_user["user_id"])
+      logger.debug { "User session create_by_magic " + log_safe_current_user.to_s }
+
+      redirect_to_correct_page(user)
+    else
```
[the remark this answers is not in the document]
misaka: Ya, that sounds like a good idea.

app/models/user.rb
misaka: Ya, I was being lazy / expedient. :)

spec/services/feature_service_spec.rb
gpeng:
```
  before do
    allow(Settings.features).to receive(:[]).with(:rspec_testing).and_return(feature_enabled)
  end
```

spec/services/magic_link/generate_and_send_service_spec.rb
```diff
@@ -0,0 +1,94 @@
+require "rails_helper"
+
+describe MagicLink::GenerateAndSendService do
+  let(:email) { "bat@localhost" }
+  let(:site)  { "http://localhost:3000" }
+
+  # subject { described_class(email: email, site: site) }
```
gpeng, line 7:
```suggestion
```

spec/support/features.rb
```diff
@@ -0,0 +1,21 @@
+RSpec.configure do |configure|
```
gpeng, line 1: Might be able to stub here rather than altering the global state. This way is probably fine too but I wonder if there are errors that could occur in `example.run` that would stop the reset happening.

misaka: I was worried about that too, it turns out that the code after `example.run` still executes even if an expectation fails, or worse, an exception is raised. Which is why I thought an `ensure` block wasn't necessary here.

Also makes sense since this is likely the same mechanism used by mocking libraries. This behaviour is even specced out:

https://relishapp.com/rspec/rspec-core/v/3-7/docs/hooks/around-hooks#an-around-hook-continues-to-run-even-if-the-example-throws-an-exception

spec/services/feature_service_spec.rb
misaka: My reasoning was that I'd rather it wasn't tied to whether we are doing `Setting.features.rspec_testing` or `Setting.features[:rspec_testing]`.

If that's not sufficiently convincing I can change it to use a stub.

gpeng: Ok. Fair enough 👍

spec/support/features.rb
gpeng: 👍

PR #1068, opened 2020-05-05
3331 send request type from publish when creating updating an allocation

TobyRet:
### Context
We have implemented `request_type` on the Allocation API endpoint.
This PR updates publish to pass the `request_type` param to the API

#### Creating an allocation

Initial (new) allocation requests send:
`request_type: "initial"`
`number_of_places: <int>`

Repeat allocations requests send:
`request_type: "repeat"`

Declining an repeat allocation send:
`request_type: "decline"`

#### Updating an allocation

Initial requests update `number_of_places`

Repeat and declined requests have the request type changed: `repeat|declined`

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/services/allocation_services/create.rb
+++ b/app/services/allocation_services/create.rb
@@ -1,0 +1,36 @@
+module AllocationServices
+  class Create
+    include ServicePattern
+
+    def initialize(requested:, accredited_body_code:, provider_id:, number_of_places: nil)
+      @requested = requested
+      @accredited_body_code = accredited_body_code
+      @provider_id = provider_id
+      @number_of_places = number_of_places
+    end
+
+    def call
+      Allocation.create(create_params)
+    end
+
+  private
+
+    attr_reader :requested, :accredited_body_code, :provider_id, :number_of_places
+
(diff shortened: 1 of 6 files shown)


TobyRet:
[Send request_type from Publish when creating/updating an Allocation](https://trello.com/c/NRo5HovS/3331-send-requesttype-from-publish-when-creating-updating-an-allocation)

app/services/allocation_services/create.rb
```diff
@@ -0,0 +1,36 @@
+module AllocationServices
+  class Create
+    include PublishService
```
misaka, line 3: Sorry to nitpick about naming, but I thought this service had something to do with publishing a course, whereas having read the module below I see it's probably about exposing `call` as a class method. Would it make sense to use a name that's more orthogonal to this service?

gpeng: Fair point. I named it meaning 'a service in the Publish app' but I see the confusion.

PR #1069, opened 2020-05-05
[3282] - Display new (initial) allocation requests

TobyRet:
### Context
- We want to present users with a table displaying initial allocation requests (i.e. where the allocation has `request_type` == `initial`)
- If initial allocation requests do not exist for the accrediting body then this table is not displayed

- Displaying of repeat and initial allocation requests is now determined by the `request_type` attribute which is now returned from the API

### Guidance to review
- As the user facing functionality for creating an `initial` request is still to be created in order to see the new table in all its glory you will need to manually create or modify an allocation record so that it's `request_type` == `initial`

### Screenshot
initial_table

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


TobyRet:
[9. New request table on "Request PE courses for 2021/22" page](https://trello.com/c/wttve8sh/3283-9-new-request-table-on-request-pe-courses-for-2021-22-page)

app/views/providers/allocations/request_status/_repeat.html.erb
```diff
@@ -0,0 +1,60 @@
...
+              </a>
+            </li>
+          </ul>
+        <% end %>
+      </td>
+    </tr>
+  <% end %>
+  </tbody>
+</table>
+<% end %>
```
aliuk2012, line 60: Do we need two versions of what will basically be the same table markup?

TobyRet: Yep true

TobyRet: I've refactored to use a single partial

app/views/providers/allocations/_request_status.html.erb
```diff
@@ -1,57 +1,60 @@
...
-  </thead>
-  <tbody class="govuk-table__body">
-    <% @allocation_statuses.each do |allocation| %>
-        <tr class="govuk-table__row">
-          <th class="govuk-table__header" data-qa="provider-name">
-            <%= allocation[:training_provider_name] %>
-          </th>
-          <td class="govuk-table__cell">
+<% if allocations.present? %>
+  <table class="govuk-table ucas-info-table govuk-!-margin-bottom-8" data-qa="<%= request_type %>-allocations-table">
```
aliuk2012, line 2: smart use of request_type

PR #1075, opened 2020-05-06
[3340] start page

gpeng:
### Context

When Signin is unavailable we need to be able to display an alternative login interface/system. 

This was previously open as https://github.com/DFE-Digital/publish-teacher-training/pull/1061 which was approved.

Add a start page that can be enabled with a feature flag. This start page will ultimately provide a magic link interface for users to login. (not included in this PR).

### Guidance to review

This is relatively straightforward but it turned out to be complex to add a test that when the feature isn't enabled we still redirect to DfE Signin. Turns out we don't have an explicit test for this (as far as I could see in any case). 

DfE Signin doesn't actually return to us when authentication fails but mocking failure out seemed to be the only way to test that it was included and we were redirecting when the user was unauthenticated and tried to access `root_path` We do implicitly test this in other tests as they fail if we don't stub out OAuth so there is coverage of a sort.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,7 +2,11 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
-    redirect_to "/auth/dfe"
+    if Settings.features.signin_intercept
+      render
+    else
+      redirect_to "/auth/dfe"
+    end
   end
 
   def create
--- a/azure/template.json
+++ b/azure/template.json
@@ -164,6 +164,13 @@
       "metadata": {
           "description": "List of resource tags as a JSON object"
       }
+    },
+    "settingsFeaturesSigninIntercept": {
+      "type": "string",
+      "defaultValue": "false",
+      "metadata": {
+        "description": "when true, intercept the signin flow with an explanatory page; used when there are issues with DfE Sign-In"
+      }
     }
   },
   "variables": {
--- a/app/views/sessions/new.html.erb
+++ b/app/views/sessions/new.html.erb
@@ -1,0 +1,26 @@
+<%= content_for :page_title, "Sign in" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+    <div class="app-banner app-banner--warning govuk-!-margin-bottom-7">
+      <div class="app-banner__message">
+        <h2 class="govuk-heading-m">You might have problems signing in</h2>
+        <p>Some of our users are having trouble signing in. We’re working on the problem now. If you can’t sign in, try again soon.</p>
+      </div>
+    </div>
+
+    <h1 class="govuk-heading-xl">
+      Sign in
+    </h1>
+
+    <p class="govuk-body-l">You must sign in to your account to manage your teacher training courses.</p>
+
+    <%= link_to "Sign in using DfE Sign-in", "/auth/dfe", class: "govuk-button" %>
+
+    <hr class="govuk-section-break govuk-section-break--l govuk-section-break--visible">
+
+    <h2 class="govuk-heading-m">Ask for an account</h2>
+    <p class="govuk-body">Don’t have an account? You can get one by emailing <%= bat_contact_mail_to %>. Please include the full names and email addresses of anyone who needs access.</p>
+  </div>
+</div>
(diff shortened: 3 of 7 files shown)


gpeng:
[Add start page](https://trello.com/c/nA3RgZZe/3340-add-start-page)

app/controllers/sessions_controller.rb
```diff
@@ -2,6 +2,11 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
+    if Settings.features.start_page
+      render "pages/start"
+      return
```
defong, line 7:
```suggestion
      render "pages/start" and return
```

https://guides.rubyonrails.org/layouts_and_rendering.html#avoiding-double-render-errors

dangling `return`

app/controllers/sessions_controller.rb
```diff
@@ -2,6 +2,11 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
+    if Settings.features.start_page
+      render "pages/start"
```
defong, line 6: shouldn't this be a redirect?

https://github.com/DFE-Digital/publish-teacher-training/blob/d38cf8b79527183ec2f716a79935ffc9d605e9eb/config/routes.rb#L11

spec/features/start_page_spec.rb
```diff
@@ -0,0 +1,117 @@
...
+      "/recruitment_cycles/#{current_recruitment_cycle.year}",
+      current_recruitment_cycle.to_jsonapi,
+    )
+
+    visit root_path
+  end
+
+  def then_i_am_redirected_to_the_start_page
+    expect(page.current_path).to eq(signin_path)
+  end
```
defong, line 84: feels misleading?

`def xxxxx_start_page`
`eq(signin_path)`

app/controllers/sessions_controller.rb
gpeng: Rubocop wouldn't allow `and return` and we need to `return` so 🤷

app/controllers/sessions_controller.rb
gpeng: It could be but this seemed the simplest way when we were in a hurry the other day. Redirecting would mean another action and/or controller and I'm not sure there's much benefit to be had for that overhead?

spec/features/start_page_spec.rb
gpeng: Yeah.. maybe that would be some benefit of having a separate action.

gpeng:
Discussed this offline. There may be a cleaner approach in configuring OmniAuth to do this rather than hijacking the flow in the sessions controller. I'm going to try that before we commit to this approach.

app/controllers/sessions_controller.rb
defong: ain't picking a fight with Rubocop

app/controllers/sessions_controller.rb
defong: There is already __the__ action `start` and the controller `page`

https://github.com/DFE-Digital/publish-teacher-training/blob/25164abdbd66bd233f2b36dcb76e0005283df332/config/routes.rb#L203

:confused:

misaka: I think that this isn't a start page, this is a signin intercept page. For example, if somebody goes to `/signin` they see this screen.

azure/template.json
```diff
@@ -164,6 +164,13 @@
       "metadata": {
           "description": "List of resource tags as a JSON object"
       }
+    },
+    "settingsFeaturesStartPage": {
```
defong, line 168:
```suggestion
    "settingsFeaturesSigninIntercept": {
```

azure/template.json
```diff
@@ -164,6 +164,13 @@
...
+      "type": "string",
+      "defaultValue": "false",
+      "metadata": {
+        "description": "true to automatically redirect unauthenticated users to DfE signin. false to display start page"
```
defong, line 172: there is on such thing as a start page

misaka: Good catch, thought I'd got all these. 👍

misaka:
I'll squash once approved.

PR #1078, opened 2020-05-06
[3364] User getting 403 errors when viewing their training providers on the courses as an accredited body page

gpeng:
### Context

Users are getting 403 responses when trying to view courses as an accredited body.

There were some remnants of this being an admin only feature left in the code. Remove them so as non-admin users don't get redirected or 403-ded

Change the way that we request courses from the API to avoid going through the provider#show endpoint and hitting the authorization failure.

NB: This requires https://github.com/DFE-Digital/teacher-training-api/pull/1354 to be merged

### Guidance to review

This needs to be reviewed as a non-admin user on the review app. I'm not sure whether we have a non-admin user that belongs to an accredited body so manual checking requires a bit of finagling in the console. I can pair on that bit if it helps.

We don't check authorization in the frontend as we rely on the API to enforce that so we should take a bit of extra care reviewing that too.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/features/providers/courses_as_an_accredited_body_spec.rb
+++ b/spec/features/providers/courses_as_an_accredited_body_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-feature "Get courses as an accredited body", type: :feature do
+feature "get courses as an accredited body", type: :feature do
   let(:organisation_training_providers_page) { PageObjects::Page::Organisations::TrainingProviders.new }
   let(:courses_as_an_accredited_body_page) { PageObjects::Page::Organisations::OrganisationCoursesAsAnAccreditedBody.new }
 
@@ -12,7 +12,7 @@ feature "Get courses as an accredited body", type: :feature do
 
   let(:training_provider2) { build :provider, accredited_bodies: [accrediting_body1, accrediting_body2], courses: [course1, course2] }
 
-  let(:user) { build :user, :admin }
+  let(:user) { build :user }
   let(:access_request) { build :access_request }
 
   before do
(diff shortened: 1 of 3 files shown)


gpeng:
[User getting 403 errors when viewing their training providers on the "courses as an accredited body" page](https://trello.com/c/OdnDymOC/3364-user-getting-403-errors-when-viewing-their-training-providers-on-the-courses-as-an-accredited-body-page)

spec/features/providers/courses_as_an_accredited_body_spec.rb
```diff
@@ -35,10 +35,15 @@
       "#{accrediting_body1.provider_code}/training_providers?recruitment_cycle_year=#{accrediting_body1.recruitment_cycle.year}",
       resource_list_to_jsonapi([training_provider2, accrediting_body2]),
     )
+    stub_api_v2_request(
+      "/recruitment_cycles/#{accrediting_body1.recruitment_cycle.year}/providers/#{training_provider2.provider_code}" \
+      "/courses?filter[accrediting_provider_code]=#{accrediting_body1.provider_code}",
+      resource_list_to_jsonapi([course1]),
+    )
```
misaka, line 42: Are we avoiding using `stub_api_v2_resource`?

gpeng: I don't think it supports `filter` qs params does it? We could add that I suppose...

misaka: I think it'd be nice to have, assuming people are still using these stubs, but not crucial for this change.

PR #1081, opened 2020-05-07
[3250] Update the error copy for search field

aliuk2012:
If the user is attempting to search for a provider which we do not have
on the system, we display an error message with some advice to contact
support to get the provider added.

### Guidance to review

- Login as an accredited body user
- Go to request new allocation (https://s121d02-3250-ptt-as.azurewebsites.net/organisations/B20/2020/allocations/request) 
- Search for a provider who would not be on the system
	- Displays error message with hint text to contact support
- Try a search for the provider but do not type in the provider name
	- Error message saying you need to type in more information

image

image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/form_objects/initial_request_form.rb
+++ b/app/form_objects/initial_request_form.rb
@@ -4,11 +4,9 @@ class InitialRequestForm
   attr_accessor :training_provider_code, :training_provider_query
 
   def add_no_results_error
-    support_email = "becomingateacher@digital.education.gov.uk"
-
     errors.add(:training_provider_query,
                "We couldn't find this organisation - please check your information and try again.
-                To add a new organisation to Publish, contact #{support_email}.")
+                To add a new organisation to Publish, contact #{Settings.service_support.contact_email_address}.")
   end
 
   def add_no_search_query_error
--- a/app/form_objects/initial_request_form.rb
+++ b/app/form_objects/initial_request_form.rb
@@ -4,7 +4,11 @@ class InitialRequestForm
   attr_accessor :training_provider_code, :training_provider_query
 
   def add_no_results_error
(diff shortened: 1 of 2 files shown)


aliuk2012:
[6c. "Who are you requesting a course for?" page - error message -organisation not on Publish](https://trello.com/c/6sCKYpG7/3250-6c-who-are-you-requesting-a-course-for-page-error-message-organisation-not-on-publish)

app/form_objects/initial_request_form.rb
```diff
@@ -4,7 +4,11 @@ class InitialRequestForm
   attr_accessor :training_provider_code, :training_provider_query
 
   def add_no_results_error
-    errors.add(:training_provider_query, "We couldn't find this provider, please check your input and try again")
+    support_email = "becomingateacher@digital.education.gov.uk"
+
+    errors.add(:training_provider_query,
+               "We couldn't find this organisation - please check your information and try again.
+                To add a new organisation to Publish, contact #{support_email}.")
```
AgaDufrat, line 11: Maybe we could do 
```suggestion
                To add a new organisation to Publish, contact #{Settings.service_support.contact_email_address}.")
```
to keep the email in one place

defong: does it need mailto?

aliuk2012: @AgaDufrat, awesome suggestion 💯 

@defong I was thinking about making it a mailto but there are two problems with that.

- The form builder doesn't support html in error messages.
- I don't think we should include markup in an error message because in the error summary the error message itself is a link so by adding a mailto link we would end up with a link inside a link

PR #1087, opened 2020-05-12
[3231] Add allocation update functionality

SamJamCul:
Co-authored-by: Toby Retallick <toby.retallick@gmail.com>

### Context

https://trello.com/c/mC7AqJvX/3231-s-11-change-my-allocations-request

* Add update method to allocations
* Hook up "change" link on allocations page

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


SamJamCul:
[[S] 11. Change my allocations request](https://trello.com/c/mC7AqJvX/3231-s-11-change-my-allocations-request)

TobyRet:
[[S] 11. Change my allocations request](https://trello.com/c/mC7AqJvX/3231-s-11-change-my-allocations-request)

app/controllers/providers/allocations_controller.rb
```diff
@@ -5,6 +5,7 @@ class AllocationsController < ApplicationController
     before_action :build_training_provider, except: %i[index initial_request]
     before_action :require_provider_to_be_accredited_body!
     before_action :require_admin_permissions!
+    before_action :build_allocation, only: %i[update edit_repeat_request]
```
TobyRet, line 8: We may want to consider refactoring these before action callbacks
See https://ukgovernmentdfe.slack.com/archives/CGVNJ2Q2F/p1589301415297900?thread_ts=1589291822.294000&cid=CGVNJ2Q2F

app/controllers/providers/allocations_controller.rb
```diff
@@ -27,6 +28,8 @@ def index
 
     def repeat_request; end
```
TobyRet, line 29: Given we have defined `edit_repeat_request` I wonder if we should rename this action (and template) `new_edit_request` to appease the Rails-ey consistency

app/views/providers/allocations/edit_repeat_request.html.erb
```diff
@@ -0,0 +1,65 @@
...
+            <span class="govuk-caption-xl"><%= @training_provider.provider_name %></span>
+            <%= page_title %>
+        </h1>
+    </legend>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+
+  <%= form_with url: provider_recruitment_cycle_allocation_path(
```
TobyRet, line 21: GOVUK form builder candidate?

app/controllers/providers/allocations_controller.rb
gpeng: I agree with @asmega in the Slack chat about getting rid of this. This one is also incorrectly named as we're not building it, we're retrieving it from the API. That's an important distinction here because it means we have the saved state of the `Allocation` so we don't really need a separate edit action (edit_repeat_request). We can just edit and render a different view (or view in a different state) depending on the state of the retrieved allocation.

We can probably have a wider discussion about the slack thread but this defo makes more sense as:

```
def allocation
  @_allocation ||= Allocation.find(<whatevs>)
end
```

app/controllers/providers/allocations_controller.rb
gpeng: Agreed `new_` would be more explicit here. I'm not sure we need the `edit_repeat_request` version though. We know the state of the `Allocation` (as per comment above) so `edit` should be fine and avoid adding too many 'non-standard' actions.

app/controllers/providers/allocations_controller.rb
```diff
@@ -44,6 +47,14 @@ def show
       @allocation = Allocation.new(number_of_places: number_of_places)
     end
 
+    def update
+      request_type = requested? ? "repeat" : "declined"
```
gpeng, line 51: We should use the constants here `Allocation::RequestTypes::REPEAT` etc

app/controllers/providers/allocations_controller.rb
```diff
@@ -72,6 +83,10 @@ def build_recruitment_cycle
       @recruitment_cycle = RecruitmentCycle.find(cycle_year).first
     end
 
+    def build_allocation
```
gpeng, line 86:
```suggestion
    def allocation
      @_allocation ||= Allocation.find(params["id"]).first
    end
```

Also... Does `.find` return a collection? 😞

app/view_objects/allocations_view.rb
```diff
@@ -79,12 +79,14 @@ def build_repeat_allocations(matching_allocation, training_provider)
       allocation_status[:status] = Status::REQUESTED
       allocation_status[:status_colour] = Colour::GREEN
       allocation_status[:requested] = Requested::YES
+      allocation_status[:id] = matching_allocation.id
```
gpeng, line 82: This method (and the one below) should have singular method names. I wonder if an `AllocationView` (i mean a singular object that is a view representation of one `Allocation`) would be a better more testable way of implementing this too? May be a bit of a stretch for this PR I realise.

app/views/providers/allocations/edit_repeat_request.html.erb
gpeng: @aliuk2012 ?

spec/features/providers/allocations_spec.rb
```diff
@@ -355,6 +395,12 @@ def when_i_click_confirm_choice
     click_on "Confirm choice"
   end
 
+  def when_i_click_change
+    stub_api_v2_resource(@training_provider)
```
gpeng, line 399: I think these should go in `given_` etc rather than this `when_I_click`

SamJamCul:
Not going to use the form builder as we're doing some quirky logic with the radio buttons

app/views/providers/allocations/edit_repeat_request.html.erb
aliuk2012: Sorry only just seen this. We should use the form builder

aliuk2012:
@SamJamCul what quirky logic are we doing? Is there anything I can help you with?

app/controllers/providers/allocations_controller.rb
```diff
@@ -25,13 +22,20 @@ def index
       )
     end
 
-    def repeat_request; end
+    def new_repeat_request
+      training_provider
+    end
+
+    def edit_repeat_request
```
gpeng, line 29: I still think we could do away with the `_repeat_request` and just have an `edit`. We know what the request type is from the retrieved `Allocation` so we can decide what to render based on that either by rendering a different template or having the state of the template depend on the request type of the allocation we're editing.

SamJamCul: Alrighty!

PR #1090, opened 2020-05-13
[3364] Add sorting to training provider courses

gpeng:
### Context

We updated the 'Courses as an accredited body' functionality to avoid 403 errors that some users were experiencing (these were due to using the providers endpoint to retrieve the courses)

Prior to the update introduced the results were sorted by course code.

Add sorting to the course results. The API doesn't currently support this and there is no pagination so we can just sort in the frontend.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -79,7 +79,10 @@ class ProvidersController < ApplicationController
     @courses = Course
       .where(recruitment_cycle_year: @recruitment_cycle.year)
       .where(provider_code: @training_provider.provider_code)
-      .where(accrediting_provider_code: @provider.provider_code).map(&:decorate)
+      .where(accrediting_provider_code: @provider.provider_code)
+      .map(&:decorate)
+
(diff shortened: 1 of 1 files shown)


gpeng:
[User getting 403 errors when viewing their training providers on the "courses as an accredited body" page](https://trello.com/c/OdnDymOC/3364-user-getting-403-errors-when-viewing-their-training-providers-on-the-courses-as-an-accredited-body-page)

defong:
ummm cannot see the mirror in #1078 or tests

fofr:
Why are they sorted by course code and not course title?

gpeng:
That's a good question. I was almost going to change it but for this I just wanted to match what was there before as I changed the implementation and the ordering had changed. I'll merge and deploy this so it's as designed and then follow up Monica & Jesse and see was this intentional. The feature is broken at the moment so I'm keen to get this out.

PR #1094, opened 2020-05-14
[3305] - Create org users page

TobyRet:
**Note** that API changes need to be merged / deployed first - https://github.com/DFE-Digital/teacher-training-api/pull/1360

### Context
As a provider
I need to see the other users that have access to my org
So that I can check the right people have access

- Creation of new page that lists users that have access to a particular provider (organisation)

users_page

### Guidance to review
- This is dependent on API changes so you will need to test locally
- Run the API on branch `3305-return-user-association-from-provider`
- Visit a providers fancy new users page `/organisations/K30/users`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/views/providers/users/index.html.erb
+++ b/app/views/providers/users/index.html.erb
@@ -1,0 +1,20 @@
+<%= content_for :page_title, "Users" %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <h1 class="govuk-heading-xl">Users</h1>
+    <p class="govuk-body">
+      These users currently have access to this account. They can manage all your course and organisation details, including publishing, closing and withdrawing courses.
+    </p>
+    <%= link_to "Invite user", request_access_provider_path(code: @provider.provider_code), class: "govuk-button" %>
+    <% @users.each do |user| %>
+      <div>
+        <h2 class="govuk-heading-m govuk-!-margin-bottom-1">
+          <%= user.first_name user.last_name %>
+        </h2>
+        <p class="govuk-body govuk-!-margin-bottom-0"><%= user.email %></p>
+      </div>
+      <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
+    <% end %>
+  </div>
+</div>
--- a/spec/features/providers/user_spec.rb
+++ b/spec/features/providers/user_spec.rb
@@ -33,7 +33,7 @@ feature "Provider users page" do
   def when_i_visit_the_providers_user_page
     stub_api_v2_request(
       "/recruitment_cycles/#{current_recruitment_cycle.year}/providers/#{@provider.provider_code}?include=users",
-      @provider.to_jsonapi(include: %i[users])
+      @provider.to_jsonapi(include: %i[users]),
     )
 
     visit "/organisations/#{@provider.provider_code}/users"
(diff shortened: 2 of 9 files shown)


TobyRet:
[[S] Create org users page (iteration 1)](https://trello.com/c/MSz5Q3Ts/3305-s-create-org-users-page-iteration-1)

app/views/providers/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      These users currently have access to this account. They can manage all your course and organisation details, including publishing, closing and withdrawing courses.
+    </p>
+    <%= link_to "Invite user", request_access_provider_path(code: @provider.provider_code), class: "govuk-button" %>
+    <% @users.each do |user| %>
+      <div>
+        <h2 class="govuk-heading-m govuk-!-margin-bottom-1">
+          <%= user.first_name user.last_name %>
+        </h2>
+        <p class="govuk-body govuk-!-margin-bottom-0"><%= user.email %></p>
+      </div>
```
aliuk2012, line 16: I'm not sure each users name has to be a heading.
also it might be better to make it a list of users and style them to look like different "sections"

app/views/providers/users/index.html.erb
```diff
@@ -0,0 +1,20 @@
...
+      <hr class="govuk-section-break govuk-section-break--visible govuk-section-break--m">
```
aliuk2012, line 17: this might not be needed if the li is styled correctly.

app/views/providers/users/index.html.erb
dankmitchell: It's consistent with APPLY - https://manage-applications-beta.herokuapp.com/users

aliuk2012: I'm not suggesting we should change the visual style just the markup. I'm seeing this as a list of users. I don't think theres enough content exposed about the user from one or more sources to warrant it to be considered as a "Section" on the page.

Also we should add some context about what the link is for. The context could be initially visually hidden for screenreader but it would also benefit visually. Without clicking on the Link I have no idea what is going to happen if I click the link. Will I email the user (theres a mail address at the bottom which gives this impression)? Would it activate/delete or edit?

spec/features/providers/user_spec.rb
```diff
@@ -0,0 +1,57 @@
+require "rails_helper"
+
+feature "Provider users page" do
+  let(:provider_users_page) { PageObjects::Page::Providers::Users::IndexPage.new }
+  let(:new_access_request_page) { PageObjects::Page::Organisations::NewManualAccessRequestPage.new }
+
+  let(:current_recruitment_cycle) { build(:recruitment_cycle) }
+
+  scenario "View a provider's user page" do
```
misaka, line 9: Just being pedantic, but this is testing more than just viewing the user's page, it's testing the request access journey. :)

TobyRet: True ... We are testing that clicking on the link takes us to the 'request access' page but we do not go any further after that because request access functionality is tested elsewhere. As this is a first iteration there is still lot of functionality missing from the page. Would be worth revisiting the specs once it's fleshed out a bit more.

PR #1099, opened 2020-05-15
[3383] Use last years allocation from API

asmega:
### Context

- Now that we have imported last years allocations we can start using them instead of dummy data

- `request_type` now part of api and not via a virtual attribute
- Remove hard coded dummy data
- Confirmation page `number_of_places` placeholder now uses API data

### Guidance to review

- Depends on https://github.com/DFE-Digital/teacher-training-api/pull/1361
- Request a repeat allocation
- This allocation should have same number_of_places as previous allocation and can be seen on the confirmation page

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] ~Product Review~

### DIFF
--- a/app/controllers/providers/allocations_controller.rb
+++ b/app/controllers/providers/allocations_controller.rb
@@ -29,25 +29,19 @@ module Providers
 
     def create
       # TODO: we need to add error handling here
-      AllocationServices::Create.call(
+
+      allocation = AllocationServices::Create.call(
         accredited_body_code: @provider.provider_code,
         provider_id: @training_provider.id,
-        requested: requested?,
+        request_type: params[:request_type],
         number_of_places: params[:number_of_places],
       )
 
-      redirect_to provider_recruitment_cycle_allocation_path(
-        requested: (params[:requested].downcase unless initial_request?),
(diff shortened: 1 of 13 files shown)


asmega:

app/controllers/providers/allocations_controller.rb
```diff
@@ -29,25 +29,19 @@ def repeat_request; end
...
     def create
       # TODO: we need to add error handling here
-      AllocationServices::Create.call(
+
+      allocation = AllocationServices::Create.call(
         accredited_body_code: @provider.provider_code,
         provider_id: @training_provider.id,
-        requested: requested?,
+        request_type: params[:request_type],
         number_of_places: params[:number_of_places],
```
misaka, line 37: Where is `params[:number_of_places]` coming from? I don't see it in the form? Still working on it or am I missing something?

asmega: this endpoint is use by multiple forms to create the allocation. in repeat request we don't send this as we will use last years allocation. for new and changing of requests this form field should be available and send to api

PR #1102, opened 2020-05-18
[3383] Show initial allocations on index page

asmega:
### Context

- Allocations index page were not showing newly create initial allocations
- This changes fixes that

- Incorrect logic meant initial allocations were not showing on the
index page
- We now grab the initial allocations and iterate thru those instead

### Guidance to review

- 
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/view_objects/allocations_view.rb
+++ b/app/view_objects/allocations_view.rb
@@ -33,9 +33,8 @@ class AllocationsView
   end
 
   def initial_allocation_statuses
-    statuses = @training_providers.map do |training_provider|
-      matching_allocation = find_matching_allocation(training_provider, initial_allocations)
-      build_initial_allocations(matching_allocation, training_provider)
+    statuses = initial_allocations.map do |allocation|
+      build_initial_allocations(allocation, allocation.provider)
     end
 
     statuses.compact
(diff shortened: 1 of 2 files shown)


asmega:

app/view_objects/allocations_view.rb
```diff
@@ -97,13 +96,17 @@ def build_repeat_allocations(matching_allocation, training_provider)
...
-    {
+    hash = {
       training_provider_name: training_provider.provider_name,
       training_provider_code: training_provider.provider_code,
       status_colour: Colour::GREEN,
       requested: Requested::YES,
       status: "#{matching_allocation.number_of_places} PLACES REQUESTED",
     }
+
+    hash[:id] = matching_allocation.id if matching_allocation.id
```
misaka, line 107: Surprised by the `if matching_allocation.id` bit, as well as the `return if matching_allocation.nil?` ... why would either not be here?

asmega: yea probably not needed. i copied the other one, which probably does need it as its possible to have a request without an allocation yet

PR #1103, opened 2020-05-18
[3383] Missing inital allocations table

asmega:
### Context

- If an accredited body does not have any associated training
providers and they create a new inital allocation the table showing
these allocations is missing
- This change fixes the above
- We now use the actual allocations to determine if there are initial
allocations

### Guidance to review

- Find a provider without any associated training providers eg `B16`
- Create an initial allocation
- This allocation should be visible in the initial allocations table on the index page

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review


asmega:

PR #1111, opened 2020-05-20
[3392] Integrate notifications endpoint

gpeng:
### Context

We are adding a feature to enable a user (from an accredited body only at this stage) to enable/disable email notifications when a course that their accredited body ratifies is created or updated.

Update the existing implementation to use the new API endpoint at `/user_notification_preferences`

### Guidance to review

Runs with [this API branch](https://github.com/DFE-Digital/teacher-training-api/pull/1369) which has yet to be approved/merged. 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/view_objects/notifications_view.rb
+++ b/app/view_objects/notifications_view.rb
@@ -1,0 +1,40 @@
+class NotificationsView
+  ORGANISATION_URL_PATTERN = /\/organisations\/(\S+)\/?/.freeze
+
+  attr_reader :user_notification_preferences
+
+  def initialize(
+    request:,
+    current_user:,
+    user_notification_preferences:
+  )
+    @request = request
+    @current_user = current_user
+    @user_notification_preferences = user_notification_preferences
+  end
+
+  def user_id
+    current_user["user_id"]
+  end
+
+  def user_email
+    current_user["info"]["email"]
+  end
+
+  def cancel_link_path
+    return Rails.application.routes.url_helpers.root_path if ORGANISATION_URL_PATTERN.match(request.referer).nil?
+
+    URI(request.referer).path
+  end
+
+  def provider_code
+    matches = ORGANISATION_URL_PATTERN.match(request.referer)
+    return if matches.nil?
+
+    matches[1]
+  end
+
+private
+
+  attr_reader :request, :current_user
+end
--- a/app/views/notifications/index.html.erb
+++ b/app/views/notifications/index.html.erb
@@ -18,14 +18,27 @@
 </div>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <%= form_with url:notifications_path, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
+    <%= form_for @user_notification_preferences,
+      url: notification_path(current_user["user_id"]),
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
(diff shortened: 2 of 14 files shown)


gpeng:
[[S] Integrate Notifications endpoints into Publish](https://trello.com/c/1nDvHsy0/3392-s-integrate-notifications-endpoints-into-publish)

app/views/notifications/index.html.erb
```diff
@@ -18,14 +18,27 @@
 </div>
 <div class="govuk-grid-row">
   <div class="govuk-grid-column-full">
-    <%= form_with url:notifications_path, builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
+    <%= form_for @user_notification_preferences,
```
TobyRet, line 21: @aliuk2012 is the form supposed to have a breadcrumb, back or cancel button?

config/routes.rb
```diff
@@ -20,7 +20,7 @@
   root to: "providers#index"
   get "/organisations", to: redirect("/")
 
-  resources :notifications, path: "/notifications", controller: "notifications", only: %i[index create]
+  resources :notifications, path: "/notifications", controller: "notifications", only: %i[index update]
```
TobyRet, line 23: Is there any particular reason why (possibly this has been discussed / agreed earlier and I've forgotten) why we have gone for the provider code as a param on the notifications index page?
```
http://localhost:3000/notifications?provider_code=Y50
```

rather than 
```
http://localhost:3000/notifications/Y50
```

I noticed that when I modified the provider code param, nothing would happen but I would expect either a 404 or 403

gpeng: It's only used to redirect the user back from where they came from. Did it redirect you to an existing provider?

gpeng: We could probably implement that better really and avoid the parameter being in the URL. Maybe fish it out of the referrer?

app/views/notifications/index.html.erb
aliuk2012: I do recall there being some form of navigation. Let me me do some digging around and I'll get back to you.

aliuk2012: We do need a "Cancel link" at the bottom of the form to take the user back to the previous page.

https://publish-courses-pr-44.herokuapp.com/notifications

config/routes.rb
aliuk2012: If theres a better way then feel free to do it. I was just using it to get the basic functionality working. The idea is that we need to take them back to the previous page they were on (i.e the providers page) and display a success message

app/view_objects/notifications_view.rb
```diff
@@ -0,0 +1,40 @@
...
+  end
+
+  def user_email
+    current_user["info"]["email"]
+  end
+
+  def cancel_link_path
+    return Rails.application.routes.url_helpers.root_path if ORGANISATION_URL_PATTERN.match(request.referer).nil?
+
+    URI(request.referer).path
```
asmega, line 27: does it explode if for some reason if can't determine the referrer? or is that not possible? also i think it's spelt `referrer`?

AgaDufrat: If you go directly to http://localhost:3000/notifications `request.referer` is nil and cancel_link_path is root_path

PR #1112, opened 2020-05-20
[3422] Remove own provider from training providers

asmega:
### Context

- https://trello.com/c/Hwg5ZBqT/3422-accredited-bodies-should-be-able-to-request-pe-allocations
- The API has changed and now training providers returns itself if there offer their won courses
- There is a view in publish where we do not want the provider itself to be show and has therefore been removed on the frontend

- There is an API change and now training providers list returns the
current provider if they themselves offer courses
- We don't want this in publish training providers page so has been
removed here

### Guidance to review

- Depends on https://github.com/DFE-Digital/teacher-training-api/pull/1374
- Find a provider that also has their own courses eg http://localhost:3000/organisations/2AT/2020/training-providers
- They should not see themselves on the training providers page

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -67,11 +67,13 @@ class ProvidersController < ApplicationController
 
   def training_providers
     @training_providers = @provider.training_providers(recruitment_cycle_year: @recruitment_cycle.year)
+    @training_providers.delete_if { |tp| tp.provider_code == @provider.provider_code }
 
     courses = Course.where(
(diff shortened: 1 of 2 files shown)


asmega:
[Accredited bodies should be able to request PE allocations](https://trello.com/c/Hwg5ZBqT/3422-accredited-bodies-should-be-able-to-request-pe-allocations)

PR #1123, opened 2020-05-22
[3398] validation no provider

asmega:
### Context

- https://trello.com/c/o5I1mnwd/3398-m-adding-validation-to-who-are-you-requesting-a-course-for-page
- There is missing validation on the first page of the initial request journey when the user does not select a radio option and continues

### Changes proposed in this pull request

- Add validation to initial request flow
- Fix deprecation warning when running test suite
- Added .rspec_parallel config to speed up local tests

### Guidance to review

- Create an initial allocation request
- On the first page do not select a radio option and continue
- Should see validation message

### Checklist

- [x] Make sure all information from the Trello card is in here
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Tested by running locally
- [ ] Product Review

PR #1124, opened 2020-05-23
[3399] - Initial allocation flow: provider search validation

TobyRet:
### Context
API has been updated to allow for two character provider searches. See https://github.com/DFE-Digital/teacher-training-api/pull/1378

- During the initial allocation flow, a validation error is displayed on the form if the accredited body searches for a provider with a query string containing only one character

### Guidance to review
- Run this branch against https://github.com/DFE-Digital/teacher-training-api/pull/1378
- Request an initial allocation. Select other provider and enter a single character in the search field. Validation error should be displayed

### Notes
- Need some copy guidance on the error message
- I've followed the existing validation pattern for `blank_search_query`

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/flows/initial_request_flow.rb
+++ b/app/flows/initial_request_flow.rb
@@ -74,7 +74,7 @@ private
   end
 
   def training_provider_search?
-    params[:training_provider_code] == "-1" && params[:training_provider_query].present?
+    params[:training_provider_code] == "-1" && params[:training_provider_query].present? && params[:training_provider_query].size > 1
   end
 
   def one_search_result?
@@ -173,7 +173,7 @@ private
 
   def number_of_places_page?
     training_provider_selected? ||
-      (params[:training_provider_query].present? && one_search_result?) || params[:change]
+      (params[:training_provider_query].present? && params[:training_provider_query].size > 1 && one_search_result?) || params[:change]
   end
 
   def check_your_information_page?
--- a/app/form_objects/initial_request_form.rb
+++ b/app/form_objects/initial_request_form.rb
@@ -5,6 +5,7 @@ class InitialRequestForm
 
   validates :training_provider_code, presence: { message: "Select or search for an organisation" }
   validates :training_provider_query, presence: { message: "You need to add some information", if: :provider_search? }
+  validates :training_provider_query, length: { minimum: 2, message: "Please enter a minimum of two characters", if: :provider_search? }
 
   def add_no_results_error
     errors.add(
--- a/spec/features/providers/allocations/initial_allocation_spec.rb
+++ b/spec/features/providers/allocations/initial_allocation_spec.rb
@@ -141,6 +141,27 @@ RSpec.feature "PE allocations" do
     and_i_see_error_message_that_i_must_add_more_info
   end
 
+  scenario "Accredited body searches for provider with string containing only one character" do
+    given_accredited_body_exists
+    given_the_accredited_body_has_an_allocation
+    given_there_is_a_training_provider_with_previous_allocations
+    # once the feature is released it should be changed to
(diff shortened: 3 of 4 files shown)


TobyRet:
[[S] - Error handling for 1 character provider searches](https://trello.com/c/55f2oRuQ/3434-s-error-handling-for-1-character-provider-searches)

app/flows/initial_request_flow.rb
```diff
@@ -189,6 +194,18 @@ def training_provider
                            end
   end
 
+  def invalid_search_query?
+    return @invalid_search_query if @invalid_search_query
```
asmega, line 198: I'm trying to move the validation logic into the form object where i think it better belongs. see https://github.com/DFE-Digital/publish-teacher-training/pull/1123/files

TobyRet: Ah brilliant - form object sounds like a good idea. Was going to ask about his possibly refactoring. 
I'll rebase onto your changes 👍

misaka:
I was testing this out in the review app, and was still getting a 500 when I tried searching with 2 characters (see [Sentry PUBLISH-TEACHER-TRAINING-74](https://sentry.io/organizations/dfe-bat/issues/1692602858/?environment=qa&project=1407453&query=is%3Aunresolved)).

I guess this just because the review app uses the qa api, which still needs 3+ chars?

app/flows/initial_request_flow.rb
```diff
@@ -10,12 +10,12 @@ def initialize(params:)
   end
 
   def template
-    if check_your_information_page?
+    if invalid_provider_search_query? || empty_search_results?
```
misaka, line 13: Just noticing that neither `valid?` nor `validate` is being called on the form object inside of this flow. Looking at [`AllocationsController#initial_request`](https://github.com/DFE-Digital/publish-teacher-training/blob/3399-two-character-provider-search/app/controllers/providers/allocations_controller.rb#L61):

```ruby
    def initial_request
      flow = InitialRequestFlow.new(params: params)

      if request.post? && flow.valid? && flow.redirect?
        redirect_to flow.redirect_path
      else
        render flow.template, locals: flow.locals
      end
    end
```

it looks as if `template` could be called without `flow.valid?` being called, if for example `request.post?` returns `false`.

Did I read that right, and is it a bug?

Would it be worth always validating `form_object` before any of the predicates here, or in other methods, are called on it?

TobyRet:
> I was testing this out in the review app, and was still getting a 500 when I tried searching with 2 characters (see [Sentry PUBLISH-TEACHER-TRAINING-74](https://sentry.io/organizations/dfe-bat/issues/1692602858/?environment=qa&project=1407453&query=is%3Aunresolved)).
> 
> I guess this just because the review app uses the qa api, which still needs 3+ chars?

Yep that's right - you will need to run locally against https://github.com/DFE-Digital/teacher-training-api/pull/1378

app/flows/initial_request_flow.rb
TobyRet: Currently the validations only apply to the "Who are you requesting a course for?" page. The continue button submits a POST and validation always kicks in ... If no option is selected or 'search for a provider' is selected and the query is either blank or one character long then the validation errors will show. 

I'm not sure if this may become an issue further down the link if validations are added on other pages ... and / or whether other validations are being worked on currently. @asmega do you have a view on this?

asmega: validation should only kick if the user submits the form. if validation occurs on a GET request of the form the user will see a load of errors on the loading of the empty form before they've filled it in

PR #1127, opened 2020-05-26
[3402] Tweak Initial request form when there are no associated providers

aliuk2012:
### Context

- https://trello.com/c/6Znq0sUB/3402-when-there-are-no-associated-providers
- On the initial request journey on the first page when there are no associated providers the radio button divider would show and only one radio button with the provider search option

- When there are no associated providers hide the radio buttons and only show the search field

### Guidance to review

- Create a new initial request for a provider that does not have any associated providers
- On the first page should not show radio buttons and only search field should be present in the form 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/flows/initial_request_flow.rb
+++ b/app/flows/initial_request_flow.rb
@@ -32,6 +32,7 @@ class InitialRequestFlow
       {
         training_providers: training_providers_without_associated,
         form_object: form_object,
+        provider: provider,
       }
     elsif pick_a_provider_page?
       {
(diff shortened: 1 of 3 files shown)


asmega:
[[S] When there are no associated providers](https://trello.com/c/6Znq0sUB/3402-when-there-are-no-associated-providers)

PR #1129, opened 2020-05-27
[3442] Allow all providers to create a new course

AgaDufrat:
### Context

**User story**
```
As a provider 
I need to publish a course and choose an accrediting body
So that I can recruit and have the course accredited by another provider
```
* If the accrediting body has opted-in to receive email notifications they are notified when a course is created - Done in #1111

* ALL providers can now create courses i.e. no more google form

### Guidance to review
- Try to create a course for a provider that is not an accredited body

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/helpers/provider_helper.rb
+++ b/app/helpers/provider_helper.rb
@@ -13,10 +13,6 @@ module ProviderHelper
     govuk_link_to("Add a new course", new_provider_recruitment_cycle_course_path(provider_code: provider.provider_code, recruitment_cycle_year: provider.recruitment_cycle_year), class: "govuk-button govuk-!-margin-bottom-2", **opts)
   end
 
-  def add_course_link(email, provider, is_current_cycle:, **opts)
-    link_to "Add a new course", add_course_url(email, provider, is_current_cycle: is_current_cycle), class: "govuk-button govuk-!-margin-bottom-2", rel: "noopener noreferrer", target: "_blank", **opts
-  end
-
   def google_form_url_for(settings, email, provider)
     settings.url + "&" +
       { settings.email_entry => email, settings.provider_code_entry => provider.provider_code }.to_query
--- a/app/helpers/provider_helper.rb
+++ b/app/helpers/provider_helper.rb
@@ -9,10 +9,6 @@ module ProviderHelper
     end
   end
 
-  def create_course_link(provider, **opts)
-    govuk_link_to("Add a new course", new_provider_recruitment_cycle_course_path(provider_code: provider.provider_code, recruitment_cycle_year: provider.recruitment_cycle_year), class: "govuk-button govuk-!-margin-bottom-2", **opts)
-  end
-
   def google_form_url_for(settings, email, provider)
     settings.url + "&" +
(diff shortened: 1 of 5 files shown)


app/views/courses/index.html.erb
```diff
@@ -15,12 +15,7 @@
 </ul>
 
 <% if @provider.courses.count > 10 %>
-  <% if @provider.accredited_body? %>
-    <%= create_course_link(@provider, data: {qa: "course-create"} ) %>
-  <% else %>
-    <%= add_course_link(current_user['info']['email'], @provider, is_current_cycle: @recruitment_cycle.current?, data: {qa: "course-create-additional"}) %>
-    <p class="govuk-body-s govuk-!-margin-bottom-6">You’ll be taken to a Google Form to fill in your course details.</p>
-  <% end %>
+  <%= create_course_link(@provider, data: {qa: "course-create"} ) %>
```
aliuk2012, line 18: could we remove the `data-qa` attribute and update siteprism?

app/views/courses/index.html.erb
```diff
@@ -40,9 +35,4 @@
   </section>
 <% end %>
 
-<% if @provider.accredited_body? %>
-  <%= create_course_link(@provider, data: {qa: "course-create"} ) %>
-<% else %>
-  <%= add_course_link(current_user['info']['email'], @provider, is_current_cycle: @recruitment_cycle.current?, data: {qa: "course-create"}) %>
-  <p class="govuk-body-s">You’ll be taken to a Google Form to fill in your course details.</p>
-<% end %>
+<%= create_course_link(@provider, data: {qa: "course-create"} ) %>
```
aliuk2012, line 38: Same as above

app/helpers/provider_helper.rb
```diff
@@ -13,10 +13,6 @@ def create_course_link(provider, **opts)
     govuk_link_to("Add a new course", new_provider_recruitment_cycle_course_path(provider_code: provider.provider_code, recruitment_cycle_year: provider.recruitment_cycle_year), class: "govuk-button govuk-!-margin-bottom-2", **opts)
   end
```
aliuk2012, line 14: As we are only using this twice in `app/views/courses/index.html.erb`, I think we should remove this helper method and just call this twice.  If we needed to use a create course link would we remember that this helper even exists if its only used twice?

aliuk2012:
Will providers still need to use a google form to create PE courses for secondary schools?

app/helpers/provider_helper.rb
aliuk2012, line 14: As we are only using this twice in app/views/courses/index.html.erb, I think we should remove this helper method and just call this twice. If we needed to use a create course link would we remember that this helper even exists if its only used twice?

AgaDufrat: Good point, I removed `create_course_link`.

I think there is still value in having the `add_course_url` helper as there is some logic in there that we unit test. I think we may also remove it in a few weeks once the allocations feature is live so I would leave it as it is for now.

PR #1131, opened 2020-05-27
[3399] - Initial allocation flow provider search validation

asmega:
- Error is displayed if the accredited body searches for a provider with
a query string containing only one character
- Also removes duplicated validation-check logic from the
InitialRequestFlow

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


asmega:
no longer needed

PR #1133, opened 2020-05-28
[3454] Add validations to repeat allocations

aliuk2012:
Fixes an 500 error a user would get if they did not select yes/no.
This commit also removed the "allocation" namespace we added awhile ago.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/views/providers/allocations/edit.html.erb
+++ b/app/views/providers/allocations/edit.html.erb
@@ -12,6 +12,7 @@
                     id: @allocation.id
                   ),
                   model: @allocation,
+                  scope: "",
                   method: :patch,
                   builder: GOVUKDesignSystemFormBuilder::FormBuilder do |form| %>
       <%= form.govuk_radio_buttons_fieldset(:old_department_id, legend: { size: 'xl', text: page_title }) do %>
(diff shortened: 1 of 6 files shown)


aliuk2012:
[Add validations to "Do you want to make a request?" page](https://trello.com/c/kfEpSkxe/3454-add-validations-to-do-you-want-to-make-a-request-page)

app/views/providers/allocations/edit.html.erb
```diff
@@ -12,6 +12,7 @@
                     id: @allocation.id
                   ),
                   model: @allocation,
+                  scope: "",
```
TobyRet, line 15: What is the rationale behind adding an empty string as a scope?

aliuk2012: The reason scope is an empty string is that if scope if not defined then it uses the model class name as a namespace for the various fields that the from builder renders. Normally this shouldn't be a problem but because @allocation could be either a InitialRequestForm or RepeatRequestFrom class and potentially others where we reuse the markup it becomes difficult to use  one single namespace. So in this instance we clear the namespace with an empty string.

PR #1139, opened 2020-05-29
[3440] Check for existing allocation then proceed

asmega:
### Context

- https://trello.com/c/dtbw2raY/3440-stop-filtering-out-providers-from-provider-search-initial-allocation-request
- When searching for a provider for an allocation we now return providers that also have allocations
- As a result if one of these are selected we now take the user to modify the allocation

- When creating a new allocation we now check if an allocation already
exists
- If yes, proceed to edit the existing allocation
- If now, proceed with the current journey to create the allocation

### Guidance to review

- Depends on https://github.com/DFE-Digital/teacher-training-api/pull/1388
- Create a new allocation. Search for a provider that already has an initial, repeat or allocation
- With Javascript the autocomplete will select the provider and on submit will take the user to edit the existing allocation
- Without Javascript the user is take to the pick a provider page where on clicking on the provider they are taken to the edit allocation page
- If the pick a provider page only contains one provider you will be redirected to edit the existing allocation

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/.rubocop.yml
+++ b/.rubocop.yml
@@ -27,3 +27,5 @@ Style/HashTransformKeys:
   Enabled: true
 Style/HashTransformValues:
   Enabled: false
+Rails/FindBy:
+  Enabled: false
--- a/app/flows/initial_request_flow.rb
+++ b/app/flows/initial_request_flow.rb
@@ -36,7 +36,7 @@ class InitialRequestFlow
       }
     elsif pick_a_provider_page?
       {
-        training_providers: training_providers_from_query_without_associated,
+        training_providers: training_providers_from_query,
       }
     else
       {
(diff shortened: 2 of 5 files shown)


asmega:
[Stop filtering out providers from provider search - initial allocation request](https://trello.com/c/dtbw2raY/3440-stop-filtering-out-providers-from-provider-search-initial-allocation-request)

.rubocop.yml
```diff
@@ -27,3 +27,5 @@ Style/HashTransformKeys:
   Enabled: true
 Style/HashTransformValues:
   Enabled: false
+Rails/FindBy:
```
TobyRet, line 30: Worth a separate commit as this is a config change?

app/views/providers/allocations/pick_a_provider.html.erb
```diff
@@ -14,8 +14,18 @@
     <ul class="govuk-list govuk-list--bullet">
       <% training_providers.each do |training_provider| %>
         <li>
-          <%= govuk_link_to "#{training_provider.provider_name} (#{training_provider.provider_code})",
-            initial_request_provider_recruitment_cycle_allocations_path(training_provider_code: training_provider.provider_code) %>
+          <% allocation = Allocation.for_provider_and_training_provider(recruitment_cycle: @recruitment_cycle, provider: @provider, training_provider: training_provider) %>
```
TobyRet, line 17: Is there any particular reason why we are running an active record query from the view, rather than returning the data from the controller?

app/flows/initial_request_flow.rb
```diff
@@ -51,14 +51,23 @@ def redirect?
     (training_provider_search? && one_search_result?) || training_provider_selected?
   end
 
+  def allocation
+    if (training_provider_search? && one_search_result?) || training_provider_selected?
+      Allocation.where(recruitment_cycle_year: recruitment_cycle.year)
```
TobyRet, line 56: Worth moving this query to a scope?

app/views/providers/allocations/pick_a_provider.html.erb
asmega: - theres no ActiveRecord here its the json api
- the api as it stands currently does not allow for loading of allocations along with training providers which is why unfortunately it is done here in the view

app/flows/initial_request_flow.rb
asmega: If you can point me in the direction of example of how this is done in json api client I will make the change

.rubocop.yml
asmega: I've now stuck this into another commit

app/views/providers/allocations/pick_a_provider.html.erb
TobyRet: Yeah sorry I mean json api 😄

TobyRet: Ok thanks for the explanation

app/flows/initial_request_flow.rb
TobyRet: Yeah forget that - mistaken this for the API

PR #1140, opened 2020-05-31
[3433] - Add validations to "How many places would you like to request?" page

TobyRet:
### Context
When the accredited body enters the number of places they require for an initial allocation, we can only accept positive integers greater than 0.

Validation error shown on "How many places would you like to request page" if user enters the following values in the 'number of places' field:
- empty value (e.g "")
- 0
- non-integer (e.g. "a")

### Guidance to review
- Go to a provider's allocation page (e.g. `/organisations/K30/2020/allocations/) and select the 'initial allocation' flow. Once you reach the 'How many places would you like to request' page, please try and break it.
- Was considering using HTML `input=number` to handle some of the validation but apparently this has [fallen out of favour due to accessibility concerns](https://technology.blog.gov.uk/2020/02/24/why-the-gov-uk-design-system-team-changed-the-input-type-for-numbers/ ).

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/flows/initial_request_flow.rb
+++ b/app/flows/initial_request_flow.rb
@@ -12,7 +12,7 @@ class InitialRequestFlow
   def template
     if check_your_information_page?
       "providers/allocations/check_your_information"
-    elsif number_of_places_page?
+    elsif number_of_places_page? || number_of_places_zero?
       "providers/allocations/number_of_places"
     elsif blank_search_query? || empty_search_results?
       "providers/allocations/initial_request"
@@ -27,6 +27,7 @@ class InitialRequestFlow
     if number_of_places_page? || check_your_information_page?
       {
         training_provider: training_provider,
+        form_object: form_object,
       }
     elsif blank_search_query? || empty_search_results?
       {
@@ -48,6 +49,8 @@ class InitialRequestFlow
   end
 
   def redirect?
+    return false if valid_number_of_places?
+
     (training_provider_search? && one_search_result?) || training_provider_selected?
   end
 
@@ -71,6 +74,14 @@ class InitialRequestFlow
 
 private
 
+  def valid_number_of_places?
+    params[:number_of_places] && !params[:number_of_places].include?(".") && params[:number_of_places].to_i.positive?
+  end
+
+  def number_of_places_zero?
+    params[:number_of_places] && params[:number_of_places].to_i.zero?
+  end
+
   def training_provider_selected?
     params[:training_provider_code].present? && params[:training_provider_code] != "-1"
   end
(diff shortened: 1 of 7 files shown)


TobyRet:
https://trello.com/c/KixfjXm5/3433-s-add-validations-to-how-many-places-would-you-like-to-request-page

app/flows/initial_request_flow.rb
```diff
@@ -179,7 +190,7 @@ def number_of_places_page?
   end
 
   def check_your_information_page?
-    training_provider_selected? && params[:number_of_places].present? && !params[:change]
+    training_provider_selected? && (params[:number_of_places].present? && params[:number_of_places].to_i > 0) && !params[:change]
```
aliuk2012, line 193:
```suggestion
    training_provider_selected? && valid_number_of_places? && !params[:change]
```

app/views/providers/allocations/number_of_places.html.erb
```diff
@@ -7,13 +7,14 @@
...
   <div class="govuk-grid-column-two-thirds">
 
-  <%= form_with url: initial_request_provider_recruitment_cycle_allocations_path(
+  <%= form_with  model: form_object, scope: "", url: initial_request_provider_recruitment_cycle_allocations_path(
                     @provider.provider_code,
                     @provider.recruitment_cycle_year,
                   ),
                 skip_enforcing_utf8: true,
-                method: :get,
+                # method: :get,
```
asmega, line 15: delete it or leave it otherwise the next person will be confused

app/form_objects/initial_request_form.rb
```diff
@@ -20,4 +21,16 @@ def add_no_results_error
...
+  def selected_number_of_places
+    return if number_of_places.nil?
+
+    if number_of_places_invalid?
+      errors.add(:number_of_places, "You must enter a number")
+    end
+  end
+
+  def number_of_places_invalid?
+    number_of_places == "" || !/\A\d+\z/.match(number_of_places) || number_of_places.to_i < 1
```
asmega, line 34: would it be better to validate what it should be and not what it should not be? abstractly something has a small set to what it should be and infinite set of what it should not be. i think capturing with whitelist would comprehend better

app/views/providers/allocations/number_of_places.html.erb
TobyRet: oops ... should have deleted

app/flows/initial_request_flow.rb
TobyRet: 👍

app/form_objects/initial_request_form.rb
TobyRet: Yep good point, I'll change

app/form_objects/initial_request_form.rb
```diff
@@ -20,4 +21,15 @@ def add_no_results_error
...
   end
+
+  def selected_number_of_places
+    return if number_of_places.nil?
+
+    errors.add(:number_of_places, "You must enter a number") unless number_of_places_valid?
+  end
+
+  def number_of_places_valid?
+    # binding.pry
```
aliuk2012, line 32: Needs to be deleted

TobyRet: gah!
Fixed

PR #1142, opened 2020-06-01
[3426] change initial allocation requests

SamJamCul:
### Context
https://trello.com/c/T6kURpH0/3426-m-change-initial-allocation-requests

This isn't using the same flow setup as creating an initial allocation, as we ran into some problems when trying to implement it. Instead we went for a new controller! There may be a case for refactoring later on to bring both of them together somehow. 

* Add edit_initial_allocations controller 
* Add templates for editing initial allocations 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers/edit_initial_allocations_controller.rb
+++ b/app/controllers/providers/edit_initial_allocations_controller.rb
@@ -1,0 +1,125 @@
+module Providers
+  class EditInitialAllocationsController < ApplicationController
+    def do_you_want
+      if request.post?
+        case params[:allocation][:request_type]
+        when AllocationsView::RequestType::INITIAL
+          redirect_to number_of_places_provider_recruitment_cycle_allocation_path(
+            provider_code: allocation.accredited_body.provider_code,
+            recruitment_cycle_year: recruitment_cycle.year,
+            training_provider_code: allocation.provider.provider_code,
+            id: allocation.id,
+          )
+        when AllocationsView::RequestType::DECLINED
+          destroy
+          redirect_to confirm_deletion_provider_recruitment_cycle_allocation_path(
+            provider_code: provider.provider_code,
+            recruitment_cycle_year: recruitment_cycle.year,
+            training_provider_code: training_provider.provider_code,
+          )
+        end
+      else
+        render locals: {
+          training_provider: training_provider,
+          allocation: allocation,
+          provider: provider,
+        }
+      end
+    end
+
+    def confirm_deletion
+      @allocation = Allocation.new(request_type: AllocationsView::RequestType::DECLINED)
+      @training_provider = training_provider
+      @provider = provider
+      @recruitment_cycle = recruitment_cycle
+      render template: "providers/allocations/show"
+    end
+
+    def number_of_places
+      if params.dig("allocation", "number_of_places")
+        allocation.number_of_places = params.dig("allocation", "number_of_places")
+      end
+
+      if request.post? && allocation.valid?
+        redirect_to check_answers_provider_recruitment_cycle_allocation_path(
+          provider_code: allocation.accredited_body.provider_code,
+          recruitment_cycle_year: recruitment_cycle.year,
+          training_provider_code: allocation.provider.provider_code,
+          number_of_places: params[:allocation][:number_of_places],
+          id: allocation.id,
+        )
+      else
+        render locals: {
+          training_provider: training_provider,
+          provider: provider,
+          allocation: allocation,
+          recruitment_cycle: recruitment_cycle,
+        }
+      end
+    end
+
+    def check_answers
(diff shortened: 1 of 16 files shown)


app/controllers/providers/edit_initial_allocations_controller.rb
```diff
@@ -0,0 +1,119 @@
+module Providers
+  class EditInitialAllocationsController < ApplicationController
+    def do_you_want
```
TobyRet, line 3: Interesting method name 😝

app/controllers/providers/edit_initial_allocations_controller.rb
```diff
@@ -0,0 +1,119 @@
...
+        case params[:allocation][:request_type]
+        when AllocationsView::RequestType::INITIAL
+          redirect_to number_of_places_provider_recruitment_cycle_allocation_path(
+            provider_code: allocation.accredited_body.provider_code,
+            recruitment_cycle_year: recruitment_cycle.year,
+            training_provider_code: allocation.provider.provider_code,
+            id: allocation.id,
+          )
+        when AllocationsView::RequestType::DECLINED
+          destroy
```
TobyRet, line 14: The allocation is destroyed before the user has a chance to confirm?

TobyRet: Ah I see - it redirects to the a page confirming the deletion ...
Worth being consistent with naming here? delete vs destroy

app/controllers/providers/edit_initial_allocations_controller.rb
```diff
@@ -0,0 +1,119 @@
...
+      else
+      render locals: {
+        training_provider: training_provider,
+        allocation: allocation,
+        provider: provider,
+      }
+      end
+    end
+
+    def confirm_deletion
```
TobyRet, line 30: 'delete_confirmation` or 'delete_successful' may signal the intent better? i.e. you are seeing a page confirming deletion has been successful

SamJamCul:
[[M] Change "initial" allocation requests.](https://trello.com/c/T6kURpH0/3426-m-change-initial-allocation-requests)

app/controllers/providers/edit_initial_allocations_controller.rb
misaka: I agree with @TobyRet that this is ... "interesting". :) Controllers are usually named after resources and methods are usually actions, so I'd expect that these are in the existing `AllocationsController` with a service object to wrap up all the messiness, or if really need be, in a `InitialAllocationController` or something that expresses the thing being worked on, instead of the actions being performed.

app/controllers/providers/edit_initial_allocations_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+            recruitment_cycle_year: recruitment_cycle.year,
+            training_provider_code: allocation.provider.provider_code,
+            id: allocation.id,
+          )
+        when AllocationsView::RequestType::DECLINED
+          destroy
+          redirect_to confirm_deletion_provider_recruitment_cycle_allocation_path(
+            provider_code: provider.provider_code,
+            recruitment_cycle_year: recruitment_cycle.year,
+            training_provider_code: training_provider.provider_code,
```
misaka, line 18: On line 10 above, we get the training provider code like this:

```ruby
            training_provider_code: allocation.provider.provider_code,
```

Which makes sense because it avoids doing a request to the API. Would it make sense to do that here too?

app/controllers/providers/edit_initial_allocations_controller.rb
```diff
@@ -0,0 +1,125 @@
...
+        redirect_to check_answers_provider_recruitment_cycle_allocation_path(
+          provider_code: allocation.accredited_body.provider_code,
+          recruitment_cycle_year: recruitment_cycle.year,
+          training_provider_code: allocation.provider.provider_code,
+          number_of_places: params[:allocation][:number_of_places],
+          id: allocation.id,
+        )
+      else
+        render locals: {
+          training_provider: training_provider,
```
misaka, line 53: In the `allocation` method below we include the `provider` in our request to retrieve the allocation, which is the training provider, so could we use that here and anywhere else we need `training_provider` so that we can avoid making an extra request to the API?

I think a good approach could be to use a `build_allocation` hook to get the `allocation`, and then a `build_training_provider` hook to use the provider set in the allocation, if that makes sense.

app/controllers/providers/edit_initial_allocations_controller.rb
SamJamCul: Yes, makes sense to make this standardise across all these changes actually

PR #1149, opened 2020-06-03
[3433] - Add new 'number_of_places' validation

TobyRet:
### Context
Discovered that a combination of letter and number was still considered 'valid' when filling out the 'number of places' field when creating or updating an initial allocation

- Adds regex to validate that 'number of places' only contains numbers
- Add Allocation model validation specs
- Remove unused code from Allocation model

### Guidance to review
When you fill out the 'number of places' field with a letter-number combination (e.g. '3a') you should now see a validation error.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/models/allocation.rb
+++ b/app/models/allocation.rb
@@ -29,13 +29,6 @@ class Allocation < Base
     number_of_places.to_i.positive?
   end
 
-  def add_invalid_number_of_places_errors
-    errors.add(
-      :number_of_places,
-      "We could not find this organisation - please check your information and try ",
-    )
-  end
-
   def repeat_request?
     request_type == Allocation::RequestTypes::REPEAT
   end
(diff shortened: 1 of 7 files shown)


app/models/allocation.rb
```diff
@@ -29,13 +29,6 @@ def has_places?
     number_of_places.to_i.positive?
   end
 
-  def add_invalid_number_of_places_errors
```
TobyRet, line 32: This was mistakenly added in https://github.com/DFE-Digital/publish-teacher-training/pull/1142

app/models/allocation.rb
```diff
@@ -29,13 +29,6 @@ def has_places?
...
-    errors.add(
-      :number_of_places,
-      "We could not find this organisation - please check your information and try ",
```
TobyRet, line 35: Just try!

TobyRet:
[[S] Add validations to "How many places would you like to request?" page](https://trello.com/c/KixfjXm5/3433-s-add-validations-to-how-many-places-would-you-like-to-request-page)

PR #1171, opened 2020-06-09
Open the secondary subject field when present

fofr:
### Context

When editing a secondary course with multiple subjects, the second subject field is closed, even if a subject is selected.

- Open the details element when the second subject is present
- Remove inline styles, replace with classes
- Tweak spacing
- Fix naming of test variable

(Ignore whitespace when reviewing)

https://trello.com/c/RyXdo7Ry/3512-open-the-secondary-subject-field-when-present

### DIFF
--- a/app/views/courses/subjects/_form_fields.html.erb
+++ b/app/views/courses/subjects/_form_fields.html.erb
@@ -16,24 +16,22 @@
     </div>
 
     <% if course.level == "secondary" %>
-      <div style="margin-bottom: -20px">
-        <details class="govuk-details" role="group" data-qa="course__subordinate_subject_accordion">
-          <summary class="govuk-details__summary" role="button">
-            <span class="govuk-details__summary-text">
-              Add a second subject (optional)
-            </span>
-          </summary>
-          <div class="govuk-details__text" id="details-content" aria-hidden="true">
-            <p>Your first subject is the main one. It will appear first in the course title. It represents the bursary or scholarship available if applicable.</p>
+      <details class="govuk-details govuk-!-margin-top-6 govuk-!-margin-bottom-0" role="group" data-qa="course__subordinate_subject_accordion">
+        <summary class="govuk-details__summary" role="button">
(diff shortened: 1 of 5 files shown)


app/views/courses/subjects/_form_fields.html.erb
```diff
@@ -16,24 +16,22 @@
...
-          </summary>
-          <div class="govuk-details__text" id="details-content" aria-hidden="true">
-            <p>Your first subject is the main one. It will appear first in the course title. It represents the bursary or scholarship available if applicable.</p>
+      <details class="govuk-details govuk-!-margin-top-6 govuk-!-margin-bottom-0" role="group" data-qa="course__subordinate_subject_details" <% if course.selected_subject_ids.second %>open<% end %>>
+        <summary class="govuk-details__summary" role="button">
+          <span class="govuk-details__summary-text">
+            Add a second subject (optional)
+          </span>
+        </summary>
+        <div class="govuk-details__text" id="details-content" aria-hidden="true">
```
aliuk2012, line 25: shouldn't the aria-hidden be set to "false" if the details element is open

app/views/courses/subjects/_form_fields.html.erb
```diff
@@ -16,24 +16,22 @@
...
+          <p>Your first subject is the main one. It will appear first in the course title. It represents the bursary or scholarship available if applicable.</p>
```
aliuk2012, line 26: should this not be using `govuk-body` class?

fofr: Probably, I didn't change this line but will add

app/views/courses/subjects/_form_fields.html.erb
fofr: Good catch

fofr: Going to remove this attribute as it isn't needed (it's not in the design system component)

PR #1178, opened 2020-06-11
[3336] - Accredited Body notifications feature interruption screen

TobyRet:
### Context

Creation of a new interruption screen for users who are associated with an accredited body. 

The interruption screen will always appear when they log in to Publish until they click "continue"
Once they have clicked continue, the interruption page will no longer appear when they log in to Publish

- Refactoring. The API should not have to manage state transitions and should be the responsibility of Publish. Publish now understands the state transitions and updates the user's state through a single API endpoint (users#update)
- Setting notification preferences  via the UI will now also update the user's state to `notifications_configured` so that the new interruption screen is not show to the user

### Guidance to review
- This PR relies on an API change - https://github.com/DFE-Digital/teacher-training-api/pull/1429
- To test the changes you will need to reinstate DFE sign in locally and update the states of your user (`:new`, `:transitioned`, `:rolled_over`, `:notifications_configured`) to view the various screens.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review


app/controllers/sessions_controller.rb
```diff
@@ -115,8 +115,15 @@ def redirect_to_correct_page(user)
...
     elsif Settings.rollover && user&.state == "transitioned"
       redirect_to rollover_path
+    elsif user&.state == "rolled_over" && user_associated_with_accredited_body?
+      redirect_to accredited_body_new_features_path
     else
       redirect_to session[:redirect_back_to] || root_path
     end
   end
+
+  def user_associated_with_accredited_body?
```
TobyRet, line 125: the `user` is actually a `session`. We can only display the `accredited_body_new_features` page if the user is associated with an accredited body. Where is the most appropriate place for this logic?

app/controllers/users_controller.rb
```diff
@@ -1,15 +1,23 @@
...
+    update_state(Settings.rollover ? rollover_path : providers_path, update_params)
   end
 
   def accept_rollover
-    accept_screen("accept_rollover_screen", providers_path)
+    update_params = { state: next_state }
+    update_state(providers_path, update_params)
+  end
+
+  def accept_accredited_body_new_features
```
TobyRet, line 12: These controller actions have very similar logic. Do we want to consolidate to have one action or keep separate (i.e. is it important for us to track the different actions via different end points?)

app/controllers/users_controller.rb
```diff
@@ -33,6 +49,20 @@ def accept_screen(method, path)
     end
   end
 
+  def next_state
```
TobyRet, line 52: We can't use `aasm` on Publish and TBH a simple approach should suffice?

app/controllers/users_controller.rb
```diff
@@ -33,6 +49,20 @@ def accept_screen(method, path)
...
+    case user_state
+    when "new"
+      "transitioned"
+    when "transitioned"
+      "rolled_over"
+    when "rolled_over"
+      "features_accepted"
+    else
+      # go back to previous page with errors?
+      raise "unknown state"
```
TobyRet, line 62: I'm assuming it's not likely this branch would be reached but if it did what should the behaviour be?

app/controllers/users_controller.rb
misaka: I think we can [define `aasm_write_state`](https://github.com/aasm/aasm/blob/master/lib/aasm/persistence/plain_persistence.rb#L14) to show AASM how to persist the state value.

misaka: (but, again, we don't _have_ to use aasm, but it might be tidier than doing it this way ... worth investigating)

app/views/pages/accredited_body_new_features.html.erb
```diff
@@ -21,5 +21,8 @@
     <p class="govuk-body">
       <%= link_to "Continue to your courses without signing up", root_path, class:"govuk-link" %>
     </p>
+    <%= form_for :user, url: accept_accredited_body_new_features_path, method: :patch do |f| %>
+      <%= f.submit "Continue", class: "govuk-button", data: { qa: 'rollover__continue' } %>
+    <% end %>
```
misaka, line 26: Don't forget to remove the link above.

SamJamCul:
[Only show "kept" users on org#users list](https://trello.com/c/Z9uH4OZB/3532-only-show-kept-users-on-orgusers-list)

SamJamCul:
[Only show "kept" users on org#users list](https://trello.com/c/Z9uH4OZB/3532-only-show-kept-users-on-orgusers-list)

TobyRet:
[[M] Support interruption screen](https://trello.com/c/8BkCq4GO/3336-m-support-interruption-screen)

TobyRet:
[[M] Support interruption screen](https://trello.com/c/8BkCq4GO/3336-m-support-interruption-screen)

app/controllers/users_controller.rb
```diff
@@ -18,8 +25,12 @@ def accept_terms
 
 private
 
+  def user
+    @user ||= User.find(current_user["user_id"]).first
+  end
+
   def accept_screen(method, path)
-    User.member(current_user["user_id"]).send(method)
+    user.send(method)
```
misaka, line 33: Not a blocker for this PR, but it'd be a good habit to start to change these to `__send__` as we encounter them.

```suggestion
    user.__send__(method)
```

app/controllers/sessions_controller.rb
```diff
@@ -108,15 +108,32 @@ def auth_hash
...
-      redirect_to rollover_path
+    elsif permitted_state
+      redirect_to redirect_path[permitted_state]
     else
       redirect_to session[:redirect_back_to] || root_path
     end
   end
+
+  def permitted_state
+    user = User.new(
```
TobyRet, line 122: we need to initialise and instance of User in order to access `aasm`'s helper methods

app/controllers/users_controller.rb
TobyRet: 👍

app/controllers/sessions_controller.rb
misaka: I think we have the user object already here, from line 27 in the `create` method:

```ruby
    user = set_user_session
```

When then gets passed to `redirect_to_correct_page` as `user_session`, so we shouldn't have to reify a user here I don't think. Maybe line 27 above should assign `@user`? We also do it on line 39 in `create_by_magic` so that may need to be `@user` too.

Seems like some tech debt from when we first implemented the sessions, can we have a look at how we might tidy this up a little bit?

app/controllers/sessions_controller.rb
```diff
@@ -108,15 +108,32 @@ def auth_hash
...
-    if user&.accept_terms_date_utc.nil?
+  def redirect_to_correct_page(user_session)
+    if user_session&.accept_terms_date_utc.nil?
       redirect_to accept_terms_path
-    elsif user&.state == "new"
-      redirect_to transition_info_path
-    elsif Settings.rollover && user&.state == "transitioned"
-      redirect_to rollover_path
+    elsif next_state
+      redirect_to redirect_path[next_state]
```
misaka, line 115: Perfect. Much tidier!

app/controllers/users_controller.rb
```diff
@@ -18,8 +25,12 @@ def accept_terms
 
 private
 
+  def user
+    @user ||= User.find(current_user["user_id"]).first
```
misaka, line 29: Similar to my comment about reifying a user in `SessionsController#next_state`, do we really need to do this here? At first glance it looks like we should already have the user object from when we created the object.

spec/models/user_spec.rb
```diff
@@ -0,0 +1,72 @@
...
+
+describe User, type: :model do
+  let(:new_user) { build(:user, :new) }
+  let(:transitioned_user) { build(:user, :transitioned) }
+  let(:rolled_over_user) { build(:user, :rolled_over, associated_with_accredited_body: true) }
+  let(:seen_accredited_body_features_user) { build(:user, :seen_accredited_body_new_features, associated_with_accredited_body: true) }
+
+  describe "transition state event" do
+    it "changes state from 'new' to 'transitioned'" do
+      stub_request(:post, "#{Settings.manage_backend.base_url}/api/v2/users").with(body: /"state":"transitioned"/)
```
misaka, line 11: I think you're testing a bit more of the stack than you need to here, tbh. `aasm` comes with a rich set of [rspec matchers](https://github.com/aasm/aasm#rspec) which would avoid testing that we actually make a request.

Now that it's done this way we can leave it, but something to keep in mind for the future.

app/controllers/notifications_controller.rb
```diff
@@ -17,15 +17,25 @@ def update
       return
     end
 
+    if user.state != User::State::NOTIFICATIONS_CONFIGURED
```
misaka, line 20:
```suggestion
    if !user.notifications_configured?
```

app/models/user.rb
```diff
@@ -1,13 +1,63 @@
...
   custom_endpoint :accept_terms, on: :member, request_method: :patch
 
-  def self.member(id)
-    new(id: id)
+  module State
+    NEW = "new".freeze
+    TRANSITIONED = "transitioned".freeze
+    ROLLED_OVER = "rolled_over".freeze
+    NOTIFICATIONS_CONFIGURED = "notifications_configured".freeze
+  end
```
misaka, line 14: The only place these are used is to check the state the user is in, which is better handled by  [`aasm`'s built-in predicates](https://github.com/aasm/aasm#usage) anyway, and in specs. Do we really need it?

app/controllers/notifications_controller.rb
TobyRet: 👍

app/models/user.rb
TobyRet: 👍

TobyRet: Yep - they are only being used for the purposes of testing. I'll remove

app/services/update_user_service.rb
```diff
@@ -0,0 +1,23 @@
...
+    @user.__send__(@method)
+  rescue JsonApiClient::Errors::ClientError, JsonApiClient::Errors::ServerError => e
+    e.env.body.delete("traces")
+    Raven.extra_context(e.env)
+
+    if e.is_a? JsonApiClient::Errors::ClientError
+      Raven.capture_exception(e)
+    else
+      raise e
+    end
```
misaka, line 21: Good call moving this out of the controller, although it feels like we should move this somewhere so we can avoid even needing a specialized service ... maybe hook this into `JsonApiClient` somehow. Something to keep in mind for later.

app/models/user.rb
```diff
@@ -1,13 +1,56 @@
...
+
+  def next_state
+    aasm
+        .states(permitted: true)
+        .map(&:name)
+        .first
+  end
+
+  def aasm_read_state(_name = :default)
+    state.try(:to_sym).presence || self.class.aasm.initial_state
```
misaka, line 45: Feels like without the context of the discussion we had this morning, it could be a bit obscure why we have to do this bit. Can we add a comment here just to explain why we have to accommodate `nil` here?

app/models/user.rb
```diff
@@ -1,13 +1,56 @@
...
+  end
+
+  def aasm_write_state(new_state, _name = :default)
+    update(state: new_state.to_s)
+  end
+
+  def aasm_ensure_initial_state
+    self.aasm_state = :new
```
misaka, line 53: Should states be strings?

spec/controllers/notifications_controller_spec.rb
```diff
@@ -50,5 +53,27 @@
         expect(response).to redirect_to(provider_path(provider_code))
       end
     end
+
+    context "user is not subscribed to notifications" do
+      let(:user) { build(:user, :rolled_over, id: 1) }
+
+      it "calls the user update service" do
```
misaka, line 60: Should we have a spec, in the context where a user does have notifications configured, that the user's state is not updated? You could just move the stub on line 24 into this context and not add a spec, but it might be nice to be explicit about it too.

spec/features/sign_in_spec.rb
```diff
@@ -45,63 +45,142 @@
...
-    rollover_page.continue.click
-
-    expect(root_page).to be_displayed
-    expect(request).to have_been_made
+  describe "Interruption screens" do
+    before do
+      stub_omniauth(user: user)
+    end
+
+    context "transition screen" do
```
misaka, line 53:
```suggestion
    describe "transition screen" do
```

spec/features/sign_in_spec.rb
```diff
@@ -45,63 +45,142 @@
...
+
+          transition_info_page.continue.click
+
+          expect(root_page).to be_displayed
+          expect(user_update_request).to have_been_made
+        end
+      end
+    end
+
+    context "rollover screen" do
```
misaka, line 118:
```suggestion
    describe "rollover screen" do
```

spec/features/sign_in_spec.rb
```diff
@@ -45,63 +45,142 @@
...
+
+        expect(rollover_page.title).to have_content("Begin preparing for the next cycle")
+        rollover_page.continue.click
+
+        expect(root_page).to be_displayed
+        expect(user_update_request).to have_been_made
+      end
+    end
+
+    context "notifications screen" do
```
misaka, line 148:
```suggestion
    describe "notifications screen" do
```

spec/features/sign_in_spec.rb
misaka: Actually, if we want this to be a context, we could say:

```suggestion
    context "new user" do
```

app/models/user.rb
misaka: Discussed elsewhere, and agreed that they should be symbols.

app/services/update_user_service.rb
TobyRet: Yeah good idea

spec/controllers/notifications_controller_spec.rb
TobyRet: Yep good point. I've added

PR #1180, opened 2020-06-12
[3509] Add error message when no option selected

SamJamCul:
### Context
https://trello.com/c/lO5L7y4F/3509-s-editing-initial-allocation-causes-500-error-if-user-does-not-select-yes-or-no

Add error messaging to editing initial allocations page

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/features/providers/allocations/edit_initial_allocation_spec.rb
+++ b/spec/features/providers/allocations/edit_initial_allocation_spec.rb
@@ -19,8 +19,10 @@ RSpec.feature "PE allocations" do
       and_i_click_change
 
       then_i_see_do_you_want_page
+      and_i_click_continue
+      then_i_see_an_error_message
 
-      when_i_select_yes
+      when_i_select_yes_with_error
       and_i_click_continue
       then_i_see_edit_number_of_places_page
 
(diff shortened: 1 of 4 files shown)


SamJamCul:
[[S] Editing initial allocation causes 500 error if user does not select yes or no](https://trello.com/c/lO5L7y4F/3509-s-editing-initial-allocation-causes-500-error-if-user-does-not-select-yes-or-no)

spec/features/providers/allocations/edit_initial_allocation_spec.rb
```diff
@@ -19,8 +19,10 @@
       and_i_click_change
 
       then_i_see_do_you_want_page
+      and_i_click_continue
+      then_i_see_an_error_message
 
-      when_i_select_yes
+      when_i_select_yes_with_error
```
TobyRet, line 25: Is this renaming is necessary? I'm not clear how adding `_with_error` helps 🤔

SamJamCul: The id for the button element actually changes when the error message gets added to it - this means the prism page object no longer works for that button. I did try adding another element to the page object, but because they both use the same "Yes" value, it was getting all kinds of issues...

PR #1186, opened 2020-06-15
Remove duplicate warning message js

aliuk2012:
There is already a FormCheckLeave module which replicates the behaviour
which this code was using. I checked where this code might of been triggered
from and spotted that all the instances were being to setup to use the
FormCheckLeave module anyway.

https://github.com/DFE-Digital/publish-teacher-training/blob/1956e57d131c621a97b3b3eeec495b4662726547/app/webpacker/scripts/form-check-leave.js#L9-L27

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


aliuk2012:
@defong 👶 steps 😉.

As we add and improve our js tests it should help iron out these oddities.

PR #1187, opened 2020-06-15
[3500] Fix accredited body searches

asmega:
### Context

- https://trello.com/c/OO8AqZZe/3500-course-creation-accredited-body-search
- Previously was only searching a very small subset or providers

- Use modified api endpoint to search all accredited bodies for suggestions

### Guidance to review

- note sure about the code climate failure given that overall test coverage has not dropped
- depends on https://github.com/DFE-Digital/teacher-training-api/pull/1425
- create a new course as user `anonimized-user-8272@example.org`
- on the page to select accredited body perform a search
- should run search and return relevant accredited bodies
- should work for non-js version

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/courses/accredited_body_controller.rb
+++ b/app/controllers/courses/accredited_body_controller.rb
@@ -33,7 +33,7 @@ module Courses
       build_provider
       build_previous_course_creation_params
       @query = params[:query]
-      @provider_suggestions = ProviderSuggestion.suggest(@query)
+      @provider_suggestions = ProviderSuggestion.suggest_any_accredited_body(@query)
     rescue JsonApiClient::Errors::ClientError => e
       @errors = e
     end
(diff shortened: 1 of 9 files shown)


asmega:
[Course creation accredited body search](https://trello.com/c/OO8AqZZe/3500-course-creation-accredited-body-search)

spec/features/courses/accredited_body/new_spec.rb
```diff
@@ -133,11 +133,4 @@ def searching_returns_some_results
       ]),
     )
   end
-
-  def searching_returns_no_results
```
asmega, line 137: unused method

app/controllers/courses/accredited_body_controller.rb
```diff
@@ -33,7 +33,7 @@ def search_new
       build_provider
       build_previous_course_creation_params
       @query = params[:query]
-      @provider_suggestions = ProviderSuggestion.suggest(@query)
```
defong, line 36: `ProviderSuggestion.suggest`

that can be blasted from code base?

asmega: this it might be the case. so i'll make `suggest` do what `suggest_any` is doing and get rid of `suggest_any`

asmega: i feel that would be another PR though

PR #1190, opened 2020-06-16
[3544] Add copy to rollover provider show page.

aliuk2012:
# Context
We added "Courses as accredited body" and "Requests for PE" feature during the last year. We need to tell users that they can still see this feature for either the current or next cycle.

# Changes proposed in this pull request
- Moves rollover content into a new partial
- adds copy relating to being able to view "Courses as accredited body" which is only shown if the provider is an accredited body 
- adds copy relating to being able to view/make request for pe but only for the current cycle
- updates some other copy after a content review.

## Before
image

## After
image

# Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/views/providers/_show_during_rollover.html.erb
+++ b/app/views/providers/_show_during_rollover.html.erb
@@ -10,6 +10,7 @@
   <li>edit locations and vacancies in the current cycle</li>
   <% if @provider.accredited_body? %>
     <li>see which courses you're the accredited body for in the current cycle</li>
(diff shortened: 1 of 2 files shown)


aliuk2012:
[Update content on cycles landing page to include accredited body view](https://trello.com/c/Ij4UC5ci/3544-update-content-on-cycles-landing-page-to-include-accredited-body-view)

PR #1195, opened 2020-06-17
[3563] Use TrainingProvider#show endpoint

asmega:
### Context

- https://trello.com/c/6S5IQpl1/3563-accredited-bodies-cant-see-the-courses-they-accredit
- When viewing a training provider we are using the Provider#show endpoint which is only accessible to users that are associated with the provider via an organisation
- Switch to use a new endpoint that is accessible to users that are part of an accredited body on the provider

- Previously using Provider#show endpoint which was only accessible if
the user was associated to that provider via an organisation
- Given a training provider TrainingProvider#show endpoint is accessible
to users that are part of any accredited body of the training provider
thru the courses they offer

### Guidance to review

- This change depends on https://github.com/DFE-Digital/teacher-training-api/pull/1431
- Login as a user of an accredited body e.g `anonimized-user-8916@example.org`
- Should be able to view accredited bodies
- Should be able to view courses of an accredited body

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers/allocations_controller.rb
+++ b/app/controllers/providers/allocations_controller.rb
@@ -8,10 +8,11 @@ module Providers
     PE_SUBJECT_CODE = "C6".freeze
 
     def index
-      @training_providers = @provider.training_providers(
+      @training_providers = TrainingProvider.where(
         recruitment_cycle_year: @recruitment_cycle.year,
-        "filter[subjects]": PE_SUBJECT_CODE,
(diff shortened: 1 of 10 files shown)


asmega:
[Accredited bodies can't see the courses they accredit](https://trello.com/c/6S5IQpl1/3563-accredited-bodies-cant-see-the-courses-they-accredit)

PR #1198, opened 2020-06-18
[Spike] Implementing GOV.UK Components

aliuk2012:
### Context
Used to see how difficult it would be to adopt GOV.UK Components.

Components replaced
- Back links
- Details (2 outstanding, see commit message)
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/Gemfile
+++ b/Gemfile
@@ -16,6 +16,7 @@ gem "webpacker"
 gem "aasm"
 
 # Used to build our forms and style them using govuk-frontend class names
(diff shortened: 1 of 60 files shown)


aliuk2012:
Build is currently failing because we are installing GOV.UK Components straight from Github and the docker image does not have git installed by default. I have asked if the gem could be added to rubygems

PR #1202, opened 2020-06-19
[3576] Preview tweaks

asmega:
### Context

- https://trello.com/c/lD6tJVsR/3576-s-publish-preview-functionality-should-match-find
- There are differences between the Publish preview functionality and how it will actually look on find. This change hopes to remove these differences

- Links are now prefixed with `http` if there is no protocol specified. This is to prevent incorrect relative urls when they should be external links
- Smart quotes used when markdown is processed
- Markdown processor does some additional work to better render lists and bullet points to due incorrect markdown entered by users
-  Copy changes to the preview pages so it matches Find
- When a course is in `rolled_over` state and is being previewed we now show the "Apply section" as part of the preview

### Guidance to review

- Find a course in rolled over state and preview
- Compare with current version on Find
- These 2 pages should be near to identical

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/helpers/application_helper.rb
+++ b/app/helpers/application_helper.rb
@@ -13,6 +13,16 @@ module ApplicationHelper
     options = { autolink: true, lax_spacing: true }
     markdown = Redcarpet::Markdown.new(render, options)
     markdown.render(source).html_safe
+
+    # Convert quotes to smart quotes
+    source_with_smart_quotes = smart_quotes(source)
+    markdown.render(source_with_smart_quotes).html_safe
+  end
+
+  def smart_quotes(string)
+    return "" if string.blank?
+
+    RubyPants.new(string, 2, ruby_pants_options).to_html
   end
 
   def enrichment_error_link(model, field, error)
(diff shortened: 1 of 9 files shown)


asmega:
[[S] Publish preview functionality should match Find](https://trello.com/c/lD6tJVsR/3576-s-publish-preview-functionality-should-match-find)

app/helpers/application_helper.rb
```diff
@@ -13,6 +13,21 @@ def markdown(source)
...
+
+    # Fix common markdown errors:
+    # - using bullets rather than *
+    # - not putting a space between * and word
+    source = source.gsub(/•\s?/, "* ").gsub(/^\*(?![\s\*])/, "* ")
+
+    # Convert quotes to smart quotes
+    source_with_smart_quotes = smart_quotes(source)
+    markdown.render(source_with_smart_quotes).html_safe
+  end
```
fofr, line 25: I'd purposefully omitted this from Publish preview, as I'd prefer providers to put in the correct markdown, and then fix it if it's broken when checking on preview.

asmega: i see your point but at the same time don't users want to see the actual preview? I guess it's a double edged sword. I think i will leave smart quotes in and remove bullet point one

asmega: line 20 has now gone

PR #1209, opened 2020-06-23
[3336] - Bug fix: state transition error

TobyRet:
### Context
- When a user updated their notification preferences after dismissing the notifications interruption screen, a state transition was accidentally being called. 

- When a user updates their notifications preferences we now check if the user is 'allowed' to transition their state to 'notifications_configured'. If true then their state is transitioned and they will no longer see the notifications interruption screen

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/notifications_controller.rb
+++ b/app/controllers/notifications_controller.rb
@@ -17,7 +17,7 @@ class NotificationsController < ApplicationController
       return
     end
 
-    unless user.notifications_configured
+    if user.may_accept_notifications_screen?
       UpdateUserService.call(user, "accept_notifications_screen!")
     end
 
--- a/spec/controllers/notifications_controller_spec.rb
+++ b/spec/controllers/notifications_controller_spec.rb
@@ -16,7 +16,7 @@ describe NotificationsController, type: :controller do
   describe "UPDATE" do
     before do
       allow_any_instance_of(ApplicationController).to receive(:current_user)
-        .and_return(current_user)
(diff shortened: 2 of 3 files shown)


app/controllers/notifications_controller.rb
```diff
@@ -17,7 +17,7 @@ def update
       return
     end
 
-    unless user.notifications_configured
+    unless user.notifications_configured?
```
TobyRet, line 20: We are now checking the state of the user via the state machine

app/controllers/notifications_controller.rb
```diff
@@ -17,7 +17,7 @@ def update
       return
     end
 
-    unless user.notifications_configured
+    if user.may_accept_notifications_screen?
```
misaka, line 20: I like this naming better, but where is this defined?

spec/controllers/notifications_controller_spec.rb
```diff
@@ -16,7 +16,7 @@
   describe "UPDATE" do
     before do
       allow_any_instance_of(ApplicationController).to receive(:current_user)
-        .and_return(current_user)
+                                                        .and_return(current_user)
```
misaka, line 19: 😻

app/controllers/notifications_controller.rb
TobyRet: It's an aasm helper

PR #1210, opened 2020-06-23
[3589] - Skip signin redirect to interruption pages

misaka:
### Context
If users try to visit to an interruption screen that they have previously actioned, and accept the interruption screen again they receive an an error as a result of the recent addition of the state machine to publish. This is most likely because these users have bookmarked the interruption pages.

Before allowing a user to see an interruption screen we first check if their state is valid. If they already have the state transition that results from accepting the interruption screen then they are redirected to the next interruption screen and so on.

### Guidance for review
Set your user state to 'rolled_over' and try and visit the following interruption screens - `/transition-info`, `/rollover`. Rather than seeing flames you should be redirected. 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/features/sign_in_spec.rb
+++ b/spec/features/sign_in_spec.rb
@@ -69,6 +69,8 @@ feature "Sign in", type: :feature do
 
       context "Rollover is enabled" do
         let(:user) { build(:user, :new) }
+        let(:transitioned_user) { build(:user, :transitioned, id: user.id) }
+
         let(:user_update_request) do
           stub_request(
             :patch,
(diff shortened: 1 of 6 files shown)


misaka:
[Skip interruption pages when user has already accepted them.](https://trello.com/c/xGMnnXOq/3589-skip-interruption-pages-when-user-has-already-accepted-them)

spec/features/sign_in_spec.rb
```diff
@@ -69,14 +70,23 @@
...
           )
                                       .with(body: /"state":"transitioned"/)
         end
-        let(:user_get_request) { stub_api_v2_request("/users/#{user.id}", user.to_jsonapi) }
+
+        let(:user_get_request) do
+          stub_request(:get, "#{Settings.manage_backend.base_url}/api/v2/users/#{user.id}").to_return(
+            { body: user.to_jsonapi.to_json, headers: { 'Content-Type': "application/vnd.api+json" } },
+            { body: user.to_jsonapi.to_json, headers: { 'Content-Type': "application/vnd.api+json" } },
+            { body: transitioned_user.to_jsonapi.to_json, headers: { 'Content-Type': "application/vnd.api+json" } },
```
TobyRet, line 87: This spec is transitioning the state of the user, hence why a modified user (i.e. `transitioned_user`) is being returned

PR #1221, opened 2020-06-25
[3602] - Fix accept terms process

TobyRet:
### Context
Currently users are unable to accept-terms (unauthorised)

- This commit reverts back to the original behaviour for allowing users to accept terms. This is because the api users update endpoint contains a guard to check if the user has accepted terms.
- Additional fix added - access request count is displayed in footer only if user has accepted terms. I noticed when manually testing the fix for accept terms that if you are an admin user `admin: true` when you reach the accept terms page, the footer makes a direct API call that results in a 403 loop. See https://github.com/DFE-Digital/publish-teacher-training/blob/2fe9b92fbb7ade233ec563280de32fab04edebb2/app/views/layouts/_footer.html.erb#L9

An additional guard has been added to the footer so that it does not blow up if the user is admin and has not accepted terms.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -184,6 +184,7 @@ private
     session[:auth_user]["state"] = user.state
     session[:auth_user]["associated_with_accredited_body"] = user.associated_with_accredited_body
     session[:auth_user]["notifications_configured"] = user.notifications_configured
+    session[:auth_user]["accepted_terms?"] = user.accept_terms_date_utc.present?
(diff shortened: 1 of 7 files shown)


TobyRet:
https://trello.com/c/CfkBwzLC/3602-bug-accept-terms

asmega:
[Bug: Accept terms](https://trello.com/c/CfkBwzLC/3602-bug-accept-terms)

PR #1224, opened 2020-06-25
[3543] Accept rollover for 2021

asmega:
### Context

- https://trello.com/c/AMMShiva/3543-m-build-final-interruption-screen-for-roll-over
- When the user accepts rollover for 2021 we need to change state to keep track
- This adds a new state so we know user has accepted for the 2021 cycle

- When a user now acknowledges rollover they are set the state of `accepted_rollover_2021`

### Guidance to review

- depends on https://github.com/DFE-Digital/publish-teacher-training/pull/1221
- this can be easily tested if the following is cherry picked https://github.com/DFE-Digital/publish-teacher-training/pull/1216
- navigate to /auth/developer
- login as different users some examples: admins, new, rolled_over, transitioned, those who have not accepted the terms
- on login should see correct screens
- logout will also work with the cherry picked branch
- login back in as the user
- should not see interrupt screens and should be sent to correct home page

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/features/sign_in_spec.rb
+++ b/spec/features/sign_in_spec.rb
@@ -43,7 +43,7 @@ feature "Sign in", type: :feature do
 
     # Redirect to DfE Signin and come back
     expect(page).to have_content("Sign out (#{user.first_name} #{user.last_name})")
-    expect(root_page).to be_displayed
+    expect(page.current_path).to eql("/rollover")
   end
 
   describe "Interruption screens" do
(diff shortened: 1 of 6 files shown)


asmega:
[[M] Build final interruption screen for roll-over](https://trello.com/c/AMMShiva/3543-m-build-final-interruption-screen-for-roll-over)

spec/features/sign_in_spec.rb
```diff
@@ -43,7 +43,7 @@
 
     # Redirect to DfE Signin and come back
     expect(page).to have_content("Sign out (#{user.first_name} #{user.last_name})")
-    expect(root_page).to be_displayed
+    expect(page.current_path).to eql("/rollover")
```
TobyRet, line 46: I'm not sure I understand this change. The spec has a user with the default state ('rolled_over'). Given that the user has this state they should be directed to root_path not "/roll_over"?

TobyRet, line 46: At first I was confused how this was passing then saw that the user's next state will redirect the user to roll_over path. 👍 
`accepted_rollover_2021: rollover_path`

PR #1229, opened 2020-06-29
[3615] Age range now mandatory when updating

asmega:
### Context

- https://trello.com/c/4OdjeI1q/3615-make-age-range-a-required-field-for-courses-in-2020-cycle
- We want to improve data collected for courses
- Therefore age range is now mandatory when updating the age range

- This is done with client side validation
- Cannot be easily done as an API side validation with the current
setup. Otherwise when updating another field this validation would be
triggered
- The codeclimate issue seems fine unless you can offer a better solution

### Guidance to review

- Find a course that has `age_range_in_years` set to `nil`
- Navigate to the edit the age range page for the course
- Attempt to save without entering an age range
- Should see validation message

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/courses/age_range_controller.rb
+++ b/app/controllers/courses/age_range_controller.rb
@@ -45,6 +45,12 @@ module Courses
     end
 
     def build_errors
+      if no_selection?
+        return {
+          age_range_in_years: [t("age_range.errors.missing_error")],
+        }
+      end
+
       if age_range_from_and_to_missing?
         return {
           age_range_in_years: [t("age_range.errors.from_and_to_error")],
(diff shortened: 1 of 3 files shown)


asmega:
[Make age-range a required field for courses in 2020 cycle](https://trello.com/c/4OdjeI1q/3615-make-age-range-a-required-field-for-courses-in-2020-cycle)

asmega:
[Make age-range a required field for courses in 2020 cycle](https://trello.com/c/4OdjeI1q/3615-make-age-range-a-required-field-for-courses-in-2020-cycle)

PR #1232, opened 2020-06-29
[3594] Add setting to disable DfE Signin

gpeng:
### Context

We want to be able to disable signin by DfE signin if there is a problem with the service.

Adds a feature flag to disable signin via DfE signin.

When the feature is disabled an attempt to login via DfE signin renders the magic link signin page (`sessions#new`) or redirects to it if the user has actually come from signin already.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/sessions_controller.rb
+++ b/app/controllers/sessions_controller.rb
@@ -2,7 +2,7 @@ class SessionsController < ApplicationController
   skip_before_action :request_login
 
   def new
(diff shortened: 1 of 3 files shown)


gpeng:
[Add setting to disable sign-in via DSI](https://trello.com/c/xOiiAlEx/3594-add-setting-to-disable-sign-in-via-dsi)

PR #1235, opened 2020-06-30
[3551] -  Refactor edit initial allocation flow

TobyRet:
### Context
Currently, the logic for creating an initial allocation is very different to updating an initial allocation. This was because of the deadline to get allocations work finished. 

- Move logic out of the controller into a flow object. This now follows a similar pattern to creating an initial allocation.
- Create separate form objects for each stage of the edit initial allocation flow.
- This change also reduces the number of routes required.

### Guidance to review
- The behaviour remains unchanged. 🤞 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review


TobyRet:
[[L] Initial Allocations Refactor](https://trello.com/c/NbW1g63O/3551-l-initial-allocations-refactor)

app/form_objects/edit_initial_request_form.rb
```diff
@@ -0,0 +1,28 @@
+class EditInitialRequestForm
```
TobyRet, line 1: Both this form and the [InitialRequest form](https://github.com/DFE-Digital/publish-teacher-training/blob/master/app/form_objects/initial_request_form.rb) have the `number of places` validation in common - potentially I could extract the common behaviour to a base class an inherit but then it may not be obvious what the base class is doing. Any opinion?

app/flows/edit_initial_request_flow.rb
```diff
@@ -0,0 +1,120 @@
...
+  def proceed_to_number_of_places_page?
+    params[:next_step] == "number_of_places" && params[:request_type].present? || number_of_places_validation_error?
+  end
+
+  def proceed_to_check_answers_page?
+    params[:next_step] == "check_answers" && params[:number_of_places].present? && form_object.valid?
+  end
+
+  def number_of_places_page?
+    (params[:step].present? && params[:step] == "number_of_places") || params[:number_of_places].present?
```
aliuk2012, line 80: This method isn't being tested

app/flows/edit_initial_request_flow.rb
```diff
@@ -0,0 +1,120 @@
...
+  end
+
+  def check_answers_page?
+    params[:step].present? && params[:step] == "check_answers"
```
aliuk2012, line 84: This method isn't being tested

app/form_objects/edit_initial_request_form.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def number_of_places_valid?
+    !number_of_places.empty? &&
+      /\A\d+\z/.match?(number_of_places) &&
+      number_of_places.to_i.positive?
+  end
+
+  def selected_request_type
+    errors.add(:request_type, "Select one option") unless number_of_places.present? || request_type_valid?
```
aliuk2012, line 22: This method isn't being tested according to code climate

app/form_objects/edit_initial_request_form.rb
```diff
@@ -0,0 +1,28 @@
...
+  end
+
+  def request_type_valid?
+    request_type == AllocationsView::RequestType::INITIAL || AllocationsView::RequestType::DECLINED
```
aliuk2012, line 26: This method isn't being tested according to code climate

app/form_objects/edit_initial_request_form.rb
SamJamCul: @asmega @aliuk2012 and I have been looking at this - it would be better if we could have separate form objects for each of these pages - rather than overloading one form object with lots of different validations. We can also use some inbuilt rails validation to more closely match what happens on the API side

`validates :number_of_places, presence: { message: "You must enter a number" },
                               numericality: { only_integer: true, greater_than: 0, message: "You must enter a number greater than 0" }`

SamJamCul:
It would be good if we could rename some of the template names being used, so that they better match the actions. In particular the `do_you_want.html.erb` doesn't line up with much anymore.

app/form_objects/edit_initial_request_form.rb
TobyRet: @aliuk2012 @asmega @SamJamCul - I think using separate form objects is potentially an idea worth considering - there may be some factors to consider though. At the code review we opted for a single for a single form object for all pages to more close mirror the 'create' initial allocation process. If we go for a multiple form object pattern (i.e. one form object per page) we will need to refactor the create initial allocation process too, so will require some more work and will have to make a case for prioritising. Also are suggesting that the multiple form object pattern become the preferred approach moving forwards? Could be worth getting input from the rest of the team.

app/views/providers/edit_initial_allocations/do_you_want.html.erb
```diff
@@ -7,14 +7,14 @@
...
+    <%= form_with url: post_edit_initial_request_provider_recruitment_cycle_allocation_path(
                          provider.provider_code,
                          provider.recruitment_cycle_year,
                          training_provider.provider_code,
                   ),
                   method: :post,
-                  model: allocation_model,
-                  scope: "allocation",
+                  model: form_object,
+                  scope: "",
```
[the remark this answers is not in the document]
TobyRet: By setting to `nil` or removing `scope` entirely Rails nests the params in another hash, which is a bit of a pain TBH. By setting to empty string we retain a flat structure

When scope set to `nil`

```
=> <ActionController::Parameters {"utf8"=>"✓", "authenticity_token"=>"pAq6PvaoTMmASO0epcTPkKnQ5JRnDgAJJAG0odVMwOd44e9AZ1w1l4xDZKobHNZmiN6dvT/b6Or0BYoMP2ElNA==", "allocations_edit_initial_request_type_form"=>{"request_type"=>"initial"}, "id"=>"903", "next_step"=>"number_of_places", "commit"=>"Continue", "on"=>:member, "controller"=>"providers/edit_initial_allocations", "action"=>"edit", "provider_code"=>"L51", "recruitment_cycle_year"=>"2020", "training_provider_code"=>"2CD"} permitted: false>
```

app/views/providers/edit_initial_allocations/number_of_places.html.erb
```diff
@@ -14,14 +14,15 @@
...
+  <%= form_with url: post_edit_initial_request_provider_recruitment_cycle_allocation_path(
                       provider.provider_code,
                       recruitment_cycle.year,
                       training_provider.provider_code,
                       id: allocation.id
                     ),
                     skip_enforcing_utf8: true,
-                    model: allocation,
+                    model: form_object,
+                    scope: "",
```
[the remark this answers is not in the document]
TobyRet: https://github.com/DFE-Digital/publish-teacher-training/pull/1235#discussion_r455893694

PR #1236, opened 2020-07-01
[3623] No default course length

asmega:
### Context

- https://trello.com/c/lMuMuYoE/3623-m-when-creating-a-course-and-completing-course-fees-section-we-should-not-set-a-default-for-the-course-length
- A default course length is being set for the user which means less that accurate is being collected about the course

- On first setting of course length we no longer set the default value
to other
- Instead the user can choose what value to select from the choice of
radios
- Before publication the user is forced to enter a value so the user
must enter the data
- This aims to collect more accurate data about the course

### Guidance to review

- https://s121d02-3623-ptt-as.azurewebsites.net/
- Create a course
- View/Edit the course in publish
- Under the `Description` tab, click `Course length and fees`
- Hit `Save` button
- Navigate back to this page
- The course length should not be `Other`
- Attempt to publish this course
- One of the reasons that prevent you from publishing should be that course length is missing and needs populating 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/decorators/course_decorator.rb
+++ b/app/decorators/course_decorator.rb
@@ -139,7 +139,7 @@ class CourseDecorator < ApplicationDecorator
   end
 
   def other_course_length?
-    %w[OneYear TwoYears].exclude?(course.course_length)
+    %w[OneYear TwoYears].exclude?(course.course_length) && !course.course_length.nil?
   end
 
   def other_age_range?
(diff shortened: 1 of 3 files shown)


asmega:
[[M] When creating a course and completing course fees section we should not set a default for the course length](https://trello.com/c/lMuMuYoE/3623-m-when-creating-a-course-and-completing-course-fees-section-we-should-not-set-a-default-for-the-course-length)

PR #1237, opened 2020-07-01
[3573] Admins panel to update provider name

asmega:
### Context

- https://trello.com/c/KXiJdzYT/3573-m-make-provider-name-editable
- Admin users of publish want to be able to update provider names for convenience

- In the contact details page admins can see the `provider_name` field
- Non-admins cannot see this field
- The API determines if the user is allowed update this field or not

### Guidance to review

- https://s121d02-3573-ptt-as.azurewebsites.net/
- This change depends on https://github.com/DFE-Digital/teacher-training-api/pull/1461
- Login in a non-admin
- Navigate to contact page of a provider
- Should not see admin panel with `provider_name` field
- Logout
- Login as admin
- Navigate to contact page of a provider
- Should see an admin panel with the `provider_name` field
- Change the `provider_name` to something else
- Submit form
- Navigate to provider
- Provider name should now be updated to new value

### Screenshots

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/views/providers/contact.html.erb_spec.rb
+++ b/spec/views/providers/contact.html.erb_spec.rb
@@ -1,0 +1,45 @@
+require "rails_helper"
+
+RSpec.describe "providers/contact.html.erb" do
+  before do
+    assign(:provider, build(:provider))
+  end
+
+  context "when not an admin" do
+    before do
+      controller.singleton_class.class_eval do
+      protected
+
+        def current_user
+          { "admin" => false }
+        end
+        helper_method :current_user
+      end
+    end
+
+    it "cannot see provider_name field" do
+      render
+
+      expect(rendered).to_not include("Provider name")
+    end
+  end
+
+  context "when an admin" do
+    before do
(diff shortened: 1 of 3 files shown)


asmega:
[[M] Make provider name editable](https://trello.com/c/KXiJdzYT/3573-m-make-provider-name-editable)

spec/views/providers/contact.html.erb_spec.rb
```diff
@@ -0,0 +1,45 @@
+require "rails_helper"
+
+RSpec.describe "providers/contact.html.erb" do
+  before do
+    assign(:provider, build(:provider))
+  end
+
+  context "when not an admin" do
+    before do
```
gpeng, line 9: We should make a helper for this. We've got a few different ways of doing it now and it's in a lot of controller and view specs. I'll put a card on my ideas board though, let's not worry about it in this PR.

PR #1243, opened 2020-07-06
[3595] Set magic link environment variables

gpeng:
### Context

We have implemented a magic link signin alternative but it isn't correctly configured.

Set the necessary environment variables so as they are available in Azure.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/azure/template.json
+++ b/azure/template.json
@@ -364,6 +364,18 @@
               {
                 "name": "SETTINGS__ROLLOVER",
                 "value": "[parameters('rollover')]"
+              },
+              {
+                "name": "SETTINGS__FEATURES__SIGNIN_INTERCEPT",
+                "value": "[parameters('settingsFeaturesSigninIntercept')]"
+              },
+              {
+                "name": "SETTINGS__FEATURES__SIGNIN_BY_EMAIL",
+                "value": "[parameters('settingsFeaturesSigninByEmail')]"
+              },
+              {
+                "name": "SETTINGS__FEATURES__DFE_SIGNIN",
+                "value": "[parameters('settingsFeaturesDfeSignin')]"
               }
             ]
           },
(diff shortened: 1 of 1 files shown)


gpeng:
[[M] Test magic links work.](https://trello.com/c/SKHQuaVQ/3595-m-test-magic-links-work)

azure/template.json
```diff
@@ -179,6 +179,13 @@
         "description": "true to offer the user the option to sign-in using a magic link sent to them by email"
       }
     },
+    "settingsFeaturesDfeSignin": {
+      "type": "string",
+      "defaultValue": "true",
+      "metadata": {
+        "description": "false to disable login via DfE Signin. Users are redirected to the magic link singin page"
```
misaka, line 186: Is this saying that by setting this to `false`, users will be redirected to the magic link signin page? There appears to also be a `settingsFeaturesSigninByEmail` parameter below, so does that one not control whether the user will use magic links?

Sorry, I know this is nit-picking, but I guess it's just highlighting something we've already noticed, that it's confusing to have multiple switches for this.

gpeng: Yes. It is a bit confusing. The ticket said that disabling signin should work whether magic links are enabled or not which is basically what this does. It redirects to the alternative login page when you try to login via signin either from signin or from publish via signin.

We did discuss what should happen if this is `false` and the user attempts to login. I originally had it as a 403 but we decided it wasn't very helpful for the user and that it should redirect to the publish signin page.

gpeng: That behaviour is already in master too, this just makes the environment variable available to be changed in Azure (I think 😬). If we're going to change the behaviour we should open another ticket I think.

PR #1250, opened 2020-07-09
[3660] After accepting terms sync session data

asmega:
### Context

- https://trello.com/c/sDVFcGD0/3660-update-session-when-user-accepts-terms
- After the user accepts term we keep the session in sync with this new data
- This is to make https://trello.com/c/IQbTVjZV/3638-m-all-interruptions-are-not-being-displayed-when-the-user-next-signs-in easier
- As we will now have a source of truth on whether or not the user has accepted terms or not

- After the user accepts term we keep the session in sync with this new data
- This way we know when to display the terms page or not
- API also performs this check
- But now publish knows when to display the terms page
- Out of scope: there seem to other bits of hairy code working around the terms. We can change those to use this mechanism once merged 

### Guidance to review

- Depends on https://github.com/DFE-Digital/teacher-training-api/pull/1469
- No behavioural change
- Login as a new user who has not accepted terms
- Accept terms
- Should now be able to use publish like normal 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/users_controller.rb
+++ b/app/controllers/users_controller.rb
@@ -16,7 +16,8 @@ class UsersController < ApplicationController
 
   def accept_terms
     if params.require(:user)[:terms_accepted] == "1"
-      User.member(current_user["user_id"]).accept_terms
+      result = User.member(current_user["user_id"]).accept_terms
+      session["auth_user"]["attributes"]["accept_terms_date_utc"] = result.first.accept_terms_date_utc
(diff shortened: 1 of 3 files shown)


asmega:
[update session when user accepts terms](https://trello.com/c/sDVFcGD0/3660-update-session-when-user-accepts-terms)

PR #1252, opened 2020-07-10
3638 Rework interrupt screens

SamJamCul:
### Context

- https://trello.com/c/IQbTVjZV/3638-m-all-interruptions-are-not-being-displayed-when-the-user-next-signs-in
- Fixes issue where different interrupt screen were being show on login
- Fixes hitting the browser back button then continue on interrupt screens would throw an error 

* Removed notifications interrupt screen
* After login we redirect to root path
* Added before action to application controller to then redirect user to correct page depending on their current state
* State machine no longer raises errors for invalid state changes
* State changes are persisted to session

### Guidance to review

- Login as new user
- Should see terms and conditions
- Accept terms and conditions
- Should see transition interrupt screen
- Continue
- Should see both rollover interrupt screens
- Should proceed to home page
- At any point should be able to set notifications unless you have not accepted terms
- There is a small issue that if you configure notifications you will see rollover screens again. This is to ensure user will rollover screens if they have gone in to the wrong state. Also as we are use a single state machine / flag we can't currently determine what screen have been seen

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review


asmega:
[[M] All Interruptions are not being displayed when the user next signs in.](https://trello.com/c/IQbTVjZV/3638-m-all-interruptions-are-not-being-displayed-when-the-user-next-signs-in)

app/controllers/errors_controller.rb
```diff
@@ -1,6 +1,5 @@
-class ErrorsController < ApplicationController
-  skip_before_action :verify_authenticity_token,
-                     :request_login
+class ErrorsController < ActionController::Base
```
TobyRet, line 1: Feels like this particular change to the ErrorsController deserves a separate commit?

app/models/user.rb
```diff
@@ -10,7 +10,7 @@ def self.member(id)
     new(id: id)
   end
 
-  aasm do
+  aasm whiny_transitions: false do
```
TobyRet, line 13: whiny? 😸

spec/rails_helper.rb
```diff
@@ -49,3 +49,7 @@
   # arbitrary gems may also be filtered via:
   # config.filter_gems_from_backtrace("gem name")
 end
+
+Capybara.register_driver :rack_test do |app|
```
TobyRet, line 53: Separate commit?

spec/models/user_spec.rb
```diff
@@ -102,7 +102,7 @@
       let(:rolled_over_user) { build(:user, :rolled_over, associated_with_accredited_body: false, notifications_configured: true) }
 
       it "raises an error when trying to change state from 'rolled_over' to 'notifications_configured'" do
-        expect { rolled_over_user.accept_notifications_screen! }.to raise_error(AASM::InvalidTransition)
+        expect { rolled_over_user.accept_notifications_screen! }.to_not raise_error
```
TobyRet, line 105: Asserting an error is not raised - test description needs to be updated

spec/models/user_spec.rb
```diff
@@ -111,7 +111,7 @@
       let(:rolled_over_user) { build(:user, :notifications_configured, associated_with_accredited_body: true, notifications_configured: true) }
 
       it "raises an error when trying to change state from 'rolled_over' to 'subscribed_to_notifications'" do
-        expect { rolled_over_user.accept_notifications_screen! }.to raise_error(AASM::InvalidTransition)
+        expect { rolled_over_user.accept_notifications_screen! }.to_not raise_error
```
TobyRet, line 114: Same here 
Asserting an error is not raised - test description needs to be updated

spec/rails_helper.rb
SamJamCul: Could do, it is mainly connected to these changes though?

app/models/user.rb
SamJamCul: Haha, it helps not throw up too much when an invalid state change is made - like when they visit a bookmarked page that showed up when they were on a different state.

spec/models/user_spec.rb
SamJamCul: Good find, changed it!

spec/models/user_spec.rb
SamJamCul: Done!

app/controllers/users_controller.rb
```diff
@@ -1,24 +1,29 @@
 class UsersController < ApplicationController
+  skip_before_action :check_interrupt_redirects
+
   def accept_transition_info
-    UpdateUserService.call(user, "accept_transition_screen!")
-    redirect_to Settings.rollover ? rollover_path : providers_path
+    user.accept_transition_screen!
```
TobyRet, line 5: Was there any particular reason why the UpdateUserService is no longer being used?

app/models/user.rb
TobyRet: Ah interesting - this stops the `invalid transition` error from being raised. Wished I had seen that earlier

spec/rails_helper.rb
TobyRet: This change is changing the way Capybara works so I'm guessing is general config change and not specific to the bug fix? Unless I'm mistaken.

app/controllers/users_controller.rb
SamJamCul: I might need @asmega to explain it well, but we found that using these `user.accept_transition_screen!` methods on the user was already updating the database object with the new state, thanks to the `!` on the end. It made the service call redundant. 

Basically the way the state machine works in the user model means changing the state does everything already

spec/rails_helper.rb
SamJamCul: Done, split it out!

app/controllers/errors_controller.rb
SamJamCul: I'm not familiar with why this change was included, I think @asmega will know more than me here

app/models/user.rb
```diff
@@ -61,4 +60,8 @@ def aasm_ensure_initial_state
   def self.generate_and_send_magic_link(email)
     MagicLink::GenerateAndSendService.call(email: email, site: site)
   end
+
+  def accepted_terms?
```
defong, line 64: Is there test? Should this be `not_accepted_terms?` instead?

app/controllers/users_controller.rb
asmega: Yes, I don't see why it was ever created it the first place or the purpose of it's existence. The class literally takes an object and method and sends that method to the object. Why don't we just invoke that method on the original object? It would literally take one line. Why do we need an entirely new class to invoke a method? For me it's smoke and mirrors and un-needed complexity for the sake of complexity. I wish to remove the `UpdateUserService` class as soon as possible.

app/models/user.rb
asmega: Good spot, a direct test appears to be missing, we will add. As for naming I think not including the negation helps with comprehension. Alternatively we can also implement `not_accepted_terms?` for convenience

app/controllers/errors_controller.rb
asmega: We can change back to inheriting from `ApplicationController` for the sake of this PR we will just need to add the additional skip filters

SamJamCul: It is a bit of a pain to switch it back, the reason it was changed seems to be that all of the default Application Controller behaviour isn't needed if you're showing an error page - it skips all of it anyway, and adding in this as an extra skip doesn't make it any more useful. I do think it could be a separate commit though, seeing as it's got quite a broad effect?

app/controllers/users_controller.rb
TobyRet: The UpdateUserService was extracted out as it the logic is required in more than one place. The service is doing more than simply calling the API, and has error handling. Misha and I have discussed and agreed it would be a good idea to ultimately remove the UpdateUserService and add the error handling to the Base class so that all api calls can benefit from it. I'll add a ticket to the ideas board, in the meantime might as well keep using it? Note that the NotificationsController is also calling this class

TobyRet: https://github.com/DFE-Digital/publish-teacher-training/pull/1178/files#r442766309

TobyRet: discussion

.rubocop.yml
```diff
@@ -29,3 +29,5 @@ Style/HashTransformValues:
   Enabled: false
 Rails/FindBy:
   Enabled: false
+Rails/ApplicationController:
```
misaka, line 32: Can we skip this check only in `ErrorsController` and not globally?

app/controllers/users_controller.rb
asmega: In which case yes this belongs in the client otherwise you'll have to implement this on every API call

asmega: But, why was this added in the first place? There must have been a specific error or case that is trying to be handled?

.rubocop.yml
asmega: I've now removed this as I've now made the `ErrorsController` inherit from `ApplicationController`

app/models/user.rb
asmega: Test has now been added

app/controllers/errors_controller.rb
asmega: This now inherits from `ApplicationController`

app/models/user.rb
```diff
@@ -22,15 +22,14 @@ def self.member(id)
...
     event :accept_rollover_screen do
-      transitions from: %i[transitioned rolled_over], to: :accepted_rollover_2021 do
+      transitions from: %i[notifications_configured transitioned rolled_over], to: :accepted_rollover_2021 do
         guard { FeatureService.enabled?("rollover.can_edit_current_and_next_cycles") }
       end
     end
 
     event :accept_notifications_screen do
-      transitions from: %i[rolled_over accepted_rollover_2021], to: :notifications_configured do
+      transitions from: %i[transitioned rolled_over accepted_rollover_2021], to: :notifications_configured do
```
misaka, line 31: Why is this event still here if we're getting rid of the notifications screen? This event is [fired from the `NotificationsController`](https://github.com/DFE-Digital/publish-teacher-training/pull/1252/files#diff-e7d2878d7ccecd38f6e3a83b5cf85053R23) but I don't see why we still want to do that. This state is strictly about tracking what interrupt pages they're viewed, not tracking whether they've configured notifications, so let's strip the notifications stuff all out.

SamJamCul: Good point! I'll get the changes working

app/controllers/users_controller.rb
TobyRet: Yeah not entirely sure regarding the historical context. I can discuss with Misha

PR #1255, opened 2020-07-13
[3662] - Allow users to sign out when 'magic link' enabled

TobyRet:
### Context
When a user signs in via magic link and then attempts to sign out they receive an error. 
```
Error
invalid_request
unrecognized route or not allowed method (GET on /)
```

This is because the app is attempting to sign out of DFE signin.

- Enable sign out if a user has signed in via 'magic link'.

### Guidance to review
- Sign in via magic link. To this add the following to your `development.local.yml`
```
features:
  signin_intercept: true
  signin_by_email: true
  dfe_signin: false
```
- Click 'sign out' in the header. You should be redirected to `/signin`.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -73,6 +73,10 @@ class ApplicationController < ActionController::Base
     !Rails.env.production? && Settings.key?(:authorised_users)
   end
 
+  def magic_link_enabled?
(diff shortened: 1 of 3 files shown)


TobyRet:
[Sign out does not work when magic link is enabled](https://trello.com/c/gsMQt50K/3662-sign-out-does-not-work-when-magic-link-is-enabled)

PR #1257, opened 2020-07-14
3643 provider suggestion error

SamJamCul:
### Context
https://trello.com/c/7MOribyu/3643-create-a-course-error-when-leaving-accredited-body-field-empty

To fix this issue we initially looked at the API to see what errors were being thrown. It turns out that for this particular endpoint, there aren't errors when a bad query is sent, and only a `bad request` status is thrown back, with no other information. This meant that Publish would send a bad query, and be unable to understand why it was bad - and so didn't load the error partial properly. 

We've got a couple of options here 
1. We can add validation to Publish that will prevent users from submitting searches that wouldn't meet the API's specifications. This prevents bad requests being sent to the API.
2. We can add proper error messaging to the API, which would allow us to send bad queries to the API, and it wouldn't just be throwing back unhelpful json. 
3. Both of these - we want helpful info for the users, but we also shouldn't be sending stuff to the API we know will cause an error. 
4. We remove the guards on the API that prevent searches with less than 3 characters from working - they don't seem to serve much of a purpose, and it would mean users who genuinely want to search for one or two char queries will be able to. 

At the moment the changes proposed alter the validation on Publish, but we could go for other options outlined above. 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/courses/accredited_body_controller.rb
+++ b/app/controllers/courses/accredited_body_controller.rb
@@ -10,9 +10,14 @@ module Courses
     end
 
     def continue
-      other_selected_with_no_autocompleted_code = course_params[:accredited_body_code] == "other" && @autocompleted_provider_code.blank?
+      code = course_params[:accredited_body_code]
+      query = @accredited_body
 
-      if other_selected_with_no_autocompleted_code
+      @errors = errors_for_search_query(code, query)
+
+      if @errors.present?
+        render :new
+      elsif other_selected_with_no_autocompleted_code?(code)
         redirect_to(
           search_new_provider_recruitment_cycle_courses_accredited_body_path(
             query: @accredited_body,
@@ -112,7 +117,7 @@ module Courses
     def errors_for_search_query(code, query)
       errors = {}
 
-      if code == "other" && query.length < 3
+      if other_selected_with_no_autocompleted_code?(code) && query.length < 2
         errors = { accredited_body: ["Accredited body search too short, enter 2 or more characters"] }
       elsif code.blank?
         errors = { accredited_body_code: ["Pick an accredited body"] }
@@ -145,6 +150,10 @@ module Courses
       {
         accredited_body_code: autocompleted_code.presence || code,
       }
+    end
+
+    def other_selected_with_no_autocompleted_code?(code)
+      code == "other" && @autocompleted_provider_code.blank?
     end
   end
 end
--- a/spec/features/courses/accredited_body/edit_spec.rb
+++ b/spec/features/courses/accredited_body/edit_spec.rb
@@ -186,7 +186,7 @@ feature "Edit accredited body", type: :feature do
           expect(accredited_body_page).to be_displayed
           expect(accredited_body_page).to have_content("search too short")
 
-          fill_in "Name of accredited body", with: "AT"
+          fill_in "Name of accredited body", with: "A"
           click_on "Save and publish changes"
 
           expect(accredited_body_page).to be_displayed
--- a/spec/features/courses/accredited_body/new_spec.rb
+++ b/spec/features/courses/accredited_body/new_spec.rb
@@ -1,6 +1,6 @@
 require "rails_helper"
 
-feature "Edit accredited body" do
+feature "New accredited body" do
   let(:current_recruitment_cycle) { build(:recruitment_cycle) }
   let(:provider) { build(:provider, sites: [build(:site)]) }
   let(:course) { build(:course, provider: provider) }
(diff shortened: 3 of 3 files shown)


SamJamCul:
[[?] Create a course: Error when leaving Accredited Body field empty](https://trello.com/c/7MOribyu/3643-create-a-course-error-when-leaving-accredited-body-field-empty)

app/controllers/courses/accredited_body_controller.rb
```diff
@@ -11,8 +11,12 @@ def edit
 
     def continue
       other_selected_with_no_autocompleted_code = course_params[:accredited_body_code] == "other" && @autocompleted_provider_code.blank?
+      @errors = { accredited_body: ["Please enter at least three letters of the name or provider code"] } if @accredited_body.length < 3
+      @errors = { accredited_body: ["Please enter the name or provider code"] } if @accredited_body.empty?
```
misaka, line 15: Can we move these into methods that help guide us through what we're doing here? Also it'd be nice to not rely on ordering of these validations as this implementation does.

```suggestion
      validate_accredited_body_name_length
      validate_accredited_body_name_presence
```

And these methods might look like:

```ruby
    def validate_accredited_body_name_length
      @errors[accredited_body] ||= []
      if @accredited_body.length > 0 && @accredited_body.length < 3
        @errors[accredited_body] << "Please enter at least three letters of the name or provider code"
      end
    end
```

SamJamCul:
> In addition to the suggestions here, I think we're missing the validation to ensure that a user selects an accredited body (one of the radio buttons on this page). We should add that here, but also we should have a validation in the API for this field, but that deserves it's own card.

It looks like it's meant to be possible for courses to be created without an accredited body - [this is an example](https://www.find-postgraduate-teacher-training.service.gov.uk/course/B28/2KN9) of a course that doesn't have one, and is live. 

We should work out what the design should be for users who get to this page but don't want to set an accredited body - I don't want to add validation that would prevent them at the moment, as this could cause issues.

SamJamCul:
@defong have been trying to add a suitable test for this - we're having some issues creating a new accredited_body_controller test, and we don't really have anything else in place that we can work from. Does anyone have any experience with creating controller tests for this area of the code?

misaka:
> > In addition to the suggestions here, I think we're missing the validation to ensure that a user selects an accredited body (one of the radio buttons on this page). We should add that here, but also we should have a validation in the API for this field, but that deserves it's own card.
> 
> It looks like it's meant to be possible for courses to be created without an accredited body - [this is an example](https://www.find-postgraduate-teacher-training.service.gov.uk/course/B28/2KN9) of a course that doesn't have one, and is live.

Courses can't not have an accredited body, which is why imo we should not allow `nil` here but always have a value, even if it's the same as the training provider.

Are we saying that the user should be able to leave this blank so that they can create a self-accredited course? This doesn't sound right to me, we already have a data problem so I'd be weary of allowing it to get any worse.

misaka:
> We've got a couple of options here
>
>   1. We can add validation to Publish that will prevent users from submitting searches that wouldn't meet the API's specifications. This prevents bad requests being sent to the API.

I think that this approach makes sense to me, as implemented here. I'm not sure how much need there is for the API to be more specific about the error ... it's not like we're doing a POST and a particular attribute is causing problems. Unless we need to discern between errors (or there's a strong reason around API design, which is also possible) I'm not sure I'd put the effort into that at this stage.

>   4. We remove the guards on the API that prevent searches with less than 3 characters from working - they don't seem to serve much of a purpose, and it would mean users who genuinely want to search for one or two char queries will be able to.

Like @TobyRet mentioned, this was originally brought in as a throttling measure. Is there a user need to remove this? I feel like removing it would take more thought than not removing it, so unless there's a need I'd probably just leave it and make the changeset smaller.

SamJamCul:
> > > In addition to the suggestions here, I think we're missing the validation to ensure that a user selects an accredited body (one of the radio buttons on this page). We should add that here, but also we should have a validation in the API for this field, but that deserves it's own card.
> > 
> > 
> > It looks like it's meant to be possible for courses to be created without an accredited body - [this is an example](https://www.find-postgraduate-teacher-training.service.gov.uk/course/B28/2KN9) of a course that doesn't have one, and is live.
> 
> Courses can't not have an accredited body, which is why imo we should not allow `nil` here but always have a value, even if it's the same as the training provider.
> 
> Are we saying that the user should be able to leave this blank so that they can create a self-accredited course? This doesn't sound right to me, we already have a data problem so I'd be weary of allowing it to get any worse.

It feels like we need a wider discussion around how users are meant to create self-accredited courses. At the moment we're implicitly presuming that accredited bodies will always self accredit courses that they create. This means we would never need to check who the accredited body is for a course they're creating, because we already 'know' that it's them.

misaka:
> It feels like we need a wider discussion around how users are meant to create self-accredited courses. At the moment we're implicitly presuming that accredited bodies will always self accredit courses that they create. This means we would never need to check who the accredited body is for a course they're creating, because we already 'know' that it's them.

We just had a call about this, and have discovered that:

- accredited bodies don't see this page, hence we don't need to consider them here
- the list of suggested providers only includes accredited bodies
- the page will bring up a list of accredited bodies that the course's training provider already works with, and optionally allow them to enter in a new accredited body through the here-discussed form field (the example taken in the Trello card has no courses, which is why that page came up with no providers)
- we do actually have 4 accredited bodies who have courses accredited by a different accredited body, but we think those are miss-use of the system, or errors, so we should go through support channels if they want to create a non-self-accredited course

I think what we want for this ticket is what was originally outlined as option 1:

> 1. We can add validation to Publish that will prevent users from submitting searches that wouldn't meet the API's specifications. This prevents bad requests being sent to the API.

Basically, validate that the user has submitted something, and don't let them create a course with a blank accredited provider code.

app/controllers/courses/accredited_body_controller.rb
```diff
@@ -10,9 +10,14 @@ def edit
...
-      other_selected_with_no_autocompleted_code = course_params[:accredited_body_code] == "other" && @autocompleted_provider_code.blank?
+      code = course_params[:accredited_body_code]
+      query = @accredited_body
 
-      if other_selected_with_no_autocompleted_code
+      @errors = errors_for_search_query(code, query)
+
+      if @errors.present?
+        render :new
+      elsif other_selected_with_no_autocompleted_code && @errors.blank?
```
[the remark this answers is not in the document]
SamJamCul: Good point!

app/controllers/courses/accredited_body_controller.rb
```diff
@@ -112,7 +117,7 @@ def build_course_params
     def errors_for_search_query(code, query)
       errors = {}
 
-      if code == "other" && query.length < 3
+      if code == "other" && query.length < 2
```
misaka, line 120: As discussed elsewhere, can we use `other_selected_with_no_autocompleted_code` here?

app/controllers/courses/accredited_body_controller.rb
```diff
@@ -146,5 +151,9 @@ def update_params
         accredited_body_code: autocompleted_code.presence || code,
       }
     end
+
+    def other_selected_with_no_autocompleted_code
```
misaka, line 155:
```suggestion
    def other_selected_with_no_autocompleted_code?
```

app/controllers/courses/accredited_body_controller.rb
SamJamCul: Done!

PR #1262, opened 2020-07-17
[3686] Fix character count on Organisation "About" page

aliuk2012:
- Updates the form on the page to use GOV.UK Design System formbuilder
- Fixes issue where accredited body optional fields were displaying as errors if there were any validation errors for "training with us" and last text area "disabilities".
- removes old copy telling users it would take up to 2hrs for content to be updated on Find.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -99,7 +99,7 @@ class ProvidersController < ApplicationController
 private
 
   def provider_params
-    params.require(:provider).permit(
+    permitted_params = params.require(:provider).permit(
       :page,
       :train_with_us,
       :train_with_disability,
@@ -113,10 +113,31 @@ private
       :address4,
       :postcode,
       :region_code,
-      accredited_bodies: %i[provider_name provider_code description],
+      accredited_bodies_attributes: %i[provider_name provider_code description],
     ).to_h # Without this, accredited_bodies is an array of params objects
     # instead of an array of plain hashes and gets serialized incorrectly
     # on its way to the backend.
+    permitted_params.merge(accredited_bodies_param).except(:accredited_bodies_attributes)
+  end
+
+  def accredited_bodies_param
+    ab = params.require(:provider).except(
+      :page,
+      :train_with_us,
+      :train_with_disability,
+      :provider_name,
+      :email,
+      :telephone,
+      :website,
+      :address1,
+      :address2,
+      :address3,
+      :address4,
+      :postcode,
+      :region_code,
+    )
+      .permit(accredited_bodies_attributes: %i[provider_name provider_code description])
+    { accredited_bodies: ab[:accredited_bodies_attributes].to_h.map { |_k, v| v } }
   end
 
   def build_provider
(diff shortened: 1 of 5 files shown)


app/controllers/providers_controller.rb
```diff
@@ -113,10 +113,31 @@ def provider_params
...
+      :website,
+      :address1,
+      :address2,
+      :address3,
+      :address4,
+      :postcode,
+      :region_code,
+    )
+      .permit(accredited_bodies_attributes: %i[provider_name provider_code description])
+    { accredited_bodies: ab[:accredited_bodies_attributes].to_h.map { |_k, v| v } }
```
asmega, line 140: fyi you can do `{a:1, b:2, c:3}.values` => `[1, 2, 3]`

aliuk2012: Thanks @asmega, I wasn't aware that you could do that. I did try your suggestion but then it opened up a whole can of worms because sometimes `ab[:accredited_bodies_attributes]` is nil and therefore `.values` raises an error. I tried to add a guard to the method and it soon went into a downward spiral from there.

benilovj: Al, you could try

`{ accredited_bodies: ab[:accredited_bodies_attributes]&.values }`

if the downstream code handles the `:accredited_bodies` key being `nil`, or alternatively you could do:

`{ accredited_bodies: ab[:accredited_bodies_attributes]&.values || [] }`

which will always return an array for `:accredited_bodies`.

aliuk2012: I gave this a try and then all the stubs request tests started to fail and it looks like the stubs need to just be a hash where as using `.values` seems to directly copy the values which still makes them `ActionController::Parameters` rather than a plain hash.

```[1] pry(#<ProvidersController>)> a = { accredited_bodies: ab[:accredited_bodies_attributes].to_h.map { |_k, v| v }}
=> {:accredited_bodies=>[{"provider_name"=>"Baz", "provider_code"=>"Z01", "description"=>"Baz"}, {"provider_name"=>"Qux", "provider_code"=>"Z02", "description"=>"Qux"}]}
[2] pry(#<ProvidersController>)> b = { accredited_bodies: ab[:accredited_bodies_attributes]&.values || [] }
=> {:accredited_bodies=>
  [<ActionController::Parameters {"provider_name"=>"Baz", "provider_code"=>"Z01", "description"=>"Baz"} permitted: true>, <ActionController::Parameters {"provider_name"=>"Qux", "provider_code"=>"Z02", "description"=>"Qux"} permitted: true>]}
[3] pry(#<ProvidersController>)> b = { accredited_bodies: ab[:accredited_bodies_attributes]&.values.to_h || [] }
```

benilovj: Sorry, you're right.

I think this works:
```suggestion
    { accredited_bodies: ab[:accredited_bodies_attributes].to_h.values }
```

aliuk2012: Thanks @benilovj  that worked perfectly

aliuk2012:
[About organisation page word count issue](https://trello.com/c/SFaHlQZn/3686-about-organisation-page-word-count-issue)

PR #1263, opened 2020-07-18
[3699] Fix character counter on "About course" page

aliuk2012:
- Switches the page to use GOV.UK Design System form builder
- Fixes styling issues where is any field had a validation error then all the fields were styled as having errors.
- Clicking on the Error summary should now take the user directly to the text area and set focus to it.

## Before
image

## After
image

### Guidance to review

- Login as a user and try the following
- visit any courses "About page" and see how many words they have left for each textarea on the page
	- edit each text area and notice how the word counter below that text area updates as you add/remove words
- Attempt to add more words than each text area allows. 
	- clicking on the link in the error summary should take you directly to that field and it should have focus (yellow border) 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/views/courses/about.html.erb
+++ b/app/views/courses/about.html.erb
@@ -54,7 +54,11 @@
           </span>
         </summary>
         <div class="govuk-details__text">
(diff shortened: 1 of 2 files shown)


aliuk2012:
[About course page word count issue](https://trello.com/c/T6U0L137/3699-about-course-page-word-count-issue)

PR #1264, opened 2020-07-20
[3644] age range validation

asmega:
### Context

- https://trello.com/c/xsGZiaX2/3644-age-range-validation-message-link
- On attempt to publish with missing age range the error summary link would not navigate to the age range page so the user could fix the issue

- On attempting to publish if age range is missing the validation summary link now goes to the age range page
- From this link on page load of the age range page errors are displayed to the user
- The age range page now uses a form object
- Validations have been moved from the controller to the form object
- The age range page now uses the gov uk formbuilder
- Browser side validations have been ported to publish side validations
- Site prism page objects no longer use data-qa references and now use ids
- Some validation messages have been improved but can still do with improvement but is outside the scope of this change

### Guidance to review

- https://s121d02-3644-ptt-as.azurewebsites.net/
- Login as an admin
- Find a rolled over course that does not have an age range eg `/organisations/17K/2021/courses`
- Attempt to publish the course
- Should see error summary with link to age range error
- Click on this error
- Should be take to age range page
- Errors should be displayed on this page on page load

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/form_objects/age_range_form.rb
+++ b/app/form_objects/age_range_form.rb
@@ -1,0 +1,94 @@
+class AgeRangeForm
+  include ActiveModel::Model
+
+  attr_accessor :age_range_in_years, :presets
+  attr_reader :course_age_range_in_years_other_from,
+              :course_age_range_in_years_other_to
+
+  def course_age_range_in_years_other_from=(value)
+    @course_age_range_in_years_other_from = if value.present?
+                                              value.to_i
+                                            else
+                                              value
+                                            end
+  end
+
+  def course_age_range_in_years_other_to=(value)
+    @course_age_range_in_years_other_to = if value.present?
+                                            value.to_i
+                                          else
+                                            value
+                                          end
+  end
+
+  def initialize(args)
+    super
+
+    if process_custom_range?
+      self.course_age_range_in_years_other_from = extract_from_years
+      self.course_age_range_in_years_other_to = extract_to_years
+      self.age_range_in_years = "other"
+    end
+  end
+
+  validates :age_range_in_years, presence: { message: I18n.t("age_range.errors.missing_error") }
+  validates :course_age_range_in_years_other_from, numericality: {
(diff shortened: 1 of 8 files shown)


app/form_objects/age_range_form.rb
```diff
@@ -0,0 +1,94 @@
...
+  attr_accessor :age_range_in_years, :presets
+  attr_reader :course_age_range_in_years_other_from,
+              :course_age_range_in_years_other_to
+
+  def course_age_range_in_years_other_from=(value)
+    @course_age_range_in_years_other_from = if value.present?
+                                              value.to_i
+                                            else
+                                              value
+                                            end
```
aliuk2012, line 13: Would it be better to have the `if` statement in a method called something like `get_value`. This could be used here and in the `course_age_range_in_years_other_to` method. Or perhaps shorthand? `(value.present?)? value.to_i : value`

asmega: - I don't think `get_value` makes sense as that is passed into the method as it not available elsewhere
- Changing to a one-line ternary here I don't think really improves readability and would merely be for vanity sake

defong: having a hard time thinking what is `value` 

Maybe it should be 
```suggestion
    @course_age_range_in_years_other_from = value.to_i if value.present?
```

we got a lot of 

`course_age_range_in_years_other_from.blank?`
`course_age_range_in_years_other_from.present?`
`course_age_range_in_years_other_from.to_i`

asmega: I was wanting to capture empty string which your version does not and instead would set to `nil`. That way you can run validations of what the user actually entered. The alternative is to make virtual attributes (another method and variable) for the user input and another set for a processed version that we can use later

PR #1269, opened 2020-07-24
[3678] - Course vacancies open (all locations)

TobyRet:
### Context
Currently, subscribed users are only being notified if a course has no vacancies (all locations).

See associated API PR - https://github.com/DFE-Digital/teacher-training-api/pull/1483. This also has notification examples.

- When a user updates course locations to from closed to open we want to trigger the vacancies updated notification. Specifically, the notification is triggered when a course has a single location and it is updated from closed to open, or when the course has multiple locations and all locations are changed to open.

### Guidance to review
- To run locally you will need to run this branch against  https://github.com/DFE-Digital/teacher-training-api/pull/1483 and have a Notify test key configured

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/courses/vacancies_controller.rb
+++ b/app/controllers/courses/vacancies_controller.rb
@@ -13,7 +13,12 @@ module Courses
         return render(:edit)
       end
 
-      @course.has_multiple_running_sites_or_study_modes? ? update_vacancies_for_multiple_sites : update_vacancies_for_a_single_site
+      if @course.has_multiple_running_sites_or_study_modes?
+        update_vacancies_for_multiple_sites
+      else
+        update_vacancies_for_a_single_site
+      end
+
       flash[:success] = "Course vacancies published"
       redirect_to provider_recruitment_cycle_courses_path(params[:provider_code], params[:recruitment_cycle_year])
     end
@@ -44,7 +49,7 @@ module Courses
         site_status.save
       end
 
-      @course.send_vacancies_filled_notification if single_site_no_vacancies?
+      @course.send_vacancies_updated_notification({ vacancies_filled: single_site_no_vacancies? })
     end
 
     def update_vacancies_for_multiple_sites
@@ -66,7 +71,8 @@ module Courses
         site_status.save
       end
 
-      @course.send_vacancies_filled_notification if multiple_sites_no_vacancies?
+      @course.send_vacancies_updated_notification({ vacancies_filled: true }) if all_sites_have_no_vacancies?
+      @course.send_vacancies_updated_notification({ vacancies_filled: false }) if all_sites_have_vacancies?
     end
 
     def build_course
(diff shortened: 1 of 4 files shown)


TobyRet:
[Vacancies open - (all open)](https://trello.com/c/FXm1vJhZ/3678-vacancies-open-all-open)

app/controllers/courses/vacancies_controller.rb
```diff
@@ -90,16 +96,24 @@ def single_site_no_vacancies?
       params[:change_vacancies_confirmation] == "no_vacancies_confirmation" && @course.has_vacancies? == true
     end
 
-    def multiple_sites_no_vacancies?
-      vacancy_statuses = params.dig(:course, :site_status_attributes)
+    def all_sites_have_no_vacancies?
+      @course.has_vacancies? == true &&
```
asmega, line 100:
```suggestion
      @course.has_vacancies? &&
```

app/models/base.rb
```diff
@@ -27,9 +27,9 @@ def run(request_method, path, params: nil, headers: {}, body: nil)
 
 private
 
-  def post_request(path)
+  def post_request(path, attributes = {})
     post_options = {
-      body: { data: { attributes: {}, type: self.class.to_s.downcase } },
+      body: { data: { attributes: attributes, type: self.class.to_s.downcase } },
```
asmega, line 32: I feel this will open the API for abuse with ad-hoc endpoints mutating data. I feel if https://github.com/DFE-Digital/teacher-training-api/pull/1483#discussion_r459972237 is addressed none of these changes are needed

TobyRet: This endpoint doesn't mutate a resource - in essence it's triggering an event. I've added some further comments here with options - https://github.com/DFE-Digital/teacher-training-api/pull/1483#discussion_r460704748

asmega: So in this particular case, yes, you're not mutating a resource but the capability to be able to do so is now possible through this mechanism

misaka:
@TobyRet Just a thought, the alternative to this is to have a custom business-logic endpoint that marks specific sites as having / not having vacancies, and that handles the notification all in one go. This means that we don't update the resource using the standard JSONAPI mechanism, but I don't think that that's a problem; while adding bulk-update semantics to resources is tricky, adding separate business-logic endpoints should be fine.

app/models/base.rb
TobyRet: I've added a new custom endpoint in order to address the concerns above - see https://github.com/DFE-Digital/publish-teacher-training/pull/1269/files#diff-72aea198f50b1d4f2407a8f9f4c904f6R15

TobyRet:
> @TobyRet Just a thought, the alternative to this is to have a custom business-logic endpoint that marks specific sites as having / not having vacancies, and that handles the notification all in one go. This means that we don't update the resource using the standard JSONAPI mechanism, but I don't think that that's a problem; while adding bulk-update semantics to resources is tricky, adding separate business-logic endpoints should be fine.

@asmega - have agreed to merge as is for now. Will seek to clarify @misaka 's comment above when he gets back from hols.

PR #1270, opened 2020-07-24
[3642] Amended log safe current user to be lean

defong:
### Context
Currently the log safe current user was logging way to much details

Amended log safe current user to be lean

### Guidance to review
``` ruby
      {
        "email_md5"=>"email_md5", 
        "sign_in_user_id"=>"sign_in_user_id", 
        "user_id"=>"user_id"
      }
```

Configure from config/settings.yml or config/settings/development.local.yml
```yml
logstash:
  type: tcp
  host: # Our hostname here
  port: # Our port here
  ssl_enable: true
```
Logs will be much more compact see [kibana example](https://kibana.logit.io/app/kibana#/doc/8ac115c0-aac1-11e8-88ea-0383c11b333a/logstash-2020.07.24?id=xrzQgHMB9CPnJIZAXLuS&_g=(refreshInterval:(pause:!t,value:0),time:(from:now-1y,to:now)))

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -61,10 +61,13 @@ class ApplicationController < ActionController::Base
 
   def log_safe_current_user(reload: false)
     if @log_safe_current_user.nil? || reload
-      @log_safe_current_user = current_user.dup
-      email = @log_safe_current_user["info"]&.fetch("email", "")
-      @log_safe_current_user.delete("info")
-      @log_safe_current_user["email_md5"] = Digest::MD5.hexdigest(email)
+      email = current_user["info"]&.fetch("email", "")
+
+      @log_safe_current_user = {
+        email_md5: Digest::MD5.hexdigest(email),
+        sign_in_user_id: current_user["uid"],
+        user_id: current_user["user_id"],
+      }
     end
     @log_safe_current_user
   end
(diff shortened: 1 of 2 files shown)


defong:
[[?] Don't log user's email](https://trello.com/c/PtxMKHwF/3642-dont-log-users-email)

app/controllers/application_controller.rb
```diff
@@ -61,10 +61,13 @@ def user_state
...
   def log_safe_current_user(reload: false)
     if @log_safe_current_user.nil? || reload
-      @log_safe_current_user = current_user.dup
-      email = @log_safe_current_user["info"]&.fetch("email", "")
-      @log_safe_current_user.delete("info")
-      @log_safe_current_user["email_md5"] = Digest::MD5.hexdigest(email)
+      email = current_user["info"]&.fetch("email", "")
+
+      @log_safe_current_user = {
+        email_md5: Digest::MD5.hexdigest(email),
```
asmega, line 67: is there any point having this given we have the 2 lines below? I don't see a scenario where we'd have to md5 every email address in the db to find the associated record

defong: I don't mind dropping it, through it was added due to 
https://trello.com/c/OhXYtbKP/1396-3x-users-incorrectly-unable-to-get-past-unknown-org-page

@misaka do we still need the it?

asmega: I think it's fine to leave then if it is needed for the edge case

PR #1281, opened 2020-07-30
[3761] add cuprite and basic system spec

misaka:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/system/happy_path_spec.rb
+++ b/spec/system/happy_path_spec.rb
@@ -1,0 +1,59 @@
+require "spec_system_helper"
+
+RSpec.feature "happy path", type: :system do
+  let(:login_page) { PageObjects::Page::DFESignIn::LoginPage.new }
+  let(:rollover_page) { PageObjects::Page::Rollover.new }
+  let(:rollover_recruitment_page) { PageObjects::Page::RolloverRecruitment.new }
+
+  scenario "use is happy" do
+
+    when_i_visit_the_root_path
+    then_i_am_redirected_to_dfe_login_page
+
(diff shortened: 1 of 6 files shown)


spec/system/happy_path_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+  let(:rollover_recruitment_page) { PageObjects::Page::RolloverRecruitment.new }
+
+  scenario "use is happy" do
+
+    when_i_visit_the_root_path
+    then_i_am_redirected_to_dfe_login_page
+
+    when_i_fill_in_my_username_and_password
+    when_i_click_sign_in_on_the_login_page
+    then_i_am_on_the_rollover_page
```
asmega, line 15: i think this will work the first time around. but after you've acknowledged i believe you don't see this page again so the test fail every subsequent time

spec/system/happy_path_spec.rb
```diff
@@ -0,0 +1,59 @@
...
+
+    when_i_click_continue_on_the_rollover_page
+    then_i_am_on_the_rollover_recruitment_page
+
+    when_i_click_continue_on_the_rollover_recruitment_page
+
+    expect(rollover_recruitment_page).to be_displayed
+    rollover_recruitment_page.continue.click
+
+    expect(root_page).to be_displayed
```
asmega, line 25: where is `root_page` defined?

misaka:
Will have to pick this up again if/when we want to finish it.

PR #1283, opened 2020-07-31
[3760] Fix copy course content

asmega:
### Context

- https://trello.com/c/fWHZmhOc/3760-copy-course-content-bug
- Fixes a regression that broke copy course content

- Previously there were 2 independent forms on the course about page
- A regression in the HTML markup caused these 2 forms to be nested and therefore invalid
- This change splits the forms so they are no longer nested
- A third non-submitable form has been added so if there are errors these can be rendered back to the user

### Guidance to review

- Login as admin or user with 2 courses
- Edit the course through the about page
- On the RHS use the copy content from another course feature
- Other course content should now be present on this course

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/views/courses/about.html.erb
+++ b/app/views/courses/about.html.erb
@@ -13,6 +13,7 @@
               data: { qa: 'enrichment-form', module: 'form-check-leave' } do |f| %>
   <%= f.hidden_field :page, value: :about %>
   <%= f.govuk_error_summary "You’ll need to correct some information." %>
+<% end %>
 
   <h1 class="govuk-heading-xl">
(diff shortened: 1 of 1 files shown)


asmega:
[Copy course content bug](https://trello.com/c/fWHZmhOc/3760-copy-course-content-bug)

asmega:
> any way we can add test to prevent regression?

Turns out there is a regression test in place. However since we are using the Rack Test driver I don't think it picked up on the subtlety of the nested form. We'd have to switchover this test to headless chrome driver first

aliuk2012:
I am reviewing this

PR #1297, opened 2020-08-10
Fixes issue with rspec parallel xml outputs.

aliuk2012:
Our currently implementation seems not to be working anyway. To simplify
our process we are now using `.rspec_parallel` as a single config file
Azure should be able to use this output to report test coverage in azure.

- Tested locally running `bundle exec rake parallel:spec` and inspected the logged out xm files found in `rspec-output`.
### Guidance to review
- view "Tests" tab for the azure job once it has been completed
image

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/.rspec_parallel
+++ b/.rspec_parallel
@@ -1,2 +1,3 @@
 --format progress
 --format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
+--require './spec/spec_helper.rb' --no-profile  --format RspecJunitFormatter --out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml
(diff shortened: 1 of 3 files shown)


.rspec_parallel
```diff
@@ -1,2 +1,3 @@
 --format progress
 --format ParallelTests::RSpec::RuntimeLogger --out tmp/parallel_runtime_rspec.log
+--require './spec/spec_helper.rb' --no-profile  --format RspecJunitFormatter --out rspec-output/rspec-output<%= ENV['TEST_ENV_NUMBER'] %>.xml
```
asmega, line 3: it should work without the `--require` option

PR #1299, opened 2020-08-11
[3756] Add "confirmed" state views for Allocations

aliuk2012:
Moved Settings.allocations_state into the features namespace. Also added
comments around the different valid options.

We decided to keep `state` key as it could potentially be used by other
methods / backend business logic

### Context

https://trello.com/c/mAtodzq4/3756-s-implement-allocations-results-screen

* Added allocations confirmed page templates
* Add view object method for returning only confirmed allocations
* Added tests for confirmed allocations view
* Changed allocation config settings naming
* Changed QA allocations feature state to confirmed for product review

### Guidance to review
- visit https://s121d02-3756-ptt-as.azurewebsites.net/ 
- login as a user of an accredited body
- view allocation page

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/view_objects/allocations_view.rb
+++ b/app/view_objects/allocations_view.rb
@@ -53,6 +53,14 @@ class AllocationsView
     statuses.compact.sort_by! { |hsh| hsh[:training_provider_name] }
   end
 
+  def confirmed_allocation_places
+    statuses = confirmed_allocations.map do |allocation|
+      build_confirmed_allocations(allocation, allocation.provider)
+    end
+
+    statuses.compact.sort_by! { |hsh| hsh[:training_provider_name] }
+  end
+
   def not_requested_allocations_statuses
     statuses = filtered_training_providers.map do |training_provider|
       matching_allocation = find_matching_allocation(training_provider, not_requested_allocations)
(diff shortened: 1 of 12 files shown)


SamJamCul:
[[S] Implement Allocations results screen](https://trello.com/c/mAtodzq4/3756-s-implement-allocations-results-screen)

app/view_objects/allocations_view.rb
```diff
@@ -85,6 +93,10 @@ def requested_allocations
     @allocations.select { |allocation| allocation.request_type.in?([RequestType::INITIAL, RequestType::REPEAT]) }
   end
 
+  def confirmed_allocations
+    @allocations.select { |allocation| allocation.request_type.in?([RequestType::INITIAL, RequestType::REPEAT]) }
```
asmega, line 97: Once we get the data finalised, this is the bit that will probably need tweaking to display the final numbers

app/views/providers/allocations/_allocation-request-confirmed-state.html.erb
```diff
@@ -0,0 +1,36 @@
+
+<%= content_for :page_title, raw("PE courses for 2021 to 2022") %>
```
asmega, line 2: not important but `raw` is not needed

aliuk2012: Good spot, lets remove that

app/view_objects/allocations_view.rb
aliuk2012: Why would this change? Are you thinking it should be another request type?

app/views/providers/allocations/_allocation-request-confirmed-state.html.erb
JonF1: Removed

PR #1305, opened 2020-08-13
[3782] Rename allocation state partials and paths

JonF1:
### Context
Allocation state partials are inconsisently named with a mix of hyphens and underscores. 

Renaming these partials to use underscores consistently/snakecase and a render partial path to recognise the rename

### Guidance to review
Screenshot 2020-08-13 at 12 31 04

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review


JonF1:
[[S]Rename three allocation filenames which are using hyphens instead of underscores](https://trello.com/c/g4NLgeau/3782-srename-three-allocation-filenames-which-are-using-hyphens-instead-of-underscores)

PR #1316, opened 2020-08-18
[3772] clean up basic details tab after fe course is added

JonF1:
### Context
*https://trello.com/c/a0Du6GZo/3772-clean-up-basic-details-tab-after-fe-course-is-added

Firstly, the 'changing your basic details' part of the basic details tab after a course is added was flawed because it would show the same information regardless of the level of course. As GCSE requirements and Age Range do not show up in the course creation set up of Further Education courses, it does not make sense to include them within this. 

Secondly, we reuse the same/very similar attributes for both the basic details tab and the confirmation page (in the course creation set up). So it makes sense to use view components going forward. This PR includes a basic level component but further thought is needed about how to fully implement view components/how granular they need to be etc. 

A decorator method to change 'changing your basic details' according to course level
A view component which refactors the level attribute in the basic details and confirmation view partials. View component gem was added. 

### Guidance to review
* Add both a Further Education and Primary/Secondary course and you can see the 'changing your basic details' section change
* Level attribute should render as normal on both partials, with the backend code just looking a bit neater and more findable. 

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


app/components/course_level_component.html.erb
```diff
@@ -0,0 +1,17 @@
+<div class="govuk-summary-list__row">
+  <dt class="govuk-summary-list__key">
+    Level
+  </dt>
+  <dd class="govuk-summary-list__value" data-qa="course__level">
+    <%= course.level&.humanize %>
+  </dd>
+  <dd class="govuk-summary-list__actions">
+    <% if changeable %>
+      <%= course_creation_change_button(
```
aliuk2012, line 10: I understand this is using a view helper but couldn't we use bunk standard rails `link_to` instead of having a view helper to output a link.

aliuk2012: I think this component should ideally be self contained

aliuk2012:
Why does PR include a "yarn upgrade" commit?

JonF1:
[[M] Clean up Basic Details sidebar tab after FE course is added](https://trello.com/c/a0Du6GZo/3772-m-clean-up-basic-details-sidebar-tab-after-fe-course-is-added)

app/components/course_level_component.html.erb
JonF1: Thanks for your help

spec/site_prism/page_objects/page/organisations/course_details.rb
```diff
@@ -32,6 +32,7 @@ class CourseDetails < CourseBase
         element :level, "[data-qa=course__level]"
         element :entry_requirements, "[data-qa=course__entry_requirements]"
         element :allocations_info, "[data-qa=course__allocations_info]"
+        element :changing_your_basic_details, "[data-qa=course__changing_your_basic_details]"
```
defong, line 35: sanity check @aliuk2012 we dropping `data-qa` right?

aliuk2012: Yes are are in the process of dropping `data-qa`

defong (approved):

app/views/courses/_basic_details_tab.html.erb
```diff
@@ -286,13 +278,9 @@
...
           <li><abbr class="app-text-decoration-underline-dotted" title="Special educational needs and disability">SEND</abbr> specialism</li>
-          <li>age range</li>
-          <li>outcome</li>
-          <li>full time or part time</li>
-          <li>course locations</li>
-          <li>application open date</li>
-          <li>course start date</li>
-          <li>GCSE requirements</li>
+          <% course.listing_basic_details.each do |details|%>
+              <li><%= "#{details}" %></li>
```
aliuk2012, line 282: Not a blocker, but I don't thinks details needed to be wrapped in double quotes
```
<%= details %>
```

aliuk2012 (approved):

app/views/courses/_basic_details_tab.html.erb
JonF1: Thanks Al, i've changed it

conorlyster (approved):
Product review approved

PR #1327, opened 2020-08-26
59 healthcheck statuscake

felixtheflex:
### Context

Configured statuscake's Terraform pipeline to monitor Publish-teacher-training's website using Ping. 'healthcheck is not working at the moment' 

We have the status cakes. We have the /healthchecks. But /healthcheck on the API is returning false (Redis queues?).

API
* Figure out what is causing it to return false, fix it.

Publish
* Do we want to be using /ping or /healthcheck on API to determine our status? Sounds to me like it should be /ping, so this would need changing.

Terraform:- configured statuscake's monitor url for each environment to */ping

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/terraform/backend.tf
+++ b/terraform/backend.tf
@@ -1,0 +1,5 @@
+terraform {
+  backend "azurerm" {
+    container_name       = "tfstatestr"
+  }
+}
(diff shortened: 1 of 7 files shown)


saliceti (approved):
OK

vigneshmsft (approved):

PR #1333, opened 2020-09-01
[32] Changes 'for example' phrase to a valid date

JonF1:
## Context
 * https://trello.com/c/d79wSVBk/32-check-that-the-when-will-applications-open-error-message-displays-the-correct-dates
* https://github.com/DFE-Digital/teacher-training-api/pull/1557

The 'for example' phrase was an invalid date. It needed to be between 13/10/2020 and 3/10/2021.

For example phrase was changed to 30/9/2021 (just an arbitrary date) that is valid because it's within the date range for the 2020 to 2021 recuitment cycle. Also, a simple copy change to change the '-' to a 'to' in line with the rest of Publish for '2020 - 2021'. 

### Guidance to review
When adding a new course you see the change when the partial is rendered for the 'when will the applications open' page

### Checklist

- [X] Attach to Trello card
- [X] Rebased master
- [X] Cleaned commit history
- [X] Product Review

### DIFF
--- a/app/views/courses/applications_open/_form_fields.html.erb
+++ b/app/views/courses/applications_open/_form_fields.html.erb
@@ -31,7 +31,7 @@
             What date will applications open?
           </legend>
           <span class="govuk-hint">
-            For example, 31 10 <%= @recruitment_cycle.year %>
+            For example, 30 9 <%= @recruitment_cycle.year %>
           </span>
           <div class="govuk-date-input">
             <div class="govuk-date-input__item">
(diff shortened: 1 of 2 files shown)


JonF1:
[[S] check that the 'when will applications open' error message displays the correct dates.](https://trello.com/c/d79wSVBk/32-check-that-the-when-will-applications-open-error-message-displays-the-correct-dates)

app/views/courses/applications_open/_form_fields.html.erb
```diff
@@ -31,7 +31,7 @@
             What date will applications open?
           </legend>
           <span class="govuk-hint">
-            For example, 31 10 <%= @recruitment_cycle.year %>
+            For example, 30 09 <%= @recruitment_cycle.year %>
```
asmega, line 34: Works without the padded zero so i think example should omit?:
```suggestion
            For example, 30 9 <%= @recruitment_cycle.year %>
```

JonF1: Thanks Phil

asmega (approved):

defong (approved):

PR #1352, opened 2020-09-14
[0096] Added the financial placeholder

defong:
### Context
During the transition from old cycle to new cycle to current cycle, the financial incentives have not been confirmed.

Added a settings in config to determine the cycle in question should have a placeholder to state along the lines of tba.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review

### DIFF
--- a/config/settings.yml
+++ b/config/settings.yml
@@ -50,6 +50,10 @@ environment:
   label: "Beta"
   selector_name: "beta"
 current_cycle: 2020
+# `financial_support_placeholder_cycle` the cycle year value should be 
+# omitted if placeholder is not required otherwise it should be the
+# new/next cycle.
+financial_support_placeholder_cycle: 2021
 application_name: publish-teacher-training
 logstash:
   type: tcp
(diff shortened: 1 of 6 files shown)


defong:
[Replace Financial Support information with placeholder in Publish preview](https://trello.com/c/1iSUsibh/96-replace-financial-support-information-with-placeholder-in-publish-preview)

config/settings.yml
```diff
@@ -50,6 +50,10 @@ environment:
   label: "Beta"
   selector_name: "beta"
 current_cycle: 2020
+# `financial_support_placeholder_cycle` the cycle year value should be 
+# omitted if placeholder is not required otherwise the it should be the
+# new cycle.
```
avinhurry, line 55: Just wondering if its usual to have comments like this in the code?

conorlyster (approved):
Product review approved

avinhurry (approved):

config/settings.yml
gpeng: I think it's ok where the code isn't self documenting and there's a potential for confusion or a lack of clarity. Obviously, we should write self documenting code that isn't confusing but we all have off days 🤣

aliuk2012 (approved):

config/settings.yml
defong: it really confusing... and i never have a non-off day

it that period where it goes from 2 (current/next) cycles and back to 1 (new/current) cycle

PR #1420, opened 2020-11-05
[379] Use Settings recruitment cycle in the routes

aliuk2012:
### Context
We shouldn't need to hardcode the current/next recruitment cycle year.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/config/routes.rb
+++ b/config/routes.rb
@@ -68,8 +68,7 @@ Rails.application.routes.draw do
 
     resources :contacts, only: %i[edit update]
 
-    # TODO: Extract year constraint to future proof for future cycles
-    resources :recruitment_cycles, param: :year, constraints: { year: /2021|2022/ }, path: "", only: :show do
+    resources :recruitment_cycles, param: :year, constraints: { year: /#{Settings.current_cycle}|#{Settings.current_cycle + 1}/ }, path: "", only: :show do
       get "/details", on: :member, to: "providers#details"
       get "/contact", on: :member, to: "providers#contact"
       put "/contact", on: :member, to: "providers#update"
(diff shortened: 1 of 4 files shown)


aliuk2012:
[[] Extract year constraint to future proof for future cycles](https://trello.com/c/jwC8kKnb/379-extract-year-constraint-to-future-proof-for-future-cycles)

config/routes.rb
```diff
@@ -69,7 +69,7 @@
     resources :contacts, only: %i[edit update]
 
     # TODO: Extract year constraint to future proof for future cycles
-    resources :recruitment_cycles, param: :year, constraints: { year: /2021|2022/ }, path: "", only: :show do
+    resources :recruitment_cycles, param: :year, constraints: { year: /#{Settings.current_cycle}|#{Settings.current_cycle + 1}/ }, path: "", only: :show do
```
defong, line 72: can we have a tests like 

https://github.com/DFE-Digital/teacher-training-api/blob/0dff3249fb4dd2150b1a4863f63ddb3ca08b96ea/spec/config/settings_spec.rb#L11

please

defong (approved):

spec/config/settings_spec.rb
```diff
@@ -0,0 +1,11 @@
+require "rails_helper"
+
+describe Settings do
+  describe "authentication" do
+    subject do
+      YAML.load_file(Rails.root.join("config/settings.yml"))
+    end
+    its(%w[current_cycle]) { should eq 2021 }
```
darokel, line 8: Could we space these two lines out? and remove the `#` below?

darokel (approved):
Just a minor cosmetic comment but looks good

PR #1468, opened 2020-12-11
Add turnstyle check in build workflow

vigneshmsft:
Add missing turnstyle check in build workflow

### DIFF
--- a/.github/workflows/build.yml
+++ b/.github/workflows/build.yml
@@ -57,6 +57,11 @@ jobs:
       run:
         working-directory: /app
     steps:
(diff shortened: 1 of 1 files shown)


defong (approved):

.github/workflows/build.yml
```diff
@@ -72,6 +72,11 @@ jobs:
       - name: Run Brakeman static analysis
         run:  rails brakeman
 
+      - uses: softprops/turnstyle@v1
```
saliceti, line 75: Can we move it to the top of the workflow?

PR #1495, opened 2021-01-08
Fail deploy job when smoke tests fail

vigneshmsft:
The fountainhead/actions-wait-for-check step does not fail
when the retrieved status check has failed.
We have to fail the job if the outcome is failure to prevent deploying to the next environment.


sfawcett123 (commented):
I would also check for timeout as well

felixtheflex (approved):

saliceti (approved):

PR #1547, opened 2021-02-17
Fix install-fetch-config for non bash shells

saliceti:
### Context
Manage key vault secrets

Use generic [ instead of bash [[

### Guidance to review
Run `make install-fetch-config`


felixtheflex (approved):

saliceti:
[Configure local dev environment via key vault](https://trello.com/c/8w5IK0gk/1089-configure-local-dev-environment-via-key-vault)

PR #1557, opened 2021-02-25
Point "View on Find" links to the correct sandbox domain

duncanjbrown:
### Context

Clicking the "view on find" link in Sandbox takes you to eg

https://sandbox.find-postgraduate-teacher-training.education.gov.uk/course/EES/A233

When you really want to go to

https://sandbox.find-postgraduate-teacher-training.service.gov.uk/course/EES/A233

Change the config so `search_ui_url` gets the correct domain when we use it to build the link.


defong (approved):
`search_ui` :banana: 
legacy naming

PR #1564, opened 2021-03-02
[1148] api migration implement publish maintenance mode and admin login only

mooktakim:
### Context

https://trello.com/c/UobfcRmi/1148-api-migration-implement-publish-maintenance-mode-and-admin-login-only

* Feature setting to enable maintenance mode, with a message.
* Admin users are able to sign in
* Non-admins redirected to sign in page
* Flash message is set on every page.

### DIFF
--- a/app/controllers/concerns/admin_only_maintenance_mode.rb
+++ b/app/controllers/concerns/admin_only_maintenance_mode.rb
@@ -10,9 +10,6 @@ private
 
   def redirect_non_admins_in_maintenance_mode
     return unless Settings.features.maintenance_mode.enabled
-
-    flash[:notice] = simple_format(Settings.features.maintenance_mode.message)
-
     return if %w[sign_in sessions].include?(controller_name)
 
     unless current_user && current_user["admin"]
--- a/app/controllers/concerns/admin_only_maintenance_mode.rb
+++ b/app/controllers/concerns/admin_only_maintenance_mode.rb
@@ -1,5 +1,6 @@
 module AdminOnlyMaintenanceMode
   extend ActiveSupport::Concern
+  include ActionView::Helpers::TextHelper
 
   included do
     before_action :redirect_non_admins_in_maintenance_mode
@@ -10,7 +11,7 @@ private
   def redirect_non_admins_in_maintenance_mode
     return unless Settings.features.maintenance_mode.enabled
 
-    flash[:notice] = Settings.features.maintenance_mode.message
+    flash[:notice] = simple_format(Settings.features.maintenance_mode.message)
 
     return if %w[sign_in sessions].include?(controller_name)
 
--- a/app/controllers/concerns/admin_only_maintenance_mode.rb
+++ b/app/controllers/concerns/admin_only_maintenance_mode.rb
@@ -12,6 +12,10 @@ private
     return unless Settings.features.maintenance_mode.enabled
     return if %w[sign_in sessions].include?(controller_name)
 
+    redirect_if_non_admin
+  end
+
+  def redirect_if_non_admin
     unless current_user && current_user["admin"]
       redirect_to sign_in_path
     end
--- a/app/controllers/concerns/admin_only_maintenance_mode.rb
+++ b/app/controllers/concerns/admin_only_maintenance_mode.rb
@@ -1,0 +1,21 @@
+module AdminOnlyMaintenanceMode
+  extend ActiveSupport::Concern
+
+  included do
+    before_action :redirect_non_admins_in_maintenance_mode
+  end
+
+private
+
+  def redirect_non_admins_in_maintenance_mode
+    return unless Settings.features.maintenance_mode.enabled
+
+    flash[:notice] = Settings.features.maintenance_mode.message
+
+    return if %w[sign_in sessions].include?(controller_name)
+
+    unless current_user && current_user["admin"]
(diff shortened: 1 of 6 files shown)


mooktakim:
[API migration - Implement Publish maintenance mode and admin login only](https://trello.com/c/UobfcRmi/1148-api-migration-implement-publish-maintenance-mode-and-admin-login-only)

app/controllers/concerns/admin_only_maintenance_mode.rb
```diff
@@ -0,0 +1,21 @@
...
+  included do
+    before_action :redirect_non_admins_in_maintenance_mode
+  end
+
+private
+
+  def redirect_non_admins_in_maintenance_mode
+    return unless Settings.features.maintenance_mode.enabled
+
+    flash[:notice] = Settings.features.maintenance_mode.message
```
gpeng, line 13: Not sure about this. It's going to end up popping up for admins on each page when maintenance mode is on. Actually that's probably not a big deal, I'll have a look what it's like on the review app.

app/controllers/concerns/admin_only_maintenance_mode.rb
```diff
@@ -0,0 +1,21 @@
...
+
+    return if %w[sign_in sessions].include?(controller_name)
```
gpeng, line 15: This is probably fine for now but I wonder if it would be better to have something that we add to the actual pages that are allowed in maintenance mode. Make it more explicit and visible on the controller rather than buried in here. This module gets included in `ApplicationController` so we'd be able to implement it in here and have it available within all of the controllers.

mooktakim: I was thinking it would be good to contain everything related to this feature inside this module. That way we don't have "if maintenance on" in multiple locations.

config/settings.yml
```diff
@@ -83,6 +83,9 @@ features:
     # actually starting when it would be set to false
     has_current_cycle_started?: true
     show_next_cycle_allocation_recruitment_page: false
+  maintenance_mode:
+    enabled: false
+    message: "This service will be unavailable from Thursday 4 March at 17:00 to Friday 5 March at 09:00, while we carry out maintenance.\n\nFind teacher training courses will still be available during this time. If you have any questions, contact us at becomingateacher@digital.education.gov.uk"
```
defong, line 88: Can we use translation en.yml file. 
https://guides.rubyonrails.org/i18n.html#using-safe-html-translations
Pretty please.

mooktakim: Yep, makes sense.

gpeng: Discussed this offline. We're going to leave in here so as we can update the message from ENV vars if need be without deploying.

app/controllers/concerns/admin_only_maintenance_mode.rb
```diff
@@ -0,0 +1,19 @@
...
+    before_action :redirect_non_admins_in_maintenance_mode
+  end
+
+private
+
+  def redirect_non_admins_in_maintenance_mode
+    return unless Settings.features.maintenance_mode.enabled
+    return if %w[sign_in sessions].include?(controller_name)
+
+    unless current_user && current_user["admin"]
```
gpeng, line 15: Can we extract this to a method... `redirect_if_non_admin` or something. Codeclimate is upset.

gpeng (approved):

saliceti:
Shall we add the environment variable as feature flag to https://github.com/DFE-Digital/publish-teacher-training/blob/master/terraform/workspace_variables/app_config.yml ?

vigneshmsft:
can we also include the feature flag here
https://github.com/DFE-Digital/publish-teacher-training/blob/master/terraform/workspace_variables/app_config.yml#L9

terraform/workspace_variables/app_config.yml
```diff
@@ -6,6 +6,9 @@ default: &default
   SETTINGS__FEATURES__SIGNIN_BY_EMAIL: false
   SETTINGS__FEATURES__SIGNIN_INTERCEPT: false
   SETTINGS__ROLLOVER: false
+  SETTINGS__FEATURES__MAINTENANCE_MODE__ENABLED: false
+  SETTINGS__FEATURES__MAINTENANCE_MODE__TITLE: "This service will be unavailable from Thursday 4 March at 17:00 to Friday 5 March at 09:00, while we carry out maintenance."
+  SETTINGS__FEATURES__MAINTENANCE_MODE__BODY: "Find teacher training courses will still be available during this time. If you have any questions, contact us at becomingateacher@digital.education.gov.uk"
```
vigneshmsft, line 11:
```suggestion
  SETTINGS__FEATURES__MAINTENANCE_MODE__TITLE: "This service will be unavailable from Thursday 4 March at 17:00 to Friday 5 March at 09:00, while we carry out maintenance."
  SETTINGS__FEATURES__MAINTENANCE_MODE__BODY: "Find teacher training courses will still be available during this time. If you have any questions, contact us at becomingateacher@digital.education.gov.uk"
```
thanks but not sure if TITLE and BODY are required here..

gpeng: I think it will be good if we can set these from the ENV in case we need to change the message if we have any problems. e.g. if we're not ready at 9am.

vigneshmsft (approved):

PR #1569, opened 2021-03-05
[1177] change secondary second subject hint to something that exists …

mooktakim:
…on list

### Context

https://trello.com/c/IozUAkYb/1177-hint-text-suggests-an-option-thats-not-possible

* Changed string to "Physics" which exists as a subject
* I could make it dynamic by choosing the first one from the dropdown list if we expect the list to change?

### Guidance to review

* Add new course, select secondary subject
* Click "Add a second subject (optional)" link

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/views/courses/subjects/_form_fields.html.erb
+++ b/app/views/courses/subjects/_form_fields.html.erb
@@ -27,7 +27,7 @@
 
           <div class="govuk-form-group">
             <%= form.label :subordinate_subject_id, 'Second subject (optional)', class: 'govuk-label' %>
-            <span class="govuk-hint">For example ‘Physics with Mathematics’</span>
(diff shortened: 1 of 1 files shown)


mooktakim:
[Hint text suggests an option that's not possible](https://trello.com/c/IozUAkYb/1177-hint-text-suggests-an-option-thats-not-possible)

AbigailMcP (approved):
I think it's ok for this to be static, not dynamic. There's an argument that the first one in the list might not necessary be (or remain to be) the best example for the input 🤷‍♀️

dankmitchell (approved):

PR #1592, opened 2021-03-29
Deploy to QA env from qa branch

vigneshmsft:
### Context

Use `qa` env as a review environment untill we build review apps support on PaaS. 

Stop deployment to `qa` from `master`
Deploy to `qa` for code pushes to `qa` branch.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review

### DIFF
--- a/.github/workflows/deploy.yml
+++ b/.github/workflows/deploy.yml
@@ -5,7 +5,7 @@ on:
     inputs:
       qa:
         description: Deploy to qa?
-        default: 'true'
+        default: 'false'
         required: true
       staging:
         description: Deploy to staging?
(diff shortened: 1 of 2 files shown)


.github/workflows/deploy.yml
```diff
@@ -5,7 +5,7 @@ on:
     inputs:
       qa:
         description: Deploy to qa?
-        default: 'true'
+        default: 'false'
```
darokel, line 8:
```suggestion
false
```

darokel (approved):

.github/workflows/deploy.yml
vigneshmsft: it has to be a 'string'

vigneshmsft: it has to be a 'string'

saliceti (approved):

PR #1616, opened 2021-04-26
[1484] Tab markup for accessibility

JonF1:
### Context
https://trello.com/c/bkB21acy/1484-tab-list-markup-course-show-page

* Adds role="tablist" to ul element
* Each list element has role "tab"
* Aria selected false to basic details tab which is not first selected on page load
* Tab index added so that tabs read as per screenshots below

### Guidance to review
Visit `organisations/1GT/2021/courses/28KJ/details` for example

Screenshot 2021-04-26 at 12 32 17
Screenshot 2021-04-26 at 12 30 25

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review

### DIFF
--- a/app/views/courses/show.html.erb
+++ b/app/views/courses/show.html.erb
@@ -45,7 +45,7 @@
 <div class="course-tabs govuk-tabs">
   <h2 class="govuk-tabs__title">Contents</h2>
 
-  <ul class="govuk-tabs__list">
+  <ul class="govuk-tabs__list" role="tablist">
     <li class="govuk-tabs__list-item govuk-tabs__list-item--selected" role="presentation">
       <%=
         link_to "#description",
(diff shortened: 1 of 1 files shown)


JonF1:
[Tab-list markup - course show page](https://trello.com/c/bkB21acy/1484-tab-list-markup-course-show-page)

defong (approved):

jesseyuen:
@JonF1 If we have `aria-selected="false"` on the non-selected tab, I would expect `aria-selected="true"` on the selected tab. Thoughts?

[More info and an example here](https://www.stefanjudis.com/blog/aria-selected-and-when-to-use-it/#the-actual-use-case-of-%60aria-selected%60).

JonF1:
> @JonF1 If we have `aria-selected="false"` on the non-selected tab, I would expect `aria-selected="true"` on the selected tab. Thoughts?
> 
> [More info and an example here](https://www.stefanjudis.com/blog/aria-selected-and-when-to-use-it/#the-actual-use-case-of-%60aria-selected%60).

Hi Jesse, for sure, there is a aria: { selected: true } as its embedded ruby on line 52, i think it didn't show up cause i didn't change the line but if you click the expand all button you can see it

jesseyuen (approved):

PR #1620, opened 2021-04-28
[1616] Amend basic app.json file

dankmitchell:
### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/Rakefile
+++ b/Rakefile
@@ -1,17 +1,18 @@
 # Add your own tasks in files placed in lib/tasks ending in .rake,
 # for example lib/tasks/capistrano.rake, and they will automatically be available to Rake.
 
-require "simplecov"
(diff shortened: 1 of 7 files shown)


.ruby-version
```diff
@@ -0,0 +1 @@
+2.7.2
```
dankmitchell, line 1: Unrelated but I wanted this for local development. `asdf` doesn't work for me

gpeng: I think you can add this to `.git/info/exclude` or do `git update-index --skip-worktree .ruby-version` to keep it out of the repo.

gpeng (approved):

PR #1661, opened 2021-05-21
[SPIKE][1757] Convert rollover interruptions pages to use cookies

FriedSock:
### Context
This is a spike, so tests are missing and extra comments are added
For the purposes of review only. Do not merge.

https://trello.com/c/GgPy61jB/1757-spike-avoid-adding-acceptedrollover2022-user-state

We want to avoid adding a new state for every time a user has to 
accept and interrupt screen. It is also likely that we will break
Apart the rollover and rollover-recruitment interrupt pages this year
as the allocations period ends before rollover begins. If we wanted
to have these two pages separate and maintain states, we could get
the user stuck in an infinite loop if we had both pages enabled
At the same time: 
 accepted_rollover -> accepted_rollover_recruitment -> accepted_rollover

This PR shows how we could use cookies to replace this mechanism. We
would lose the ability to query how many users have seen these screens
and accepted them. But we could add timestamps to the user and wipe
them after every rollover if that is something we really need. Since
We know rollover and allocations happen once a year, at roughly the same
time, using cookies with a 6 month expiry takes care of wiping for us.

### Guidance to review

Read the code comments. Try it out locally by enabling `rollover.can_edit_current_and_next_cycles` and `rollover.show_next_cycle_allocation_recruitment_page`, you should be greeted with these interrupt screens.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -123,23 +123,26 @@ private
       redirect_to accept_terms_path
     elsif user_state_to_redirect_paths[user_from_session.aasm.current_state]
       redirect_to user_state_to_redirect_paths[user_from_session.aasm.current_state]
+    elsif show_rollover_page?
+      redirect_to rollover_path
+    elsif show_rollover_recruitment_page?
+      redirect_to rollover_recruitment_path
     elsif use_redirect_back_to
       redirect_to session[:redirect_back_to] if session[:redirect_back_to].present?
       session.delete(:redirect_back_to)
     end
   end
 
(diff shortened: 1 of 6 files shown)


FriedSock:
[[SPIKE] Avoid adding accepted_rollover_2022 user state](https://trello.com/c/GgPy61jB/1757-spike-avoid-adding-acceptedrollover2022-user-state)

PR #1684, opened 2021-05-27
[1809] Use interrupt page acknowledgements for rollover interrupt pages

FriedSock:
### Context
https://trello.com/c/ztZfy8sZ/1809-avoid-adding-acceptedrollover2022-user-state

- Replaces the old mechanism for determining when to show interrupt pages based on user state
- Breaks apart the 2 rollover interrupt pages so they can be turned on independently 
- Calls a new endpoint on TTAPI for creating acknowledgements for interrupt pages per user/page type/recruitment cycle: https://github.com/DFE-Digital/teacher-training-api/pull/1942

### Guidance to review

Run locally in combination with https://github.com/DFE-Digital/teacher-training-api/pull/1942

- Check that the pages do not show up when rollover flags are off
- Create a new recruitment cycle on TTAPI (for the next year after the one you have set for current)
- Turn on `rollover.can_edit_current_and_next_cycles` and `rollover.show_next_cycle_allocation_recruitment_page` independently to check each page in isolation.
- Sign out and sign in again to verify that the acknowledgement is persisted

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -210,7 +210,7 @@ private
     session[:auth_user]["attributes"] = user.attributes
 
     add_provider_count_cookie
-    add_interrupt_pages(user)
+    add_interrupt_pages
   end
 
   def user_from_session
@@ -226,10 +226,10 @@ private
     logger.error "Error setting the provider_count cookie: #{e.class.name}, #{e.message}"
   end
 
-  def add_interrupt_pages(user)
+  def add_interrupt_pages
     return unless FeatureService.enabled?("rollover.can_edit_current_and_next_cycles") || FeatureService.enabled?("rollover.show_next_cycle_allocation_recruitment_page")
 
-    InterruptPageAcknowledgement.where(user_id: user.id, recruitment_cycle_year: Settings.current_cycle.next).all.each do |acknowledgement|
+    InterruptPageAcknowledgement.where(user_id: current_user["user_id"], recruitment_cycle_year: Settings.current_cycle.next).all.each do |acknowledgement|
       key = "accepted_#{acknowledgement.page}"
       session[:auth_user][key] = true
     end
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -227,6 +227,8 @@ private
   end
 
   def add_interrupt_pages(user)
+    return unless FeatureService.enabled?("rollover.can_edit_current_and_next_cycles") || FeatureService.enabled?("rollover.show_next_cycle_allocation_recruitment_page")
+
     InterruptPageAcknowledgement.where(user_id: user.id, recruitment_cycle_year: Settings.current_cycle.next).all.each do |acknowledgement|
       key = "accepted_#{acknowledgement.page}"
       session[:auth_user][key] = true
--- a/app/controllers/application_controller.rb
+++ b/app/controllers/application_controller.rb
@@ -229,7 +229,7 @@ private
   def add_interrupt_pages
     return unless FeatureService.enabled?("rollover.can_edit_current_and_next_cycles") || FeatureService.enabled?("rollover.show_next_cycle_allocation_recruitment_page")
 
-    InterruptPageAcknowledgement.where(user_id: current_user["user_id"], recruitment_cycle_year: Settings.current_cycle.next).all.each do |acknowledgement|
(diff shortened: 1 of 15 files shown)


spec/controllers/application_controller_spec.rb
```diff
@@ -191,6 +212,11 @@
             .to eq 2
         end
 
+        it "sets acknowledges pages", 'feature_rollover.can_edit_current_and_next_cycles': true do
+          expect(controller.request.session[:auth_user]["accepted_lovely_page"])
```
AbigailMcP, line 216: :)

FriedSock:
[Avoid adding accepted_rollover_2022 user state](https://trello.com/c/ztZfy8sZ/1809-avoid-adding-acceptedrollover2022-user-state)

FriedSock:
I would really prefer not to "fix" the thing codeclimate complains about. An elsif statement is simple. To make it stop complaining you either need to create more methods to have more indirection. or create a list of predicates and actions like this:
```
  def check_interrupt_redirects(use_redirect_back_to: true)
    # Codeclimate prefers this kind of madness
    redirects = [
      [-> { !user_from_session.accepted_terms? }, -> { redirect_to accept_terms_path }],
      [-> { user_from_session.aasm.current_state == :new }, -> { redirect_to transition_info_path }],
      [-> { show_rollover_page? }, -> { redirect_to rollover_path }],
      [-> { show_rollover_recruitment_page? }, -> { redirect_to rollover_recruitment_path }],
      [
        -> { use_redirect_back_to },
        lambda {
          redirect_to session[:redirect_back_to] if session[:redirect_back_to].present?
          session.delete(:redirect_back_to)
        },
      ],
    ]

    redirects.each do |predicate, action|
      next unless predicate.call

      action.call

      # rubocop thinks this is better than just "return"
      return nil
    end
  end
```

Which I think is overkill for a small elsif statement

AbigailMcP:
> Which I think is overkill for a small elsif statement

this 🙏 

I'm very happy to leave as is!

AbigailMcP (commented):
Happy with this! Just need to wait until the API one is in before merging this, so just holding off the 'Approve' until then in case we get merge happy!

app/controllers/application_controller.rb
```diff
@@ -222,6 +226,15 @@ def add_provider_count_cookie
     logger.error "Error setting the provider_count cookie: #{e.class.name}, #{e.message}"
   end
 
+  def add_interrupt_pages
+    return unless FeatureService.enabled?("rollover.can_edit_current_and_next_cycles") || FeatureService.enabled?("rollover.show_next_cycle_allocation_recruitment_page")
+
+    InterruptPageAcknowledgement.where(user_id: current_user["user_id"], recruitment_cycle_year: Settings.current_cycle.next).all.each do |acknowledgement|
```
gpeng, line 232: This looks good. Hard to get my head around sending the next cycle year but I'm not sure I've any better idea.

app/models/user.rb
```diff
@@ -13,19 +13,15 @@ def self.member(id)
   aasm whiny_transitions: false do
     state :new, initial: true
     state :transitioned
+    # TODO:  Remove all these after migrating users all to "transitioned",
```
gpeng, line 16: I've started a ticket for this https://trello.com/c/6wQFnuhP/1443-migrate-users-to-transitioned if you want to remove the TODO. No problem if you want to leave it there for context though.

spec/support/features.rb
```diff
@@ -1,25 +1,38 @@
-RSpec.configure do |configure|
+RSpec.configure do |config|
   # Allow examples to be tagged with "feature :some_feature". This will turn that
   # feature just for the duration of that test.
-  configure.around :each do |example|
-    original_features = {}
+  normalise_feature_name = ->(feature) { feature.to_s.delete_prefix("feature_").to_sym }
```
gpeng, line 4: We're definitely outgrowing features in settings. This is getting complicated 🤣

gpeng (commented):
This is looking good 👍

app/controllers/application_controller.rb
FriedSock: I was thinking an alternative would be naming the pages per year, eg `rollover_2022`. But then it would mean we would need to keep updating the enum on TTAPI, or just drop validation altogether

AbigailMcP (approved):

PR #1727, opened 2021-06-10
[1918] Fixed to show previous allocations

defong:
### Context
Allocations and accredited provider and its training providers
Previously the list of training providers (for making `repeat` request type) has actual relationship with its training providers. Based on that they are running a PE course for next year as will as they have previously being allocated.

Now the same list will be based solely on previously being allocated.

### Guidance to review
Commits are probably best to view as 2 parts with the first 3 of making the changes and its related tests. The latter 3 as make the changes downstream.

You need to actually spin this up, and use https://github.com/DFE-Digital/teacher-training-api/pull/1972 

Allocations
-  is a closed loop system
- it has a sliding window of time
  - aka `journey mode`
  - it has it own lifecycle from `open`  -> `closed` -> `confirmed`
- it has its own request_type, `initial` (first timer), `repeat` (subsequence), `declined` (dead end)
- it spawns across two allocation cycles, 
  - the `previous` which holds the list of training providers that is encouraged to be used for this `current`
  - the list of training providers (for making `repeat` request type)
    - may not be related to the accredited body due to time preceptive so may not be an existing provider or still an training provider

#### Notes
It is a bit fragile as new allocations can be made with non existing training provider

#### Before
image
#### After
image

#### Before
image

#### After
image

#### Before
image

#### After
image

image

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/controllers/providers/allocations_controller_spec.rb
+++ b/spec/controllers/providers/allocations_controller_spec.rb
@@ -1,0 +1,124 @@
+require "rails_helper"
+
+RSpec.describe Providers::AllocationsController, type: :controller do
+  let(:current_recruitment_cycle) { build(:recruitment_cycle) }
+  let(:accredited_body) { build(:provider, accredited_body?: is_accredited_body, recruitment_cycle: current_recruitment_cycle) }
+  let(:is_accredited_body) { true }
+  let(:user) { build(:user) }
+  let(:allocations_state) { "open" }
+  let(:current_user) do
+    {
+      user_id: 1,
+      uid: SecureRandom.uuid,
+      info: {
+        email: "dave@example.com",
+      },
+      admin: user.admin,
+      attributes: user.attributes,
+    }.with_indifferent_access
+  end
+
+  before do
+    allow(controller).to receive(:current_user).and_return(current_user)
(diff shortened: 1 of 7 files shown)


defong:
[Allocations page 'tweak'](https://trello.com/c/4tNE2tay/1918-allocations-page-tweak)

spec/controllers/providers/allocations_controller_spec.rb
```diff
@@ -0,0 +1,124 @@
+require "rails_helper"
+
+RSpec.describe Providers::AllocationsController, type: :controller do
```
gpeng, line 3: Aren't we duplicating a lot testing here? We have a feature spec which we prefer over these.

gpeng (commented):
Looks good. Not sure we need the controller spec.

spec/controllers/providers/allocations_controller_spec.rb
defong: No, they test all the low hanging fruits that is not tested in the feature spec.

Such as which partial template is expected to be rendered, and the training providers that are all expected to be.

gpeng (approved):

PR #1733, opened 2021-06-15
Add view_component gem

davidgisbey:
### Context

We can remove a lot of logic from the views and unit test it with view components 

- Add view_component gem 

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review


defong (changes requested):
Huh?

https://github.com/DFE-Digital/publish-teacher-training/pull/1722

PR #1749, opened 2021-06-23
Add `title_with_error_prefix` helper, and use it for all pages with forms

paulrobertlloyd:
### Context

When validating a page, the design system recommends that we:

> add ‘Error: ’ to the beginning of the page <title> so screen readers read it out as soon as possible

We are only doing this in a few places.

* Add `title_with_error_prefix` used on other BAT services.
* Update page titles to use this helper
* Add this helper to pages that have form validation and display an error summary (not all pages include a summary when they probably should, but can cover the in another PR)

### Guidance to review

Am I testing for the right errors on these pages? Tests pass, and going through the app, `Error: ` is prefixed to those pages when showing validation.

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -40,6 +40,10 @@ module ViewHelper
     govuk_mail_to bat_contact_email_address, name || bat_contact_email_address_with_wrap, **kwargs
   end
 
+  def title_with_error_prefix(title, error)
+    "#{t('page_titles.error_prefix') if error}#{title}"
+  end
+
   def enrichment_error_url(provider_code:, course:, field:)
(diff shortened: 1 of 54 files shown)


frankieroberto (approved):
🙌 

Not for this PR, but I wonder whether it’d be helpful to have a common rspec test we could use for this, as it’s so easy to forget?

Maybe something like

```ruby
expect(page).to show_errors([
  "First error message", 
  "Second error message"
 ])
```

and the test could check:

1. `Error:` prefix in page title
2. Error summary shown with `There is a problem` title
3. Error summary appears before `<h1>`?
4. Each error message is included in the summary, and they each link to a valid element ID.
5. The fields that the error summary links to have the red border class, and also contain the same error text

PR #1761, opened 2021-06-25
[1719] Log web requests to big query

FriedSock:
### Context
https://trello.com/c/AhxWkotk/1719-add-logging-of-web-requests-to-bigquery

Adds a concern and sidekiq job for sending request data to google biquery

### Guidance to review

Have a hander at the google bigquery UI and see if there is anything in the events table (if you have access)

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


app/jobs/send_request_events_to_bigquery.rb
```diff
@@ -0,0 +1,10 @@
+class SendRequestEventsToBigquery < ActiveJob::Base
+  TABLE_NAME = 'events'.freeze
```
misaka, line 2: We're going to change Find and probably Apply to make this a configuration point. I'd suggest using `Settings.google.big_query_table`

config/initializers/bigquery.rb
```diff
@@ -0,0 +1,12 @@
+require 'google/cloud/bigquery'
+
+BIG_QUERY_API_JSON_KEY = Settings.google.bigquery.api_json_key
+
+if BIG_QUERY_API_JSON_KEY.present?
```
misaka, line 5: Would it make more sense to use the feature flag as the condition here?

```suggestion
  if FeatureService.enabled?(:send_request_data_to_bigquery)
```

app/jobs/send_request_events_to_bigquery.rb
```diff
@@ -0,0 +1,10 @@
+class SendRequestEventsToBigquery < ApplicationJob
+  def perform(request_event_json)
+    return unless FeatureService.enabled?(:send_request_data_to_bigquery)
```
misaka, line 3: Just thinking about the alternative branch here, where the feature isn't enabled. I'm assuming that the SideKiq job will be dropped at this point?

misaka (approved):
Looks good. Events coming through to BQ too. 👍

misaka (approved):

PR #1768, opened 2021-06-25
Tidy up forms that use GOV.UK Form Builder

paulrobertlloyd:
### Context

Where we’re using the GOV.UK Form Builder, let’s use it consistently.

* Set `page_title` variable where this is needed twice
* Use `f` for forms that use the GOV.UK Form Builder, `form` for those that don’t, so that it’s clear which parts of the code base are using the older methods.
* Consistent code style, especially for `form_with`
* Consistent layout, with error summary always above the page heading
* Use localisation strings where possible

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


gpeng (approved):
We should probably decide at some point whether we're using the locale file(s) or not and be consistent. Not in this PR obvs, just sayin' 😁

paulrobertlloyd:
> We should probably decide at some point whether we're using the locale file(s) or not and be consistent.

Yeah, this is my bad. Was hoping to move all forms over to the form builder, but this turned out to be a more complex task than I imagined, given the way forms for creating a new course are generated.

PR #1775, opened 2021-06-30
[2089] Fix delete review app timeout

saliceti:
### Context
Delete review app may fail if the concurrent deploy takes too long

Increase to 30 min and fail the workflow immediately if the timeout is reached

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review


saliceti:
[Fix delete review app timeout](https://trello.com/c/s3prIGus/2089-fix-delete-review-app-timeout)

avinhurry (approved):

PR #1782, opened 2021-07-01
[2121] Enrich missing UKPRN validation with link to UKPRN form

jhenderson:
### Context
https://trello.com/c/OnSfBQvM/2121-s-display-required-notice-for-ukprn-with-form-to-set-it-on-publish

- Enrich UKPRN validation error with a link that takes the user to a new form page to capture the UKPRN value

### Guidance to review
- Checkout this [Training API branch](https://github.com/DFE-Digital/teacher-training-api/pull/2022) locally and run it
- Load up the Publish app locally so that it connects to the API mentioned previously
- Find a course from the 2022-2023 recruitment cycle
- Try and publish it
- You should get a validation error about missing UKPRN
- Click on the validation error which should take you to the new form
- Play around with validation to make sure it works, then enter a valid value
- Trying publishing the course again - should not display any validation error relating to UKPRN

### Screenshots
# Validation error
Screenshot 2021-07-01 at 14 14 45

# New form to capture UKPRN
Screenshot 2021-07-01 at 14 16 27

### Checklist
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [x] Product Review

### DIFF
--- a/app/views/providers/references/edit.html.erb
+++ b/app/views/providers/references/edit.html.erb
@@ -1,0 +1,35 @@
+<% page_title = @form_object.lead_school? ? "Organisation reference numbers" : t("helpers.label.provider.ukprn") %>
+<%= content_for :page_title, title_with_error_prefix(page_title, @form_object.errors.present?) %>
+
+<% content_for :before_content do %>
+  <%= govuk_back_link_to(details_provider_recruitment_cycle_path(@provider.provider_code, @provider.recruitment_cycle_year)) %>
+<% end %>
+
+<div class="govuk-grid-row">
+  <div class="govuk-grid-column-two-thirds">
+    <%= form_with(
+      model: @form_object,
+      url: provider_recruitment_cycle_references_path(@provider.provider_code, @provider.recruitment_cycle_year),
+      method: :post,
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder,
+    ) do |f| %>
+
+      <%= f.govuk_error_summary %>
+
+      <% if @form_object.lead_school? %>
+        <h1 class="govuk-heading-l">
+          <span class="govuk-caption-l"><%= @provider.provider_name %></span>
+          <%= page_title %>
+        </h1>
+
+        <%= f.govuk_text_field(:ukprn, label: { text: t("helpers.label.provider.ukprn"), size: "m" }, width: 10, data: { qa: "ukprn" }) %>
+
+        <%= f.govuk_text_field(:urn, label: { text: t("helpers.label.provider.urn"), size: "m" }, width: 10, data: { qa: "urn" }) %>
+      <% else %>
+        <%= f.govuk_text_field(:ukprn, caption: { text: @provider.provider_name, size: "l" }, label: { text: t("helpers.label.provider.ukprn"), size: "l" }, width: 10, data: { qa: "ukprn" }) %>
+      <% end %>
+
+      <%= f.govuk_submit "Save changes" %>
+    <% end %>
+  </div>
+</div>
--- a/app/controllers/providers/references_controller.rb
+++ b/app/controllers/providers/references_controller.rb
@@ -1,0 +1,31 @@
+module Providers
+  class ReferencesController < ApplicationController
+    def edit
+      @form_object = ProviderReferencesForm.new(provider)
+    end
+
+    def update
+      @form_object = ProviderReferencesForm.new(provider, params: provider_params)
+
+      if @form_object.save
+        redirect_to details_provider_recruitment_cycle_path(provider.provider_code, recruitment_cycle.year)
+      else
+        render :edit
+      end
+    end
+
+  private
+
+    def provider
+      @provider ||= Provider.where(recruitment_cycle_year: recruitment_cycle.year).find(params[:provider_code]).first
+    end
+
+    def recruitment_cycle
+      @recruitment_cycle ||= RecruitmentCycle.find(params[:recruitment_cycle_year]).first
+    end
+
+    def provider_params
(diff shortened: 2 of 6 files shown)


jhenderson:
[[S] Display required notice for UKPRN with form to set it on publish](https://trello.com/c/OnSfBQvM/2121-s-display-required-notice-for-ukprn-with-form-to-set-it-on-publish)

app/views/providers/ukprns/edit.html.erb
```diff
@@ -0,0 +1,30 @@
...
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder,
+    ) do |f| %>
+
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= @provider.provider_name %></span>
+      </h1>
+
+      <%= f.govuk_text_field(:ukprn, label: { text: page_title, size: "m" }, width: 10, data: { qa: "ukprn" }) %>
```
paulrobertlloyd, line 25:
```suggestion
      <%= f.govuk_text_field(:ukprn,
        caption: { text: @provider.provider_name, size: "l" },
        label: { text: page_title, size: "l", tag: "h1" },
        width: 10,
        data: { qa: "ukprn" }) %>
```

jhenderson: We have since add another field `urn` which makes the argument for the heading to sit above the two fields. Example: 

Screenshot 2021-07-02 at 16 31 39

Let me know what you think.

darokel (approved):
👍

app/views/providers/references/edit.html.erb
```diff
@@ -0,0 +1,32 @@
...
+    <%= form_with(
+      model: @form_object,
+      url: provider_recruitment_cycle_references_path(@provider.provider_code, @provider.recruitment_cycle_year),
+      method: :post,
+      builder: GOVUKDesignSystemFormBuilder::FormBuilder,
+    ) do |f| %>
+      <%= f.govuk_error_summary %>
+
+      <h1 class="govuk-heading-l">
+        <span class="govuk-caption-l"><%= @provider.provider_name %></span>
```
paulrobertlloyd, line 20: Ah, okay. In that case, just ensure this page has a heading:

```suggestion
        <span class="govuk-caption-l"><%= @provider.provider_name %></span>
        <%= page_title %>
```

app/views/providers/references/edit.html.erb
```diff
@@ -0,0 +1,32 @@
...
+      </h1>
+
+      <%= f.govuk_text_field(:ukprn, label: { text: "UK provider reference number (UKPRN)", size: "m" }, width: 10, data: { qa: "ukprn" }) %>
+
+      <% if @form_object.lead_school? %>
+        <%= f.govuk_text_field(:urn, label: { text: "Unique reference number (URN)", size: "m" }, width: 10, data: { qa: "urn" }) %>
+      <% end  %>
```
paulrobertlloyd, line 27:
```suggestion
      <% end %>
```

config/locales/helpers.yml
```diff
@@ -13,7 +13,7 @@ en:
         email: Email address
         telephone: Telephone number
         website: Website
-        ukprn: UK provider reference number (UKPRN)
+        references: Provider reference numbers
```
paulrobertlloyd, line 16: Don’t need this, as this is not a form label.

```suggestion
```

app/views/providers/references/edit.html.erb
```diff
@@ -0,0 +1,32 @@
+<% page_title = t("helpers.label.provider.references") %>
```
paulrobertlloyd, line 1:
```suggestion
<% page_title = "Provider reference numbers" %>
```

paulrobertlloyd:
@jhenderson There’s probably a case for having a slightly different design for one field versus 2; I can update this PR later if you like? Else the changes I’ve suggested above are a good stop gap.

jhenderson:
> @jhenderson There’s probably a case for having a slightly different design for one field versus 2; I can update this PR later if you like? Else the changes I’ve suggested above are a good stop gap.

Yeah, @gpeng and I discussed this and decided to have one form capture the UKPRN for non lead schools or UKPRN and URN for lead schools. The reason been was two fold:

1. When the user tries to publish and the provider doesn't have values to UKPRN and URN, it's easier for them fill them both out at the same time rather than have to it in two separate steps how it does mean that on the details page, the link for each field would point to the same page
2. Due to time contraints it was easier to have one page handle two scenarios instead of having two pages

Happy for you make any changes after the PR is merged

app/views/providers/references/edit.html.erb
jhenderson: If the form only displays one field, should we still display the head "Provider reference numbers"?

paulrobertlloyd:
Have updated this form so that it works when one field or 2 (this is what I meant by “having a slightly different design for one field versus 2”).

### Only asking for UKPRN (not a lead school):

Screenshot 2021-07-02 at 20 09 39

### Asking for UKPRN and URN (lead school):

Screenshot 2021-07-02 at 20 10 04

Changed the title in this instance to *Organisation reference numbers* as we refer to _Organisation_ elsewhere in Publish.

* * *

Sidenote: I shouldn’t need to add the localised strings for label text, but for some reason GOV.UK Form Builder isn’t picking this form up as being under the `provider` namespace. 🤷🏼

paulrobertlloyd:
In an ideal world, I think you’d want to take users to the `Organisation > About your organisation` page, with the missing fields validated, but understand if that’s too big a change given the pending rollover deadline.

PR #1797, opened 2021-07-06
Remove UCAS Apply GCSE requirements from next cycle

george-bit:
### Context

Hot fix to prevent this from displaying to providers. Test will follow.
https://trello.com/c/76ayE2pT/3600-%F0%9F%93%90-remove-ucas-gcse-requirements-from-new-course-creation-flow

Update to `CourseBasicDetailConcern` to skip over step and fix back link. Conditionally render UCAS row on confirmation page.

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/services/course_creation_step_service.rb
+++ b/app/services/course_creation_step_service.rb
@@ -19,6 +19,26 @@ class CourseCreationStepService
   end
 
   def get_workflow_steps(course)
+    if course.provider.recruitment_cycle_year.to_i >= Provider::CHANGES_INTRODUCED_IN_2022_CYCLE
+      workflows_for_2022_cycle_onwards(course)
+    else
+      workflows_for_current_cycle(course)
+    end
+  end
+
+private
+
+  def workflows_for_2022_cycle_onwards(course)
+    if course.is_further_education?
+      new_further_education_workflow_steps
+    elsif course.is_uni_or_scitt?
+      new_uni_or_scitt_workflow_steps
+    elsif course.is_school_direct?
+      new_school_direct_workflow_steps
+    end
+  end
+
+  def workflows_for_current_cycle(course)
     if course.is_further_education?
       further_education_workflow_steps
     elsif course.is_uni_or_scitt?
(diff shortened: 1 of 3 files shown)


app/services/course_creation_step_service.rb
```diff
@@ -18,7 +18,27 @@ def get_previous_step(steps, current_step)
     steps[previous_step_index]
   end
 
-  def get_workflow_steps(course)
+  def get_workflow_steps(course, provider)
+    if provider.recruitment_cycle_year.to_i >= Provider::CHANGES_INTRODUCED_IN_2022_CYCLE
```
davidgisbey, line 22: a bit nitty but can we just use `course.provider.recruitment_cycle.year` here so we don't have to pass the provider in.

george-bit: Not nitty at all, think it's the sensible thing to do here. Have just refactored 👍

app/services/course_creation_step_service.rb
```diff
@@ -77,4 +97,52 @@ def further_education_workflow_steps
...
+      courses_list
+      level
+      outcome
+      full_or_part_time
+      location
+      applications_open
+      start_date
+      confirmation
+    ]
+  end
```
defong, line 147: Is this a dupe?
```suggestion
  def new_further_education_workflow_steps
    further_education_workflow_steps
  end
```

app/services/course_creation_step_service.rb
```diff
@@ -77,4 +97,52 @@ def further_education_workflow_steps
...
+      courses_list
+      level
+      subjects
+      modern_languages
+      age_range
+      outcome
+      apprenticeship
+      full_or_part_time
+      location
+      applications_open
```
defong, line 130: `[*(:entry_requirement if course.provider.recruitment_cycle_year.to_i >= Provider::CHANGES_INTRODUCED_IN_2022_CYCLE)]`

Not fussed

defong (approved):
Let's get this in

app/services/course_creation_step_service.rb
george-bit: yep, will refactor 👍

george-bit:
@defong cool I think we need to wait for https://github.com/DFE-Digital/teacher-training-api/pull/2041 to be merged first otherwise providers won't be able to to create courses

PR #1799, opened 2021-07-06
[2170] Support both UKPRN validation formats

gpeng:
### Context

We need to change the validation error format which is set in TTAPI.
Unfortunately this is affects the functionality in publish.

Corresponding TTAPI PR -> https://github.com/DFE-Digital/teacher-training-api/pull/2042

Support both formats for now so as we don't break when the API changes
are made.

I'll take the old format out when the change has been made and deployed on TTAPI.

### Guidance to review

* Run publish against the current TTAPI
* Try to publish a course on a provider without a UKPRN
* Click the error message
* The UKPRN form should open

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/helpers/view_helper.rb
+++ b/app/helpers/view_helper.rb
@@ -147,9 +147,15 @@ private
         provider_recruitment_cycle_visas_path(provider_code, course.recruitment_cycle_year),
       "You must provide a Unique Reference Number (URN) for all course locations" =>
         provider_recruitment_cycle_sites_path(provider_code, course.recruitment_cycle_year),
+      "Enter a Unique Reference Number (URN) for all course locations" =>
+        provider_recruitment_cycle_sites_path(provider_code, course.recruitment_cycle_year),
       "You must provide a UK provider reference number (UKPRN)" =>
         provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
       "You must provide a UK provider reference number (UKPRN) and URN" =>
+        provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
+      "Enter a UK Provider Reference Number (UKPRN)" =>
+        provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
+      "Enter a UK Provider Reference Number (UKPRN) and URN" =>
         provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
       "Enter degree requirements" =>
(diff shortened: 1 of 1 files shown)


dankmitchell (approved):

app/helpers/view_helper.rb
```diff
@@ -151,6 +151,10 @@ def base_errors_hash(provider_code, course)
         provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
       "You must provide a UK provider reference number (UKPRN) and URN" =>
         provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
+      "You must provide a UK Provider Reference Number (UKPRN)" =>
+        provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
+      "You must provide a UK Provider Reference Number (UKPRN) and URN" =>
+        provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
```
paulrobertlloyd, line 157: Trying to move away from using ‘You must’ in validation strings:

```suggestion
      "Enter a UK Provider Reference Number (UKPRN)" =>
        provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
      "Enter a UK Provider Reference Number (UKPRN) and URN" =>
        provider_recruitment_cycle_references_path(provider_code, course.recruitment_cycle_year),
```

gpeng: 🤦 We should really settle on what these messages are supposed to be or we're going to be doing this forever and it's a faff. These strings have to match TTAPI for the functionality to work (I know 🤮 but this is the world we currently live in). @gregknight1 Could you just confirm and I'll sort it out?

gregknight1: @gpeng @paulrobertlloyd Yep, 'Enter..' is the right instruction. Just fyi - I have tickets to review validation messages in Publish and Register. And also a ticket to review capitalisation in both services - they just keep getting pushed back due to transition stuff.

gpeng: I've spoken to @gregknight1 . We're going to update to `Enter` as you suggest @paulrobertlloyd (and in one of the preceding messages that isn't in this PR yet) and then @gregknight1 will review capitalisation etc as a wider bit of work to get it all consistently correct.

gpeng: Updated in e192ee2

paulrobertlloyd (approved):

PR #1829, opened 2021-07-26
Display validation messages in 'red' when errors are present

TobyRet:
### Context

If a course page is showing validation, the messages for new Degree and GCSE fields should be red (`app-inset-text--error`), not blue (`app-inset-text--important`).

- Pass in the errors into the `DegreeRowContentComponent` and `GcseRequirementsComponent`.
- Declare a `#inset_text_css_classes` method on each component which returns the correct css classes depending on whether errors are present or not
 
### Guidance to review

Spin up Publish / API with a recent DB dumb. Find a provider who has courses in the next cycle (e.g. Claydon High School) and select a rolled over course. You should see that the GCSE and Degree fields are `blue` but when you try to publish, the error summary appears and those fields are now displayed in `red`.

This feels like a bit of a hack, however, currently we are reliant on the API to provide us with validation error messages. This seems to be the most pragmatic solution given that Publish and API are to be merged at some point?

|Before (when publishing the course) |After (when publishing the course)|
--- | --- |
|before|after|

### Trello
https://trello.com/c/wmYHB2RN/3620-%F0%9F%93%90-new-field-messages-should-be-red-if-course-page-is-showing-validation-publish

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/components/degree_row_content_component.html.erb
+++ b/app/components/degree_row_content_component.html.erb
@@ -9,7 +9,7 @@
     </p>
   <% end %>
 <% else %>
-  <%= govuk_inset_text(classes: "app-inset-text--narrow-border app-inset-text--important") do %>
+  <%= govuk_inset_text(classes: inset_text_css_classes) do %>
     <p class="govuk-heading-s app-inset-text__title">Do you require a minimum degree classification?</p>
     <p class="govuk-body">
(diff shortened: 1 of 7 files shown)


paulrobertlloyd (approved):
👌

PR #1845, opened 2021-08-06
Govuk components v2

peteryates:
Help make some failing specs pass while upgrading to GOV.UK Components v2.

### DIFF
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -1,3 +1,5 @@
+require "rails_helper"
+
 RSpec.feature "View helpers", type: :helper do
   describe "#enrichment_error_link" do
     context "with a course" do
--- a/spec/helpers/application_helper_spec.rb
+++ b/spec/helpers/application_helper_spec.rb
@@ -16,32 +16,52 @@ RSpec.feature "View helpers", type: :helper do
   end
 
   describe "#enrichment_summary" do
-    it "returns correct content" do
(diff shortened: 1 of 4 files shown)


spec/helpers/application_helper_spec.rb
```diff
@@ -14,32 +16,52 @@
...
-        expect(output[:value]).to eq("<div class=\"govuk-inset-text app-inset-text--narrow-border app-inset-text--error\"><a class=\"govuk-link\" href=\"/organisations/#{@provider.provider_code}/#{@course.recruitment_cycle_year}/courses/#{@course.course_code}/about?display_errors=true#about_course_wrapper\">Enter something about the course</a></div>")
+      before do
+        helper.enrichment_summary(summary_list, :course, "About course", "", [:about_course])
+      end
+
+      it "renders a value containing an error link within inset text" do
+        expect(subject).to have_css(".govuk-summary-list__key", text: "About course")
+        expect(subject).to have_css(".govuk-summary-list__value > .app-inset-text--error > a", text: error_message)
+
+        expect(subject).to have_link(error_message, href: "/organisations/#{@provider.provider_code}/#{@course.recruitment_cycle_year}/courses/#{@course.course_code}/about?display_errors=true#about_course_wrapper")
```
peteryates, line 64: `@course.recruitment_cycle_year` is blank, looks like it was before too. The test still passes but there's a `//` in the path, should probably be addressed by someone with a better idea of what's going on! 😅

PR #1851, opened 2021-08-12
Update content following a review

frankieroberto:
We did another round of reviewing this content, following some user testing.

A standard sentence has been added to the preview about degree subject needing to be closely related to the course subject – for secondary courses only.

The entry requirement content for the preview page has been separated from the entry requirement content shown in the course summary page to providers, as there are some slight differences.

This also copies across a bugfix for capitalisation of the GCSE subjects (only English should be capitalised) from Find.

See also: [Changes made in prototype](https://github.com/DFE-Digital/find-teacher-training-prototype/pull/45)


paulrobertlloyd:
@frankieroberto Is this the corresponding PR on Find? https://github.com/DFE-Digital/find-teacher-training/pull/873

frankieroberto:
@paulrobertlloyd yep, although there are a couple of updates to make to that too. Coming up!

spec/components/gcse_preview_component_spec.rb
```diff
@@ -41,12 +41,12 @@
 
         render_inline(described_class.new(course: course))
 
-        expect(page).to have_content("Candidates with pending GCSEs will be considered")
```
defong, line 44: image
Should it's counterpart be updated to reflect the changes as well?

defong (commented):
minor inconsistency

app/components/gcse_requirements_component.rb
```diff
@@ -20,12 +20,29 @@ def inset_text_css_classes
...
+  def required_gcse_summary_content(course)
+    case course.level
+    when "primary"
+      "Grade #{course.gcse_grade_required} (C) or above in English, maths and science, or equivalent qualification"
+    when "secondary"
+      "Grade #{course.gcse_grade_required} (C) or above in English and maths, or equivalent qualification"
+    end
+  end
+
   def required_gcse_content(course)
```
davidgisbey, line 32: Can this come out now then?

app/components/degree_preview_component.rb
```diff
@@ -8,6 +8,17 @@ def initialize(course:)
...
 private
 
+  def subject_name(course)
+    case course.subjects.count
+    when 1
+      course.subjects.first.subject_name
+    when 2
+      course.subjects.first.subject_name + " or " + course.subjects.last.subject_name
+    else
+      course.name
```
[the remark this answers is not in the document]
frankieroberto: @defong I had assumed that the default (`course.name`) would only be used if there are no subjects associated with the course - unless there are any courses with 3 subjects? Does that happen?

davidgisbey (approved):
LGTM once we've decided if we're happy with using the courses name for more than 2 subjects

spec/components/gcse_preview_component_spec.rb
frankieroberto: @defong not sure what you mean here? The content means the same, but is just worded slightly differently, as the right hand side version is provider-facing and a summary of their answers.

That said, just spotted that the paragraph break between "Candidates with pending GCSEs will be considered" and "Equivalency tests will be accepted in English or maths" needs to be restored, my bad.

frankieroberto: (pushed a commit to fix the missing paragraph break in the summary)

PR #1785, opened 2021-07-05
[2091] Remove note about limiting to 37 locations

archferns:
### Context
Related to https://github.com/DFE-Digital/teacher-training-api/pull/2025,
Remove note about limiting to 37 locations, as providers will now be allowed more than 37 locations.

### Changes proposed in this pull request
**Locations page Current**
image

**Locations page Proposed**
image

**Guidance page Current**
image

**Guidance page Proposed**
image

### Checklist

- [ ] Make sure all information from the Trello card is in here
- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Tested by running locally
- [ ] Product Review

PR #1857, opened 2021-08-16
[2339] Remove sites limit method

dankmitchell:
We no longer limit the number of sites per provider, see #1785 

🚢

### DIFF
--- a/app/views/sites/index.html.erb
+++ b/app/views/sites/index.html.erb
@@ -33,10 +33,6 @@
         <%= render @sites %>
       </tbody>
     </table>
-    <% if @provider.can_add_more_sites? %>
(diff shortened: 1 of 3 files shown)


dankmitchell:
[[S] Remove `can_add_more_sites?` from TTAPI and publish](https://trello.com/c/EQEGSnMD/2339-s-remove-canaddmoresites-from-ttapi-and-publish)

gpeng (approved):

defong:
Don't this needed to be only for next recruitment cycle?

dankmitchell:
> Don't this needed to be only for next recruitment cycle?

Originally, yes, however, we've already hardcoded to `true` anyway.

PR #1975, opened 2021-10-27
[271] Add AllocationUplift data to AllocationView

stellenberger:
### Context

- [This trello ticket](https://trello.com/c/wT8SermD/271-add-pe-uplift-section-in-pe-allocations)

- Change the allocation view to include the new uplift data

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/spec/factories/allocations.rb
+++ b/spec/factories/allocations.rb
@@ -3,8 +3,9 @@ FactoryBot.define do
     sequence(:id)
     association :provider
     association :accredited_body, factory: %i[provider accredited_body]
-    number_of_places { nil }
 
+    number_of_places { nil }
+    confirmed_number_of_places { nil }
     trait :repeat do
       request_type { "repeat" }
     end
@@ -18,7 +19,13 @@ FactoryBot.define do
     end
 
     trait :with_allocation_uplift do
-      association :allocation_uplift, uplifts: 5
+      association :allocation_uplift
+    end
+
+    after :build do |allocation|
+      # Neccessary hack to get attributes to persist through JSONAPI build
+
+      allocation.allocation_uplift&.uplifts = 5
     end
   end
 end
--- a/spec/factories/allocations.rb
+++ b/spec/factories/allocations.rb
@@ -18,7 +18,7 @@ FactoryBot.define do
       request_type { "initial" }
     end
 
-    trait :with_allocation_uplift do
+    trait :with_uplift do
       association :allocation_uplift
     end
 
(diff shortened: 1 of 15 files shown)


spec/factories/allocations.rb
```diff
@@ -16,5 +17,15 @@
     trait :initial do
       request_type { "initial" }
     end
+
+    trait :with_allocation_uplift do
+      association :allocation_uplift
+    end
+
+    after :build do |allocation|
+      # Neccessary hack to get attributes to persist through JSONAPI build
```
malcolmbaig, line 26: Can you elaborate? Not currently clear to me what this is for. Looks like `FactoryBot.build(:allocation, :with_allocation_uplift)` builds the objects I would expect with the correct data?

spec/factories/allocations.rb
```diff
@@ -16,5 +17,15 @@
     trait :initial do
       request_type { "initial" }
     end
+
+    trait :with_allocation_uplift do
```
malcolmbaig, line 21:
```suggestion
    trait :with_uplift do
```
Might be a little nicer for console work and when writing tests.

stellenberger: Agreed, pushed changes now 👍

spec/support/serializers/allocation_uplift_serializer.rb
```diff
@@ -0,0 +1,9 @@
+class AllocationUpliftSerializer < JSONAPI::Serializable::Resource
+  type "allocation_uplifts"
+
+  belongs_to :allocation
+
+  attributes(*FactoryBot.attributes_for("allocation_uplift").keys)
+
+  attribute :uplifts
```
malcolmbaig, line 8: Are both of these lines necessary?

spec/factories/allocations.rb
malcolmbaig: Suggest we add a link to the issue in the comment. This should all be quite temporary as we move to merge Publish/TTAPI, but worth clarifying any potential confusion.

Can this `after :build` be nested under the `trait :with_uplift`, so that it only occurs when that trait is used?

spec/support/serializers/allocation_uplift_serializer.rb
stellenberger: No, in fact, this file isn't necessary so I'll remove it now

spec/factories/allocations.rb
stellenberger: Agreed, added both suggestions 👍

malcolmbaig (approved):

PR #1979, opened 2021-11-02
[xxx] Update PE course form link

gpeng:
### Context

PE forms are requested via a Google form. This is linked from the course subject page and there is a different form for each year. We add the new form when we open rollover (in the 'next_cycle' settings) and then this should be transferred to the `current_cycle` when the new cycle is opened. We hadn't done that this year so it reverted back to last years form.

* Change the link for the current cycle to the 2022/23 form
* Change the name to make it more obvious what the link is

### Guidance to review

* Run the app
* Create a new secondary course
* Click the 'separate Google form' link on the subject page
* Marvel at the 2022/23 form that opens
* The forms are different for accredited bodies and non-accredited providers so we should check both

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/helpers/provider_helper.rb
+++ b/app/helpers/provider_helper.rb
@@ -3,9 +3,9 @@ module ProviderHelper
     cycle_key = is_current_cycle(provider.recruitment_cycle_year) ? "current_cycle" : "next_cycle"
 
     if provider.accredited_body?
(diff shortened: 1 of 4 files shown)


gpeng:
[Fix the PE forms link](https://trello.com/c/01sYxMgA/342-fix-the-pe-forms-link)

stellenberger (approved):

PR #1987, opened 2021-11-12
[373] Handle nil sessions

gpeng:
### Context

We recently added this to clear the persona login if the user clicks the back link onto the index page. We also need to handle when the user first arrives at the index page without a session.

* Handle session[:auth_user] being nil

### Checklist

- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/personas_controller.rb
+++ b/app/controllers/personas_controller.rb
@@ -14,7 +14,7 @@ if AuthenticationService.persona?
   private
 
     def clear_current_sessions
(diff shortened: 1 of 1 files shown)


gpeng:
[Personas are broken on QA](https://trello.com/c/Zkjzu7uv/373-personas-are-broken-on-qa)

defong (approved):

PR #2077, opened 2022-01-14
[0570] Added redirect for the details page to new publish

defong:
### Context
Redirects

Added redirect for the details, about, contact & visas pages to new publish

### Guidance to review

Use case along the lines of a bookmark:
-  navigate and try to force your way pass

> ie  
`/organisations/T92/2022/details`
`/organisations/T92/2022/about`
`/organisations/T92/2022/contact`
`/organisations/T92/2022/visas`

### Checklist

- [x] Publish / TTAPI Merge - does this code change affect a part of the app that's currently being migrated to TTAPI? If so, speak with the dev doing the migration and ensure they've accounted for this change.
- [x] Attach to Trello card
- [x] Rebased master
- [x] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/app/controllers/providers_controller.rb
+++ b/app/controllers/providers_controller.rb
@@ -26,7 +26,7 @@ class ProvidersController < ApplicationController
   end
 
   def details
-    redirect_to_new_publish_url_details_provider_recruitment_cycle_path if FeatureService.enabled?("new_publish.about_your_org")
+    redirect_to_new_publish_equivalent if FeatureService.enabled?("new_publish.about_your_org")
 
     redirect_to_contact_page_with_ukprn_error if @provider.ukprn.blank?
 
@@ -35,13 +35,13 @@ class ProvidersController < ApplicationController
   end
 
   def contact
-    redirect_to_new_publish_url_contact_provider_recruitment_cycle_path if FeatureService.enabled?("new_publish.about_your_org")
+    redirect_to_new_publish_equivalent if FeatureService.enabled?("new_publish.about_your_org")
 
     show_deep_linked_errors(%i[email telephone website address1 address3 address4 postcode])
   end
 
   def about
-    redirect_to_new_publish_url_about_provider_recruitment_cycle_path if FeatureService.enabled?("new_publish.about_your_org")
+    redirect_to_new_publish_equivalent if FeatureService.enabled?("new_publish.about_your_org")
 
     show_deep_linked_errors(%i[train_with_us train_with_disability])
   end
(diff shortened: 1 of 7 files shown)


defong:
[Add controller redirects to Old Publish for About Your Org actions](https://trello.com/c/FlyDX4qq/570-add-controller-redirects-to-old-publish-for-about-your-org-actions)

app/controllers/providers_controller.rb
```diff
@@ -26,6 +28,8 @@ def show
   end
 
   def details
+    redirect_to_new_publish_url_details_provider_recruitment_cycle_path if FeatureService.enabled?("new_publish.about_your_org")
```
malcolmbaig, line 31: I think we should do the same for `contact`, `about`, and `update` as well (ie—all actions that have now been reimplemented in New Publish). Also, `Providers::VisasController`.

defong: Oh right 😂

app/controllers/providers_controller.rb
```diff
@@ -177,6 +183,22 @@ def providers
...
+  end
+
+  def redirect_to_new_publish_url_details_provider_recruitment_cycle_path
+    redirect_to new_publish_url(details_provider_recruitment_cycle_path(@provider.provider_code, @provider.recruitment_cycle_year))
+  end
+
+  def redirect_to_new_publish_url_show_provider_recruitment_cycle_path
+    redirect_to new_publish_url(provider_recruitment_cycle_path(@provider.provider_code, @provider.recruitment_cycle_year))
+  end
+
```
malcolmbaig, line 201: Does something like this work?

A before action in this controller:
`before_action :redirect_to_new_publish_equivalent, only: [:details, :contact, :about, :update]`

With the method definition being:
```
def redirect_to_new_publish_equivalent
  redirect_to new_publish_url(request.current_path)
end
```

malcolmbaig: Actually that could be added to NewPublishHelper and accept the `request` object as an arg. Could then use that wherever needed as we migrate.

defong: done

app/controllers/providers_controller.rb
```diff
@@ -26,17 +26,23 @@ def show
...
   def details
+    redirect_to_new_publish_equivalent if FeatureService.enabled?("new_publish.about_your_org")
+
     redirect_to_contact_page_with_ukprn_error if @provider.ukprn.blank?
 
     @errors = flash[:error_summary]
     flash.delete(:error_summary)
   end
 
   def contact
```
malcolmbaig, line 37: I think we need a rebase here — there is now a separate controller for contacts

spec/controllers/application_controller_spec.rb
```diff
@@ -413,6 +413,14 @@ def index
     end
   end
 
+  describe "#redirect_to_new_publish_equivalent" do
+    it "returns a correct url linking back to the old publish" do
+      allow(request).to receive(:path).and_return("/haircut")
```
malcolmbaig, line 418: 😂

malcolmbaig (approved):

mbocc (approved):
clicking onto "about your org" leads to new Publish - thank you!

PR #2145, opened 2022-03-08
Update gmp in Dockerfile to fix vulnerability

RMcVelia:
### Context

SNYK has highlighted vulnerability in gmp 6.2.1-r0 in the base image
This is fixed in 6.2.1-r1

Change to Dockerfile to update gmp to the new version

### Guidance to review

Check build completes successfully

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/Dockerfile
+++ b/Dockerfile
@@ -1,4 +1,5 @@
 FROM ruby:2.7.5-alpine3.15
+# Remove apk add for gmp 6.2.1-r1 when the base image is updated
 RUN apk add --update --no-cache tzdata && \
     cp /usr/share/zoneinfo/Europe/London /etc/localtime && \
     echo "Europe/London" > /etc/timezone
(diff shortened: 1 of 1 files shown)


Dockerfile
```diff
@@ -7,6 +7,9 @@ RUN apk add --update --no-cache --virtual runtime-dependances \
  yarn \
  openssl-dev
 
+# Remove once base image ruby:2.7.5-alpine3.15 has been updated with latest gmp
```
NickGraham101, line 10: Is it worth putting another comment at the top next to the FROM line so we don't miss this when upgrading the base image?

RMcVelia: done

saliceti (approved):

PR #2149, opened 2022-03-14
Fix GHCR login for build-no-cache workflow

RMcVelia:
### Context
Fix Github Container Registry login for the build-no-cache workflow.

Alter workflow to use GHCR secrets instead of Dockerhub

### Guidance to review
Run workflow against branch

### Checklist

- [ ] Attach to Trello card
- [ ] Rebased master
- [ ] Cleaned commit history
- [ ] Product Review

### DIFF
--- a/.github/workflows/build-nocache.yml
+++ b/.github/workflows/build-nocache.yml
@@ -48,12 +48,13 @@ jobs:
           secret: ${{ env.KEY_VAULT_INFRA_SECRET_NAME }}
           key: SNYK_TOKEN
 
-      - name: Login to DockerHub
+      - name: Login to Github Container Registry
         if: github.actor != 'dependabot[bot]'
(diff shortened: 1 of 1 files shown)


.github/workflows/build-nocache.yml
```diff
@@ -48,12 +48,13 @@ jobs:
           secret: ${{ env.KEY_VAULT_INFRA_SECRET_NAME }}
           key: SNYK_TOKEN
 
-      - name: Login to DockerHub
+      - name: Login to Github Container Registry
         if: github.actor != 'dependabot[bot]'
-        uses: docker/login-action@v1
+        uses: docker/login-action@v1.12.0
```
saliceti, line 53: Can we keep it `@v1` ? This means it will always pick the latest minor update

saliceti (approved):

PR #2169, opened 2022-04-14
Enable courses and accredited body

darokel:
### Context

Turn on the following migration sections in prod:

- Courses
- Courses as an accredited body
- Users


config/settings/qa.yml
```diff
@@ -14,10 +14,5 @@ environment:
   selector_name: "qa"
 authentication:
   mode: persona   # none critical systems, ie localhost
-features:
-  rollover:
-    can_edit_current_and_next_cycles: true
```
darokel, line 19: This shouldn't be on right now so removing too. It's set to false in settings

defong: oh yeah that was @frankieroberto doing

defong (approved):

